PR #4, opened 2021-03-10
Add option to persist schedules to a file

jamesbornholt:
Large tests can produce failing schedules that are too unwieldy to
copy/paste into a call to `replay`. This change adds a new configuration
that can instead write schedules out to a file in a chosen directory.
When this option is enabled, failure messages instead print the name of
the file the schedule was saved to. Failing schedules saved in a file
can be replayed with a new `replay_from_file` convenience function.

The default config remains printing schedules to stdout/stderr, as this
seems like the best fit for use in CI pipelines, where the failing
schedule can be printed in the log file rather than needing to recover a
file from the build server.x

---

### DIFF
--- a/src/runtime/failure.rs
+++ b/src/runtime/failure.rs
@@ -1,0 +1,52 @@
+use std::fs::OpenOptions;
+use std::io::{ErrorKind, Write};
+use std::path::{Path, PathBuf};
+
+use crate::scheduler::serialization::serialize_schedule;
+use crate::scheduler::Schedule;
+use crate::{Config, FailurePersistence};
+
+/// Produce a message describing how to replay a failing schedule.
+pub fn persist_failure(schedule: &Schedule, message: String, config: &Config) -> String {
+    let serialized_schedule = serialize_schedule(schedule);
+    // Try to persist to a file, but fall through to stdout if that fails for some reason
+    if let FailurePersistence::File { directory } = &config.failure_persistence {
+        match persist_failure_to_file(&serialized_schedule, directory.as_ref()) {
+            Ok(path) => return format!("{}\nfailing schedule persisted to file: {}\npass that path to `shuttle::replay_from_file` to replay the failure", message, path.display()),
+            Err(e) => eprintln!("failed to persist schedule to file (error: {}), falling back to printing the schedule", e),
+        }
+    }
+    format!(
+        "{}\nfailing schedule: \"{}\"\npass that path to `shuttle::replay` to replay the failure",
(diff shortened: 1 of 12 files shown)


src/runtime/failure.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::scheduler::serialization::serialize_schedule;
+use crate::scheduler::Schedule;
+use crate::{Config, FailurePersistence};
+
+/// Produce a message describing how to replay a failing schedule.
+pub fn persist_failure(schedule: &Schedule, message: String, config: &Config) -> String {
+    let serialized_schedule = serialize_schedule(schedule);
+    // Try to persist to a file, but fall through to stdout if that fails for some reason
+    if let FailurePersistence::File { directory } = &config.failure_persistence {
+        let rslt = persist_failure_to_file(&serialized_schedule, directory.as_ref());
```
jorajeev, line 14: Can we just name this `result` ?

src/runtime/failure.rs
```diff
@@ -0,0 +1,52 @@
...
+    // Try to find the first usable filename. This is quadratic but we don't expect a ton of
+    // conflicts here.
+    let mut i = 0;
+    let dir = if let Some(dir) = destination {
+        dir.clone()
+    } else {
+        std::env::current_dir()?
+    };
+    let (path, mut file) = loop {
+        let path = dir.clone().join(Path::new(&format!("schedule{}.txt", i)));
```
jorajeev, line 37: How about `format!("schedule{:03}.txt", i)` here, so the schedules are sorted in a directory listing?

src/runtime/failure.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Produce a message describing how to replay a failing schedule.
+pub fn persist_failure(schedule: &Schedule, message: String, config: &Config) -> String {
+    let serialized_schedule = serialize_schedule(schedule);
+    // Try to persist to a file, but fall through to stdout if that fails for some reason
+    if let FailurePersistence::File { directory } = &config.failure_persistence {
+        let rslt = persist_failure_to_file(&serialized_schedule, directory.as_ref());
+        if let Ok(path) = rslt {
+            return format!("{}\nfailing schedule persisted to file: {}\npass that path to `shuttle::replay_from_file` to replay the failure", message, path.display());
+        }
```
jorajeev, line 17: How about printing an error message to warn that writing to the file failed, and we're falling back to console?

jorajeev (approved):

PR #7, opened 2021-03-19
Add Default impl to RwLock

jorajeev:
Add Default impl to RwLock

---


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle"
-version = "0.0.1"
+version = "0.0.2"
```
jamesbornholt, line 3: Let's do this in a separate commit so we can start writing a changelog too.

jorajeev: Yeah, good point.  Resubmitting

jamesbornholt (approved):

PR #10, opened 2021-03-26
Add some basic tests for metrics and fix an underflow

jamesbornholt:
If a `MetricsScheduler` gets dropped without running, the Drop handler
triggers an underflow in its `info!` call. This only happens if an interested
subscriber is registered. This change adds some basic tests that validate
the `MetricsScheduler` is generating the right values, including in the
Drop case.

---

### DIFF
--- a/tests/basic/metrics.rs
+++ b/tests/basic/metrics.rs
@@ -1,0 +1,118 @@
+use shuttle::scheduler::RandomScheduler;
+use shuttle::{check_random, thread, Runner};
+use std::panic::{catch_unwind, AssertUnwindSafe};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use tracing::field::{Field, Visit};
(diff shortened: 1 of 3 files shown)


jamesbornholt:
Looks like the build failures are Rust 1.51 issues, so opened a separate PR for that, and will rebase this one before merging: https://github.com/awslabs/shuttle/pull/11

tests/basic/metrics.rs
```diff
@@ -0,0 +1,118 @@
...
+                });
+            },
+            run_iterations,
+        );
+    }));
+
+    assert_eq!(result.is_err(), panic_iteration <= run_iterations);
+    assert_eq!(
+        metrics.iterations.load(Ordering::SeqCst),
+        run_iterations.min(panic_iteration)
```
jorajeev, line 91: That's a cute trick.

jorajeev (approved):

PR #14, opened 2021-04-01
Add config for max steps per iteration

jamesbornholt:
This change adds a configuration for an upper bound on the number of schedule steps
a single test iteration can take. When the bound is hit, the config controls whether
the test is considered failed or whether to continue to the next iteration (i.e., to
abort the current one and move on). By default, I'm setting the behavior to fail the
test if 1,000,000 steps are taken, so we can get earlier warnings about fairness issues
in Shuttle tests.

The DFS scheduler already had an option like this, which I've now removed and replaced
with the general config mechanism. Schedulers can still return `None` as the next task
to end an iteration early, though that functionality isn't used any more.

---


src/scheduler/dfs.rs
```diff
@@ -21,18 +20,15 @@ pub struct DfsScheduler {
...
 
 impl DfsScheduler {
-    /// Construct a new DFSScheduler with an optional bound on how many iterations to run
-    /// and an optional bound on the maximum depth to explore. A DFSScheduler can optionally
-    /// allow random data to be generated by the test (using `shuttle::rand`). Enabling random
-    /// data makes the DFS search incomplete, as it will not explore all possible values for
-    /// the random choices. To ensure determinism, each execution will use the same sequence of
-    /// random choices.
-    pub fn new(max_iterations: Option<usize>, max_depth: Option<usize>, allow_random_data: bool) -> Self {
+    /// Construct a new DFSScheduler. A DFSScheduler can optionally allow random data to be
```
jorajeev, line 23: Could you add back the comment about the `max_iterations` argument?

src/scheduler/mod.rs
```diff
@@ -12,6 +11,8 @@ mod round_robin;
 pub(crate) mod metrics;
 pub(crate) mod serialization;
 
+pub use crate::runtime::task::TaskId;
```
jorajeev, line 14: Why make this pub?

src/scheduler/dfs.rs
jamesbornholt: Oops, done

src/scheduler/mod.rs
jamesbornholt: It's to be able to implement the `Scheduler` trait, which is also pub, in client code (in this case, in a unit test). I'm actually surprised the compiler didn't complain about it before.

jorajeev (approved):

PR #15, opened 2021-04-03
Speed up two slow tests

jamesbornholt:
- producer_consumer_random can now use shuttle::rand to generate random
  parameters on every iterations, so the outer loop can go away. The diff looks
  big because of the indentation change.
- two_readers_and_one_writer can run the writer on the main thread, which
  dramatically reduces the search space for DFS

---


jorajeev (approved):
Strange that the diff algorithm is thrown off by the indentation change.  I thought it was more robust.

jamesbornholt:
Turns out there's a setting for that in the diff view:
image

PR #16, opened 2021-04-08
Allow spawning of an arbitrary number of tasks.

jorajeev:
Previously, we required users to specify (via the `max_tasks` parameter in the `Config`) the maximum number of tasks that could be spawned during an execution.  This change lifts that restriction, and allows spawning of an arbitrary number of tasks.

Note: The PCT scheduler still only supports creation of upto 16 tasks.  A forthcoming PR will update PCT to also support spawning of an arbitrary number of tasks.

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -1,5 +1,5 @@
 use crate::runtime::failure::persist_failure;
-use crate::runtime::task::{Task, TaskId, TaskState, TaskType, MAX_INLINE_TASKS};
+use crate::runtime::task::{Task, TaskId, TaskState, TaskType, DEFAULT_INLINE_TASKS};
 use crate::runtime::thread::future::ThreadFuture;
 use crate::scheduler::{Schedule, Scheduler};
 use crate::{Config, MaxSteps};
@@ -87,7 +87,7 @@ impl Execution {
                         .tasks
                         .iter()
                         .map(|t| (t.id, t.state))
-                        .collect::<SmallVec<[_; MAX_INLINE_TASKS]>>();
+                        .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
                     if task_states.iter().any(|(_, s)| *s == TaskState::Blocked) {
                         panic!(
                             "{}",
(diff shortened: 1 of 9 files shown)


tests/basic/config.rs
```diff
@@ -3,32 +3,29 @@ use shuttle::sync::Mutex;
 use shuttle::{thread, Config, Runner};
 use std::sync::Arc;
 
-fn check_max_tasks(max_tasks: usize, num_spawn: usize) {
-    let mut config = Config::new();
-    config.max_tasks = max_tasks;
-
-    let scheduler = RandomScheduler::new(100);
+#[test]
+fn many_tasks_with_mutex() {
```
jamesbornholt, line 7: Once we fix PCT, we should duplicate this test and run it under PCT as well?

src/runtime/execution.rs
```diff
@@ -201,10 +201,9 @@ impl ScheduledTask {
 
 impl ExecutionState {
     fn new(config: Config, scheduler: Rc<RefCell<dyn Scheduler>>, initial_schedule: Schedule) -> Self {
-        let max_tasks = config.max_tasks;
         Self {
             config,
-            tasks: Vec::with_capacity(max_tasks),
+            tasks: Vec::with_capacity(DEFAULT_INLINE_TASKS),
```
jamesbornholt, line 206: Might be worth changing this field into a `SmallVec` just for consistency.

jamesbornholt (approved):

src/runtime/task/mod.rs
```diff
@@ -175,31 +176,35 @@ impl From<TaskId> for usize {
...
     pub fn is_empty(&self) -> bool {
         self.tasks.iter().all(|b| !*b)
     }
 
     pub fn insert(&mut self, tid: TaskId) {
-        self.tasks[tid.0] = true;
+        if tid.0 >= self.tasks.len() {
+            self.tasks.resize(2 * self.tasks.len(), false);
+        }
+        assert!(self.tasks.len() > tid.0);
```
jamesbornholt, line 202: What happens if our current size is 32 and then we ask to insert task 70? I think we should just resize to length `tid.0 + 1` and let the `BitVec` decide how it wants to amortize that cost (it just delegates to `Vec` IIRC).

jamesbornholt (changes requested):

tests/basic/config.rs
jorajeev: 👍

src/runtime/task/mod.rs
jorajeev: Good catch; fixing

jamesbornholt (approved):

PR #18, opened 2021-04-10
Add hints to `yield_now` to allow schedulers to deprioritize a yielding thread

jamesbornholt:
Tasks that implement busy-wait loops or other infinite loops are a fairness
issue for schedulers like PCT, which will run a thread indefinitely until it
blocks. This change makes both sync and async versions of `yield_now` act as
a hint to the scheduler that the current task should be deprioritized. Only
the PCT scheduler acts on this hint, by moving the current task to the lowest
priority. We could also make the DFS scheduler react to this hint by not
allowing it to re-schedule the current task immediately, but that would be
unsound.

---

### DIFF
--- a/tests/asynch/pct.rs
+++ b/tests/asynch/pct.rs
@@ -1,0 +1,73 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::task::{Context, Poll};
+
+use shuttle::scheduler::PctScheduler;
+use shuttle::sync::Arc;
+use shuttle::{asynch, Config, MaxSteps, Runner};
+
+/// Like [`shuttle::asynch::yield_now`] but doesn't request a yield from the scheduler
+struct UnfairYieldNow {
+    yielded: bool,
+}
+
+impl Future for UnfairYieldNow {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
+        if self.yielded {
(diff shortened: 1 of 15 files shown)


tests/asynch/pct.rs
```diff
@@ -0,0 +1,36 @@
...
+                let count = count.clone();
+                asynch::spawn(async move {
+                    count.fetch_add(1, Ordering::SeqCst);
+                })
+            })
+            .collect::<Vec<_>>();
+
+        asynch::block_on(async move {
+            while count.load(Ordering::SeqCst) < NUM_TASKS {
+                asynch::yield_now().await;
```
jorajeev, line 32: How about adding a variant of this test without the yield, showing that PCT finds the schedule where the test panics because we hit the max_steps limit?

tests/basic/pct.rs
```diff
@@ -48,3 +51,32 @@ fn one_step() {
...
+            .map(|_| {
+                let count = count.clone();
+                thread::spawn(move || {
+                    count.fetch_add(1, Ordering::SeqCst);
+                })
+            })
+            .collect::<Vec<_>>();
+
+        while count.load(Ordering::SeqCst) < NUM_THREADS {
+            thread::yield_now();
```
jorajeev, line 79: Same as above -- add a variant of the test that panics without the yield?

jorajeev (commented):
Looks good -- maybe add tests that show PCT finds the unfair executions?

jamesbornholt:
Good idea, done.

jorajeev (approved):

PR #19, opened 2021-04-13
Update PCT scheduler to support an arbitrary number of tasks.

jorajeev:
PR#16 added support for spawning an arbitrary number of tasks, but the PCT scheduler still required the number of tasks to be specified statically (hardcoded to 16).  This PR extends the PCT scheduler to support an arbitrary number of tasks.  The main idea is that whenever a new task is seen, it is assigned an arbitrary priority.

---

### DIFF
--- a/src/scheduler/pct.rs
+++ b/src/scheduler/pct.rs
@@ -3,7 +3,7 @@ use crate::scheduler::data::random::RandomDataSource;
 use crate::scheduler::data::DataSource;
 use crate::scheduler::{Schedule, Scheduler};
 use rand::rngs::OsRng;
-use rand::seq::SliceRandom;
+use rand::seq::{index::sample, SliceRandom};
 use rand::{Rng, RngCore, SeedableRng};
 use rand_pcg::Pcg64Mcg;
 
@@ -19,7 +19,7 @@ pub struct PctScheduler {
     max_iterations: usize,
     max_depth: usize,
     iterations: usize,
-    // invariant: always contains all possible TaskIds (never added or removed)
+    // invariant: queue is downward closed; contains all elements in range [0, len)
     priority_queue: Vec<TaskId>,
     // invariant: length is self.max_depth - 1
     change_points: Vec<usize>,
@@ -41,8 +41,6 @@ impl PctScheduler {
 
         let rng = Pcg64Mcg::seed_from_u64(seed);
 
-        // TODO This implementation crashes if we have an application that spawns more than `DEFAULT_INLINE_TASKS`
-        // TODO Fix the code so that we can handle an arbitrary number of tasks
         Self {
             max_iterations,
             max_depth,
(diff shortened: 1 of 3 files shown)


src/scheduler/pct.rs
```diff
@@ -100,6 +97,13 @@ impl Scheduler for PctScheduler {
         // > enables/creates a second thread.
         // TODO is this really correct? need to think about it more
         if runnable.len() > 1 {
+            // If any new tasks were created, assign them priorities at random
+            let known_tasks = self.priority_queue.len();
+            let max_tasks = usize::from(*runnable.iter().max().unwrap());
+            for tid in known_tasks..max_tasks {
+                let index = self.rng.gen_range(1, self.priority_queue.len() + 1);
+                self.priority_queue.insert(index, TaskId::from(tid));
+            }
```
jamesbornholt, line 106: I think maybe we need to do this outside the `if runnable.len() > 1` check? It's possible that `runnable` contains a single task ID that we happen not to have seen before. I can't think of a way our current synchronization primitives would allow this (task spawns don't immediately block the spawner), but we should do the right thing anyway.

src/scheduler/pct.rs
```diff
@@ -100,6 +97,13 @@ impl Scheduler for PctScheduler {
         // > enables/creates a second thread.
         // TODO is this really correct? need to think about it more
         if runnable.len() > 1 {
+            // If any new tasks were created, assign them priorities at random
+            let known_tasks = self.priority_queue.len();
+            let max_tasks = usize::from(*runnable.iter().max().unwrap());
+            for tid in known_tasks..max_tasks {
+                let index = self.rng.gen_range(1, self.priority_queue.len() + 1);
```
jamesbornholt, line 104: Why is this offset by one? (a) the new task should be allowed to be top priority (b) `insert` should panic if given an index of `len + 1`

src/scheduler/pct.rs
```diff
@@ -68,22 +66,21 @@ impl Scheduler for PctScheduler {
...
-            let rng = &mut self.rng;
-            let max_steps = self.max_steps;
-            if max_steps > 1 {
-                for point in self.change_points.iter_mut() {
-                    *point = rng.gen_range(1, max_steps);
-                }
-            }
+            let num_points = std::cmp::min(self.max_depth - 1, self.max_steps - 1);
+            // sample(R, L, n) returns n distinct values in the range [0, L)
+            // but we want values in range [1, self.max_steps] so we offset by 1
```
jamesbornholt, line 79: Neat, this is way cleaner.

tests/basic/pct.rs
```diff
@@ -94,3 +94,154 @@ fn yield_spin_loop_fair() {
 fn yield_spin_loop_unfair() {
     yield_spin_loop(false);
 }
+
+#[test]
+#[should_panic(expected = "called `Option::unwrap()` on a `None`")]
+// Based on Fig 1(a) of the PCT paper.  We model NULL pointer dereference with an Option unwrap
```
jamesbornholt, line 100: Maybe use `expect` so we can target the exact "null dereference" in `should_panic` rather than any unwrap failing (same for the other tests)

tests/basic/pct.rs
```diff
@@ -94,3 +94,154 @@ fn yield_spin_loop_fair() {
...
+    // So probability of hitting the bug in 20 iterations = 1 - (1 - 1/2)^20 > 99.99%
+    let scheduler = PctScheduler::new(1, 20);
+    let runner = Runner::new(scheduler, Default::default());
+    runner.run(|| {
+        let t1 = Arc::new(Mutex::new(None));
+        let t2 = Arc::clone(&t1);
+
+        thread::spawn(move || {
+            for _ in 0..COUNT {
+                thread::yield_now();
```
jamesbornholt, line 113: Do these tests still work the way you expect now that `yield_now` deprioritizes?

tests/basic/pct.rs
```diff
@@ -94,3 +94,154 @@ fn yield_spin_loop_fair() {
...
+            thread::yield_now();
+        }
+    });
+}
+
+#[test]
+#[should_panic(expected = "called `Option::unwrap()` on a `None`")]
+fn figure1b_pct() {
+    // n=2, k=20, d=2, so probability of finding the bug in one iteration is at least 1/(2*20)
+    // So probability of hitting the bug in 200 iterations = 1 - (1 - 1/40)^200 > 99.36%
```
jamesbornholt, line 181: My totally arbitrary target for these was 99.9%, maybe bump the iterations a little to get there (same for `figure_1c`).

jamesbornholt (commented):
This was way easier than I expected! Love the new tests.

src/scheduler/pct.rs
jorajeev: Good call, will fix.

src/scheduler/pct.rs
jorajeev: `gen_range(l, h)` returns a value in `[l, h)`  so the max the above will return is `len`.
As for the lower bound, yeah, I agree we should allow it to be 0.  Fixing.

tests/basic/pct.rs
jorajeev: Done.

tests/basic/pct.rs
jorajeev: Yeah, I was working on these before you PR.  Will replace.

tests/basic/pct.rs
jorajeev: Done.

jamesbornholt (approved):

PR #25, opened 2021-05-27
Add a timeout option to Config

jamesbornholt:
In some applications, we want to run Shuttle for as many iterations as
possible within some time budget to increase our confidence in the
code's correctness. This change adds a new `max_time` config that will
end a test once the timeout is reached. The timeout is only checked
between test iterations, so it won't abort midway through a single test
iteration.

To help guard against slow tests, `Runner::run` now returns the number
of iterations it ran. In normal operation this will be whatever number
of iterations the scheduler dictated (e.g., through `max_iterations`
options), but if the timeout is reached it may be smaller. Callers can
check this value to make sure the test wasn't unreasonably slow.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -203,6 +203,12 @@ pub struct Config {
     /// Maximum number of steps a single iteration of a test can take, and how to react when the
     /// limit is reached
     pub max_steps: MaxSteps,
+
+    /// Time limit for an entire test. If set, calls to [`Runner::run`] will return when the time
+    /// limit is exceeded or the [`Scheduler`](crate::scheduler::Scheduler) chooses to stop (e.g.,
+    /// by hitting its maximum number of iterations), whichever comes first. This time limit will
+    /// not abort a currently running test iteration; the limit is only checked between iterations.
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -203,6 +203,13 @@ pub struct Config {
     /// Maximum number of steps a single iteration of a test can take, and how to react when the
     /// limit is reached
     pub max_steps: MaxSteps,
+
+    /// Time limit for an entire test. If set, calls to [`Runner::run`] will return when the time
+    /// limit is reached or the [`Scheduler`](crate::scheduler::Scheduler) chooses to stop (e.g., by
```
jorajeev, line 208: Nit: the check below is `start.elapsed() > t` so this should read "limit is exceeded" (or make the check below use `>=` instead of `>`

jorajeev (changes requested):

src/lib.rs
jamesbornholt: Good catch, fixed (and also fixed a similar nit in the same comment about when the limit is checked).

jorajeev (approved):

PR #28, opened 2021-06-09
Use continuations instead of futures to implement atomic tasks

JacobVanGeffen:
This PR changes the implementation of both synchronous tasks (threads) and asynchronous tasks (futures) to use continuations instead of futures. Previously, futures were used to implement both, as futures are more lightweight. However, doing so would prevent Shuttle from correctly discovering bugs when multiple futures need to run asynchronously. Also, the overhead of running one continuation per futures should not be prohibitively expensive.

---


src/asynch.rs
```diff
@@ -6,7 +6,8 @@
 //! [`futures::executor`]: https://docs.rs/futures/0.3.13/futures/executor/index.html
 
 use crate::runtime::execution::ExecutionState;
-use crate::runtime::task::{TaskId, TaskType};
+use crate::runtime::task::PendingTask;
+use crate::runtime::task::TaskId;
```
jamesbornholt, line 10: merge these into one line (your IDE might be generating them this way by default)

src/runtime/task/mod.rs
```diff
@@ -51,18 +70,23 @@ pub(crate) struct Task {
 }
 
 impl Task {
-    pub(crate) fn new(future: BoxFuture<'static, ()>, id: TaskId, task_type: TaskType, name: Option<String>) -> Self {
-        let waker = make_waker(id);
+    /// Create a task from a pending task
+    pub(crate) fn from_pending(task_id: TaskId, pending_task: PendingTask, name: Option<String>) -> Self {
```
jamesbornholt, line 74: probably just keep the name `id`

tests/asynch/basic.rs
```diff
@@ -96,6 +96,43 @@ fn async_spawn_chain() {
     );
 }
 
+#[test]
+fn async_thread_yield() {
+    // This tests if thread::yield_now can be called from within an async block
```
jamesbornholt, line 101: clever

tests/asynch/basic.rs
```diff
@@ -96,6 +96,43 @@ fn async_spawn_chain() {
...
+                thread::yield_now();
+            });
+            asynch::spawn(async move {});
+        },
+        None,
+    )
+}
+
+#[test]
+#[should_panic(expected = "assertion failed: `(left == right)`")]
```
jamesbornholt, line 114: This assumes the Rust assertion message will never change. It might be better to add a custom message to the `assert_eq!` and check for that.

jamesbornholt (commented):
Nice! I have a few small comments and a couple of suggestions about how we can make this a bit simpler.

src/runtime/task/async_task.rs
```diff
@@ -0,0 +1,43 @@
...
+    continuation.initialize(Box::new(move || {
+        // TODO should I be updating the waker on each iteration?
+        let waker = ExecutionState::with(|state| state.current_mut().waker());
+        let cx = &mut Context::from_waker(&waker);
+        while future.as_mut().poll(cx) == Poll::Pending {
+            ExecutionState::with(|state| {
+                // Want block_after_polling here because block_after_running will later be called
+                // Need to block here before thread::switch (if we didn't wake ourselves)
+                // BUT we don't want the execution loop to block with block_after_running in the case that we did wake ourselves
+                state.current_mut().block_after_polling();
```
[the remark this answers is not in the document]
JacobVanGeffen: (Commenting summary of our discussion for completeness)
We need to block before `thread::switch` because the scheduler may choose to continue the same task and avoid a context switch as an optimization.
Since `block_after_running` is called during this context switch, it will be skipped due to the optimization.
Previously, this wasn't an issue because futures could not call `thread::switch`.
The overall solution seems to be removing `block_after_running` from the context switch and instead requiring that futures mark themselves as blocked before calling `thread::switch`.

src/runtime/task/async_task.rs
```diff
@@ -0,0 +1,43 @@
...
+            });
+            // NOTE: After the thread::switch, step() will return TaskResult::Yielded,
+            // .    causing state.current_mut().block_after_running() to be called
+            thread::switch();
+            // TODO handle case where we are woken up after finishing?
```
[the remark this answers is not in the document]
JacobVanGeffen: The loop (and also continuation) will end once the future's `poll` function returns ready, so doing nothing is fine here.

src/runtime/task/mod.rs
```diff
@@ -51,18 +70,23 @@ pub(crate) struct Task {
...
             state: TaskState::Runnable,
             task_type,
 
-            future: Rc::new(RefCell::new(future)),
+            inner: Rc::new(RefCell::new(inner)),
 
             waiter: None,
-            waker,
+            // TODO Maybe don't make a waker for SyncTask?
+            waker: make_waker(task_id),
```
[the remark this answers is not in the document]
JacobVanGeffen: No real reason. Can change it back.

src/runtime/task/async_task.rs
```diff
@@ -0,0 +1,43 @@
+use crate::runtime::execution::ExecutionState;
+use crate::runtime::thread::continuation::{ContinuationPool, PooledContinuation};
+use crate::runtime::{task::*, thread};
+use futures::future::BoxFuture;
+use std::task::{Context, Poll};
+
+pub(crate) struct AsyncTask {
+    continuation: PooledContinuation,
+}
```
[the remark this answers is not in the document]
JacobVanGeffen: Good point, they can be the same type with two constructors.

tests/asynch/basic.rs
```diff
@@ -96,6 +96,43 @@ fn async_spawn_chain() {
...
+            });
+            asynch::spawn(async move {
+                assert_eq!(r.load(Ordering::SeqCst), 0);
+            });
+        },
+        None,
+    )
+}
+
 #[test]
```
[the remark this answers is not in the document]
JacobVanGeffen: I tried doing that, but couldn't get it to act any differently than it would before the PR. Any ideas for this?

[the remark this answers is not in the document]
jamesbornholt: Something like this fails today but I'd expect it to pass after your PR:

```
#[test]
fn async_mutex() {
    check_dfs(
        move || {
            let lock = Arc::new(Mutex::new(0u64));

            let t1 = {
                let lock = Arc::clone(&lock);
                asynch::spawn(async move {
                    let mut l = lock.lock().unwrap();
                    *l += 1;
                })
            };

            let t2 = asynch::block_on(async move {
                t1.await.unwrap();
                *lock.lock().unwrap()
            });

            assert_eq!(t2, 1);
        },
        None
    )
}
```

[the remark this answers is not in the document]
JacobVanGeffen: Oh nice! So this test doesn't actually pass with the current PR, but it passes once I remove `block_after_running` from the execution loop like we discussed. Not totally sure why yet so I'll do a bit more digging.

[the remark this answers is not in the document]
JacobVanGeffen: Seems to be that the reason is `lock.lock()` calls `thread::switch()`, then goes into the context switch. In there, `block_after_running` incorrectly blocks because the task is an async and it didn't wake itself up. So overall, not doing `block_after_running` in the context switch is for the best.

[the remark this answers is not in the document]
jamesbornholt: Yeah, that sounds right. I think moving the block-unless-self-woken check closer to the future logic, rather than baking it into the core of the executor as it is today, is for the best.

src/runtime/execution.rs
```diff
@@ -59,8 +56,8 @@ impl Execution {
 
         EXECUTION_STATE.set(&state, move || {
             // Spawn `f` as the first task
-            let first_task = ThreadFuture::new(config.stack_size, f);
-            ExecutionState::spawn(first_task, TaskType::Thread, None);
+            let first_task = AtomicTask::from_closure(config.stack_size, Box::new(f));
```
jamesbornholt, line 59: shouldn't need to box `f` here

src/runtime/task/mod.rs
```diff
@@ -1,45 +1,101 @@
...
 
 pub(crate) const DEFAULT_INLINE_TASKS: usize = 16;
 
+#[derive(Debug)]
+pub(crate) enum TaskResult {
+    Yielded,
+    Finished,
+}
+
+pub(crate) struct AtomicTask {
```
jamesbornholt, line 43: So now I'm wondering whether `AtomicTask` needs to exist at all :-)

Could we just inline the continuation and task type into `Task` itself, so `Task.inner` becomes `Rc<RefCell<PooledContinuation>>`, and `Execution::step` just calls `inner.resume()`? What do you think? Do we gain much from this abstraction?

src/asynch.rs
```diff
@@ -20,9 +20,9 @@ where
     T: Send + 'static,
 {
     let result = std::sync::Arc::new(std::sync::Mutex::new(None));
+    let stack_size = ExecutionState::with(|s| s.config.stack_size);
     let task_id = ExecutionState::spawn(
-        Wrapper::new(fut, std::sync::Arc::clone(&result)),
-        TaskType::Future,
+        AtomicTask::from_future(stack_size, Box::pin(Wrapper::new(fut, std::sync::Arc::clone(&result)))),
```
jamesbornholt, line 25: if you make the change I suggest below about removing the BoxFuture from `from_future`, you can remove `Box::pin` here

src/runtime/task/mod.rs
JacobVanGeffen: I tried doing something like this, but ended up changing a lot of unrelated stuff since I didn't wrap the `PooledContinuation` like that. I think this would be fine. The abstraction lets us construct `AtomicTasks`s in the library, then `Task`s in the execution loop, which doesn't really matter as far as I can tell.

jorajeev: +1 to James's comment.  It seems that now `Task` is the only abstraction we need.

jorajeev (approved):

src/asynch.rs
JacobVanGeffen: If we do this, we still have to `Box::pin` inside `from_future` right?

jamesbornholt: Right, but that makes a little more sense from an abstraction perspective -- the caller doesn't care whether the future is pinned, only the implementation does.

src/runtime/execution.rs
```diff
@@ -80,7 +77,7 @@ impl Execution {
             match state.current_task {
                 ScheduledTask::Some(tid) => {
                     let task = state.get(tid);
-                    Some((Rc::clone(&task.future), task.waker()))
+                    Some(Rc::clone(&task.inner))
```
JacobVanGeffen, line 80: Re: Removing AtomicTask.
I'm running into an issue with lifetimes when I try to return the task here.
Specifically, if I want to return `Some(state.get(tid))`, I get the following complaint:
```
lifetime may not live long enough
returning this value requires that `'1` must outlive `'2`rustc
execution.rs(73, 44): has type `&'1 mut ExecutionState`
execution.rs(73, 49): return type of closure is Option<&'2 Task>
```
Is there a way to fix this without changing `ExecutionState`'s `tasks` in `Rc`s?

JacobVanGeffen: Answer: I should be returning the continuation `Rc` here instead.

src/runtime/execution.rs
```diff
@@ -232,14 +225,26 @@ impl ExecutionState {
 
     /// Spawn a new task for a future. This doesn't create a yield point; the caller should do that
     /// if it wants to give the new task a chance to run immediately.
-    pub(crate) fn spawn(
-        future: impl Future<Output = ()> + 'static + Send,
-        task_type: TaskType,
-        name: Option<String>,
-    ) -> TaskId {
+    pub(crate) fn spawn_from_future<T, F>(future: F, stack_size: usize, name: Option<String>) -> TaskId
```
jamesbornholt, line 228: nit: I would call these `spawn_future` and `spawn_thread` respectively

src/runtime/task/mod.rs
```diff
@@ -36,10 +38,10 @@ pub(crate) const DEFAULT_INLINE_TASKS: usize = 16;
 pub(crate) struct Task {
     pub(super) id: TaskId,
     pub(super) state: TaskState,
-    // We use this to decide whether to block this task if it returns Poll::Pending
+    // We use this to check `block_unless_self_woken` is only called from a Future task
     task_type: TaskType,
 
-    pub(super) future: Rc<RefCell<BoxFuture<'static, ()>>>,
+    pub(super) inner: Rc<RefCell<PooledContinuation>>,
```
jamesbornholt, line 44: nit: call this field `continuation`?

jamesbornholt (commented):
A few final small things.

jamesbornholt (approved):

PR #29, opened 2021-06-10
Include default features for `rand`

jamesbornholt:
We're using `rand::seq::index`, which turns out to require the `alloc`
feature of that crate, but the documentation doesn't specify that. This
has been causing some funkiness in builds of packages that depend on
Shuttle, so let's just fix it here. To be honest, I don't even remember
why we turned off the default features for rand to start with---we
certainly don't expect Shuttle to work in `no_std` environments.

---


bkragl (approved):
I can confirm that this fixes the build error I encountered (in a Shuttle client). Thanks!

PR #30, opened 2021-06-10
Better handling for poisoned locks

jamesbornholt:
A common pattern in Rust code is to acquire a lock in a `Drop`
implementation, e.g. to return a resource to a shared pool when dropped.
If code panics while holding that same lock, the Drop handler will try
to enter the lock during unwinding, but the lock is poisoned. In real
Rust code that usually creates a double-panic (`.lock().unwrap()` will
fail). In Shuttle test code it's also a double panic, but the second
panic exposes an unhelpful error about internal Shuttle state.

This change propagates the poison status of `Mutex`s and `RwLock`s so
that the second panic in this scenarios is more helpful and looks the
same as in regular Rust code. To do this, we change where some of the
bookkeeping happens in the `Drop` impls for lock guards, so that the
lock is returned to a usable state. Tests that double-panicked before
will still double-panic after this change, but the message will be the same
as it would be under a non-Shuttle execution (that the lock is poisoned)
rather than exposing the guts of Shuttle's lock implementations.

One thing that's weird here is that the semantics of poisoning require
that we acquire the lock before returning the poison error (so that
client code can `into_inner()` the poison issue and continue as normal
if it so pleases). This change probably won't work very well in code that
does that, as the lock is being returned to a state that will allow it to be
acquired, but isn't unblocking any pending threads.

---

### DIFF
--- a/src/sync/rwlock.rs
+++ b/src/sync/rwlock.rs
@@ -5,7 +5,7 @@ use std::cell::RefCell;
 use std::ops::{Deref, DerefMut};
 use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::rc::Rc;
-use std::sync::{LockResult, TryLockResult};
+use std::sync::{LockResult, PoisonError, TryLockError, TryLockResult};
 use tracing::trace;
 
 /// A reader-writer lock, the same as [`std::sync::RwLock`].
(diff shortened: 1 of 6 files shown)


src/sync/rwlock.rs
```diff
@@ -254,31 +269,35 @@ impl<T> Drop for RwLockReadGuard<'_, T> {
...
-        RwLock::<T>::unblock_waiters(&*state, me, RwLockType::Read);
-        trace!(
-            holder = ?state.holder,
-            waiting_readers = ?state.waiting_readers,
-            waiting_writers = ?state.waiting_writers,
-            "released Read lock on rwlock {:p}",
-            self.state
-        );
+
+        if ExecutionState::should_stop() {
```
jorajeev, line 293: Can you say why this check had to be moved later?

jamesbornholt: This is the check that bails out if we're currently panicking. The idea is to try to return the lock to some sort of usable state before bailing out, so that code further along the unwinding path can re-acquire the lock.

We originally added this check to avoid trying to access `ExecutionState` during a panic. Now that I think about it, there might be some cleaner way to do this where we get rid of `should_stop` and have these drop handlers instead be more cautious about whether they can access `ExecutionState` (with a method like `ExecutionState::try_with` or something). I'll think about that some more, but in the meantime I think change this is still an improvement on the current situation.

jorajeev (approved):

PR #33, opened 2021-06-12
Implement SeqCst atomics

jamesbornholt:
This change adds support for the `std::sync::atomic` module.
Sequentially consistent atomics are easy for us to model---we get
sequential consistency for free by just remembering the current value of
each atomic variable. We don't have a way to model other relaxed atomic
behaviors though. This change chooses to treat *all* atomic operations
as if they were SeqCst, and emits a warning the first time it encounters
non-SeqCst atomic operations.

The implementation here is almost certainly too conservative about
inserting context switches between atomic operations. In the future, we
should convince ourselves about which context switches are necessary and
remove the unnecessary ones. That would be a breaking change since it
would invalidate previously recorded scehdules.

---

### DIFF
--- a/src/sync/atomic/mod.rs
+++ b/src/sync/atomic/mod.rs
@@ -1,0 +1,189 @@
+//! Atomic types
+//!
+//! Atomic types provide primitive shared-memory communication between threads, and are the building
+//! blocks of other concurrent types.
+//!
+//! This module defines atomic versions of a select number of primitive types, the same as
+//! [`std::sync::atomic`] in the standard library. See that module's documentation for more details.
+//!
+//! # Warning about relaxed behaviors
+//!
+//! Shuttle does not faithfully model behaviors of relaxed atomic orderings (those other than
+//! [`SeqCst`](Ordering::SeqCst)). Code that uses these orderings may contain bugs that Shuttle is
+//! unable to find if the bug requires the relaxed behavior to occur. **Shuttle models *all* atomic
+//! operations as if they were using SeqCst ordering.**
+//!
+//! For example, consider this test that uses a `flag` variable to indicate that data is present in
+//! a separate `data` variable:
+//! ```
+//! # use std::sync::Arc;
+//! # use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+//! # use std::thread;
+//! let flag = Arc::new(AtomicBool::new(false));
+//! let data = Arc::new(AtomicU64::new(0));
+//!
+//! {
+//!     let flag = Arc::clone(&flag);
+//!     let data = Arc::clone(&data);
+//!     thread::spawn(move|| {
+//!         data.store(42, Ordering::Relaxed);
+//!         flag.store(true, Ordering::Relaxed);
+//!     });
+//! }
+//!
+//! if flag.load(Ordering::Relaxed) {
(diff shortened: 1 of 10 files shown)


src/sync/atomic/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// Base type for atomic implementations. This type handles generating the right interleavings for
+/// all atomics. The interesting stuff is in `load`, `store`, `swap`, and `fetch_update`; all other
+/// atomic operations are implemented in terms of those four primitives.
+#[derive(Debug)]
+struct Atomic<T> {
+    inner: RefCell<T>,
+}
+
+unsafe impl<T: Sync> Sync for Atomic<T> {}
```
jorajeev, line 122: Add a comment saying why this is okay for Shuttle (as we've done in other places)?

src/sync/atomic/ptr.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+impl<T> std::fmt::Debug for AtomicPtr<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        std::fmt::Debug::fmt(unsafe { &self.raw_load() }, f)
+    }
+}
+
+unsafe impl<T> Send for AtomicPtr<T> {}
```
jorajeev, line 27: Comments here?

src/sync/atomic/ptr.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// containing the previous value. On success this value is guaranteed to be equal to
+    /// `current`.
+    pub fn compare_exchange(
+        &self,
+        current: *mut T,
+        new: *mut T,
+        success: Ordering,
+        failure: Ordering,
+    ) -> Result<*mut T, *mut T> {
+        self.fetch_update(success, failure, |val| (val == current).then(|| new))
```
jorajeev, line 94: Nice use of `fetch_update`

src/sync/atomic/ptr.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// containing the previous value. On success this value is guaranteed to be equal to
+    /// `current`.
+    pub fn compare_exchange_weak(
+        &self,
+        current: *mut T,
+        new: *mut T,
+        success: Ordering,
+        failure: Ordering,
+    ) -> Result<*mut T, *mut T> {
+        self.compare_exchange(current, new, success, failure)
```
jorajeev, line 110: The real `compare_and_exchange` is allowed to "spuriously fail" even if the comparison succeeds.  This one will not, which in theory means we are underapproximating here.  Not worth doing anything about, but maybe we should note that here.

src/sync/atomic/ptr.rs
jamesbornholt: This one's actually not a Shuttle-specific hack — it's to indicate that `AtomicPtr` makes a raw pointer Send + Sync even though it's still unsafe to dereference (the std implementation does the same manual impl). Added a comment to that effect.

src/sync/atomic/ptr.rs
jamesbornholt: Added a TODO, the same as we currently do for missing spurious failures in Condvar.

jorajeev (approved):

PR #34, opened 2021-06-16
Add option to not persist failures

jamesbornholt:
Sometimes we don't want to print failures (most notably when replaying a
schedule that we already know fails). This change adds an additional
setting for the existing `failure_persistence` config that disables
printing failing schedules entirely. It also changes the File variant of
that config to be an anonymous struct, since its meaning is pretty
clear.

---

### DIFF
--- a/tests/basic/replay.rs
+++ b/tests/basic/replay.rs
@@ -1,7 +1,8 @@
-use crate::{check_replay_roundtrip, check_replay_roundtrip_file};
+use crate::{check_replay_roundtrip, check_replay_roundtrip_file, Config, FailurePersistence};
 use shuttle::scheduler::{PctScheduler, ReplayScheduler, Schedule};
 use shuttle::sync::Mutex;
 use shuttle::{replay, thread, Runner};
+use std::panic;
 use std::sync::Arc;
 use test_env_log::test;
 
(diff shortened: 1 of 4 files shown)


tests/basic/replay.rs
```diff
@@ -133,3 +135,20 @@ fn replay_deadlock3_drop_mutex() {
...
+        let scheduler = PctScheduler::new(2, 100);
+        let mut config = Config::new();
+        config.failure_persistence = FailurePersistence::None;
+        let runner = Runner::new(scheduler, config);
+        runner.run(concurrent_increment_buggy);
+    })
+    .expect_err("test should panic");
+    let output = result.downcast::<String>().unwrap();
+    assert!(output.contains("counter is wrong"));
+    // All our current failure persistence modes print the word "schedule", so check that's missing
```
jorajeev, line 152: Was thinking we should record this assumption somewhere... though that's probably not a big deal as this test will catch a violation anyway.

jorajeev (approved):

tests/basic/replay.rs
jamesbornholt: Yeah, I was thinking the same but wasn’t really sure how to do it. Even  testing this at all seemed like overkill anyway, so I think we’re fine with it being potentially a bit fragile.

PR #37, opened 2021-06-26
Add function to get the number of context switches in the current execution

bkragl:
This is useful, for example, to record a concurrent history for linearizability
checking. The number of context switches essentially serves as timestamp.

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -156,6 +156,8 @@ pub(crate) struct ExecutionState {
     next_task: ScheduledTask,
     // whether the current task has asked to yield
     has_yielded: bool,
+    // the number of scheduling decisions made so far
+    context_switches: usize,
 
     scheduler: Rc<RefCell<dyn Scheduler>>,
     current_schedule: Schedule,
@@ -198,6 +200,7 @@ impl ExecutionState {
             current_task: ScheduledTask::None,
             next_task: ScheduledTask::None,
             has_yielded: false,
+            context_switches: 0,
             scheduler,
             current_schedule: initial_schedule,
             current_span_entered: None,
@@ -346,6 +349,10 @@ impl ExecutionState {
         self.tasks.get_mut(id.0).unwrap()
     }
 
+    pub(crate) fn context_switches() -> usize {
+        Self::with(|state| state.context_switches)
+    }
+
     /// Run the scheduler to choose the next task to run. `has_yielded` should be false if the
     /// scheduler is being invoked from within a running task, in which case `schedule` should not
     /// panic.
@@ -355,6 +362,8 @@ impl ExecutionState {
         if self.next_task != ScheduledTask::None {
             return;
         }
+
+        self.context_switches += 1;
 
         match self.config.max_steps {
             MaxSteps::None => {}
--- a/tests/basic/execution.rs
+++ b/tests/basic/execution.rs
@@ -1,5 +1,5 @@
 use shuttle::scheduler::RandomScheduler;
-use shuttle::{check, thread, Config, MaxSteps, Runner};
+use shuttle::{check, check_dfs, thread, Config, MaxSteps, Runner};
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use test_env_log::test;
 // Not actually trying to explore interleavings involving AtomicUsize, just using to smuggle a
@@ -26,9 +26,6 @@ fn basic_scheduler_test() {
 
 #[test]
 fn max_steps_none() {
-    use std::sync::atomic::{AtomicUsize, Ordering};
-    use std::sync::Arc;
-
     let counter = Arc::new(AtomicUsize::new(0));
     let counter_clone = Arc::clone(&counter);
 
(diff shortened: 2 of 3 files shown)


tests/basic/execution.rs
```diff
@@ -1,15 +1,17 @@
...
+use shuttle::sync::Mutex;
+use shuttle::{check, check_random, thread, Config, MaxSteps, Runner};
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use test_env_log::test;
-// Not actually trying to explore interleavings involving AtomicUsize, just using to smuggle a
-// mutable counter across threads
-use std::sync::atomic::{AtomicUsize, Ordering};
+// We use `std::sync::atomic::AtomicUsize` to smuggle a mutable counter across threads
+// without causing interleavings.
+use std::sync::atomic::{AtomicUsize as StdAtomicUsize, Ordering};
```
jamesbornholt, line 9: Instead of the awkward renaming here, I would just `use shuttle::sync::atomic::AtomicUsize` inside the tests where you want it.

tests/basic/execution.rs
```diff
@@ -169,3 +160,116 @@ fn max_steps_early_exit_scheduler() {
...
+}
+
+#[test]
+fn context_switches_atomic() {
+    check_random(
+        move || {
+            let mut threads = vec![];
+            // The actual atomic we test
+            let counter = Arc::new(AtomicUsize::new(0));
+            // Specification counters
```
jamesbornholt, line 177: These feel too closely tied to the implementation and so maybe not worth testing like this — I think I’d rather just put the math in a comment (like we do in the DFS tests).

tests/basic/execution.rs
```diff
@@ -169,3 +160,116 @@ fn max_steps_early_exit_scheduler() {
...
+
+#[test]
+#[should_panic]
+fn context_switches_outside_execution() {
+    shuttle::context_switches();
+}
+
+#[test]
+fn context_switches_atomic() {
+    check_random(
```
jamesbornholt, line 172: Would be nice to do a slightly smaller version of this test (spawn 2 threads instead of 3?) and use `check_dfs`.

tests/basic/execution.rs
```diff
@@ -169,3 +160,116 @@ fn max_steps_early_exit_scheduler() {
...
+                initial_context_switches + 2 * before_atomic_count + spawn_count + join_count + termination_count
+            );
+        },
+        100,
+    );
+}
+
+#[test]
+fn context_switches_mutex() {
+    check_random(
```
jamesbornholt, line 252: `check_dfs`

src/runtime/execution.rs
```diff
@@ -429,6 +434,16 @@ impl ExecutionState {
     }
 }
 
+/// The number of context switches that happened so far in the current Shuttle execution.
+///
+/// Note that this is the number of *possible* context switches, i.e., including times when the
+/// scheduler decided to continue with the same task.
+///
+/// Panics if called outside of a Shuttle execution.
+pub fn context_switches() -> usize {
```
jamesbornholt, line 443: Not super important, but I don’t really like bare functions like this. I’d consider implementing this as a method on `ExecutionState`, in the same vein as `me()`, and then put the public implementation in lib.rs that just calls that method.

tests/basic/execution.rs
bkragl: I don't find it that awkward. My thinking was to make it easy to see in the tests that a non-Shuttle atomic is used, in particular since a test I added uses both the Shuttle and the std `AtomicUsize`. But I'm fine with reverting this, since the test will probably change to only use the Shuttle atomic.

tests/basic/execution.rs
bkragl: Yeah, unfortunately doing something precise here is somewhat tied to the implementation. I definitely wanted to have an assertion in the threads (during the execution), not just at the end. What I could do is to get rid of the std counters and instead do this in the threads:

```rust
let count = counter.fetch_add(1, Ordering::SeqCst) + 1;

// We saw at least one context switch for each `fetch_add`.
assert!(shuttle::context_switches() >= count);

// We did not see the last context switch of this thread.
assert!(shuttle::context_switches() < 11);
```

The first assertion could be strengthened to `>= 2 + 2 * count` in the current implementation (initial context switch, spawn and first context switch per `fetch_add`, and second context switch for `fetch_add` of current thread). It would be more work to update the test when we optimize context switches, but if we assert the final expected number of context switches at the end, we would anyway have to update the test in that case. How much precision do you want?

tests/basic/execution.rs
bkragl: Sounds good.

tests/basic/execution.rs
bkragl: 👌

src/runtime/execution.rs
bkragl: Will do.

tests/basic/execution.rs
jamesbornholt: I like tying the assertion to the current value of the counter like that. I think strengthening it to `>= 2 + 2 * count` is also good as long as there's a comment explaining why that's the magic formula.

tests/basic/execution.rs
jamesbornholt: If you were going to use both in the same test, you could also just fully qualify the single use of `shuttle::sync::atomic::AtomicUsize::new(0)`.

jamesbornholt (approved):

PR #38, opened 2021-06-29
Add support for tracking thread causality.

jorajeev:
This change adds support for tracking causality among thread operations.
We do this in the standard way, by associating a vector clock with each thread.
The i'th element of a thread's vector clock denotes its knowledge of the clock
of thread i. Clocks are partially ordered using a pointwise ordering <.
The main property we want is that for any pair of events p, q:
    (p causally precedes q)  iff  (clock at p  <  clock at q).

We update the code for thread spawn and join, as well as the various synchronization
objects (Atomics, Barriers, CondVars, Mutexes, RwLocks and mpsc channels) to track
causality by updating vector clocks appropriately.

This change does not currently properly track causality for async interactions;
those will be done in a subsequent PR.

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -1,4 +1,5 @@
 use crate::runtime::failure::persist_failure;
+use crate::runtime::task::clock::VectorClock;
 use crate::runtime::task::{Task, TaskId, TaskState, DEFAULT_INLINE_TASKS};
 use crate::scheduler::{Schedule, Scheduler};
 use crate::{Config, MaxSteps};
@@ -57,7 +58,7 @@ impl Execution {
 
         EXECUTION_STATE.set(&state, move || {
             // Spawn `f` as the first task
-            ExecutionState::spawn_thread(f, config.stack_size, None);
+            ExecutionState::spawn_thread(f, config.stack_size, None, Some(VectorClock::new()));
 
             // Run the test to completion
             while self.step(config) {}
@@ -234,19 +235,33 @@ impl ExecutionState {
     {
         Self::with(|state| {
             let task_id = TaskId(state.tasks.len());
-            let task = Task::from_future(future, stack_size, task_id, name);
+            let clock = state.increment_clock_mut(); // Increment the parent's clock
+            clock.extend(task_id); // and extend it with an entry for the new task
+            let task = Task::from_future(future, stack_size, task_id, name, clock.clone());
             state.tasks.push(task);
             task_id
         })
     }
 
-    pub(crate) fn spawn_thread<F>(f: F, stack_size: usize, name: Option<String>) -> TaskId
+    pub(crate) fn spawn_thread<F>(
+        f: F,
+        stack_size: usize,
+        name: Option<String>,
+        mut initial_clock: Option<VectorClock>,
+    ) -> TaskId
     where
         F: FnOnce() + Send + 'static,
     {
         Self::with(|state| {
             let task_id = TaskId(state.tasks.len());
-            let task = Task::from_closure(f, stack_size, task_id, name);
+            let clock = if let Some(ref mut clock) = initial_clock {
+                clock
+            } else {
+                // Inherit the clock of the parent thread (which spawned this task)
+                state.increment_clock_mut()
+            };
+            clock.extend(task_id); // and extend it with an entry for the new thread
+            let task = Task::from_closure(f, stack_size, task_id, name, clock.clone());
             state.tasks.push(task);
             task_id
         })
@@ -351,6 +366,35 @@ impl ExecutionState {
 
     pub(crate) fn context_switches() -> usize {
         Self::with(|state| state.context_switches)
+    }
+
+    pub(crate) fn get_clock(&self, id: TaskId) -> &VectorClock {
+        &self.tasks.get(id.0).unwrap().clock
+    }
+
+    pub(crate) fn get_clock_mut(&mut self, id: TaskId) -> &mut VectorClock {
+        &mut self.tasks.get_mut(id.0).unwrap().clock
+    }
+
+    // Increment the current thread's clock entry and update its clock with the one provided.
+    pub(crate) fn update_clock(&mut self, clock: &VectorClock) {
+        let task = self.current_mut();
+        task.clock.increment(task.id);
+        task.clock.update(clock);
+    }
+
+    // Increment the current thread's clock and return a shared reference to it
+    pub(crate) fn increment_clock(&mut self) -> &VectorClock {
+        let task = self.current_mut();
+        task.clock.increment(task.id);
+        &task.clock
+    }
+
+    // Increment the current thread's clock and return a mutable reference to it
+    pub(crate) fn increment_clock_mut(&mut self) -> &mut VectorClock {
+        let task = self.current_mut();
+        task.clock.increment(task.id);
+        &mut task.clock
     }
 
     /// Run the scheduler to choose the next task to run. `has_yielded` should be false if the
--- a/src/runtime/task/clock.rs
+++ b/src/runtime/task/clock.rs
@@ -1,0 +1,132 @@
+use crate::runtime::task::{TaskId, DEFAULT_INLINE_TASKS};
+use smallvec::{smallvec, SmallVec};
+use std::cmp::{Ordering, PartialOrd};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct VectorClock {
+    pub(crate) time: SmallVec<[u32; DEFAULT_INLINE_TASKS]>,
+}
+
+impl VectorClock {
+    pub(crate) fn new() -> Self {
+        Self { time: SmallVec::new() }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new_from(v: &[u32]) -> Self {
+        Self {
+            time: SmallVec::from(v),
+        }
+    }
+
+    // Zero extend clock to accommodate `task_id` tasks.
+    pub(crate) fn extend(&mut self, task_id: TaskId) {
+        let num_new_tasks = 1 + task_id.0 - self.time.len();
+        let clock: SmallVec<[_; DEFAULT_INLINE_TASKS]> = smallvec![0u32; num_new_tasks];
+        self.time.extend_from_slice(&clock);
+    }
+
+    pub(crate) fn increment(&mut self, task_id: TaskId) {
+        self.time[task_id.0] += 1;
+    }
+
+    // Update the clock of `self` with the clock from `other`
+    pub(crate) fn update(&mut self, other: &Self) {
+        let n1 = self.time.len();
+        let n2 = other.time.len();
+        for i in 0..n1.min(n2) {
+            self.time[i] = self.time[i].max(other.time[i])
+        }
+        for i in n1..n2 {
+            // could be empty
+            self.time.push(other.time[i]);
+        }
+    }
+
+    pub fn get(&self, i: usize) -> u32 {
+        self.time[i]
+    }
+}
+
+impl std::ops::Deref for VectorClock {
+    type Target = [u32];
+    fn deref(&self) -> &Self::Target {
+        &self.time[..]
+    }
+}
+
+fn unify(a: Ordering, b: Ordering) -> Option<Ordering> {
+    use Ordering::*;
+
+    match (a, b) {
+        (Equal, Equal) => Some(Equal),
+        (Less, Greater) | (Greater, Less) => None,
+        (Less, _) | (_, Less) => Some(Less),
+        (Greater, _) | (_, Greater) => Some(Greater),
+    }
+}
+
+impl PartialOrd for VectorClock {
+    // Compare vector clocks
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let n1 = self.time.len();
+        let n2 = other.time.len();
+        // if (n1<n2), then other can't have happened before self, similarly for (n1>n2)
+        let mut ord = n1.cmp(&n2);
+        for i in 0..n1.min(n2) {
+            ord = unify(ord, self.time[i].cmp(&other.time[i]))?; // return if incomparable
+        }
+        Some(ord)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn vector_clock() {
+        let v1 = VectorClock::new_from(&[1, 2, 3, 4]);
+        let v2 = VectorClock::new_from(&[1, 2, 4, 5]);
+        let v3 = VectorClock::new_from(&[1, 2, 3, 1]);
+        let v4 = VectorClock::new_from(&[1, 2, 4, 1]);
+        let v5 = VectorClock::new_from(&[1, 2, 3, 4]);
+        assert!(v1 < v2 && v1 > v3 && v1 == v5);
+        assert!(v2 > v3 && v2 > v4);
+        assert!(v3 < v4);
+        assert_eq!(v1.partial_cmp(&v4), None);
+
+        let v1 = VectorClock::new_from(&[1, 2, 3, 4]);
+        let v2 = VectorClock::new_from(&[1, 2, 2]);
+        let v3 = VectorClock::new_from(&[1, 2, 3]);
+        let v4 = VectorClock::new_from(&[1, 2, 4]);
+        assert!(v1 > v2);
+        assert!(v1 > v3);
+        assert_eq!(v1.partial_cmp(&v4), None);
+
+        let v1 = VectorClock::new_from(&[]);
+        let v2 = VectorClock::new_from(&[1]);
+        assert!(v1 < v2);
+
+        let v1 = VectorClock::new_from(&[1, 2, 1]);
+        let v2 = VectorClock::new_from(&[1, 3]);
+        let v3 = VectorClock::new_from(&[1, 1, 1, 2]);
+        let v4 = VectorClock::new_from(&[1, 1, 2]);
+
+        let mut v = v1.clone();
+        v.update(&v2);
+        assert_eq!(v, VectorClock::new_from(&[1, 3, 1]));
+
+        let mut v = v1.clone();
+        v.update(&v3);
+        assert_eq!(v, VectorClock::new_from(&[1, 2, 1, 2]));
+
+        let mut v = v1.clone();
+        v.update(&v4);
+        assert_eq!(v, VectorClock::new_from(&[1, 2, 2]));
+
+        let mut v = v1.clone();
+        v.update(&VectorClock::new());
+        assert_eq!(v, v1);
+    }
+}
--- a/src/runtime/task/mod.rs
+++ b/src/runtime/task/mod.rs
@@ -1,4 +1,5 @@
 use crate::runtime::execution::ExecutionState;
+use crate::runtime::task::clock::VectorClock;
 use crate::runtime::thread;
 use crate::runtime::thread::continuation::{ContinuationPool, PooledContinuation};
 use bitvec::prelude::*;
@@ -9,6 +10,7 @@ use std::fmt::Debug;
 use std::rc::Rc;
 use std::task::Context;
 
+pub(crate) mod clock;
 pub(crate) mod waker;
 use waker::make_waker;
 
@@ -42,6 +44,8 @@ pub(crate) struct Task {
 
     pub(super) continuation: Rc<RefCell<PooledContinuation>>,
 
+    pub(crate) clock: VectorClock,
+
     waiter: Option<TaskId>,
 
     waker: Waker,
@@ -53,10 +57,18 @@ pub(crate) struct Task {
 
 impl Task {
     /// Create a task from a continuation
-    fn new<F>(f: F, stack_size: usize, id: TaskId, task_type: TaskType, name: Option<String>) -> Self
+    fn new<F>(
+        f: F,
+        stack_size: usize,
+        id: TaskId,
+        task_type: TaskType,
+        name: Option<String>,
+        clock: VectorClock,
+    ) -> Self
     where
         F: FnOnce() + Send + 'static,
     {
+        assert!(id.0 < clock.time.len());
         let mut continuation = ContinuationPool::acquire(stack_size);
         continuation.initialize(Box::new(f));
         let waker = make_waker(id);
@@ -66,6 +78,7 @@ impl Task {
             state: TaskState::Runnable,
             task_type,
             continuation,
+            clock,
             waiter: None,
             waker,
             woken_by_self: false,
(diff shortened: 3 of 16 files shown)


src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
...
+    fn from(v: VectorClock) -> SmallVec<[u32; DEFAULT_INLINE_TASKS]> {
+        v.time
+    }
+}
+
+impl<'a> From<&'a VectorClock> for &'a SmallVec<[u32; DEFAULT_INLINE_TASKS]> {
+    fn from(v: &'a VectorClock) -> &'a SmallVec<[u32; DEFAULT_INLINE_TASKS]> {
+        &v.time
+    }
+}
```
jamesbornholt, line 356: I think these are all unused. Even if they weren't, `new_from` covers the need for `From` impls since `SmallVec<[T; _]>` derefs to `[T]`, and if we wanted the `IntoIterator` we should just instead implement `Deref<Target = [u32]>` or something.

src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
         write!(f, "}}")
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct VectorClock {
+    pub(crate) time: SmallVec<[u32; DEFAULT_INLINE_TASKS]>,
```
jamesbornholt, line 293: Is it worth introducing a wrapper type for timestamps, `Timestamp(u32)` or something?

src/runtime/task/mod.rs
```diff
@@ -53,34 +59,49 @@ pub(crate) struct Task {
...
+        clock: VectorClock,
+    ) -> Self
     where
         F: FnOnce() + Send + 'static,
     {
         let mut continuation = ContinuationPool::acquire(stack_size);
         continuation.initialize(Box::new(f));
         let waker = make_waker(id);
         let continuation = Rc::new(RefCell::new(continuation));
+        assert!(id.0 < clock.time.len());
```
jamesbornholt, line 77: nit: probably move this assert to the top of the method body

src/sync/barrier.rs
```diff
@@ -64,22 +66,35 @@ impl Barrier {
...
+            // Unblock each waiter, updating its clock with the barrier's clock
+            let clock = state.clock.clone();
+            ExecutionState::with(|s| {
+                for tid in state.waiters.drain() {
+                    debug_assert_ne!(tid, me);
                     let t = s.get_mut(tid);
                     debug_assert!(t.blocked());
+                    t.clock.increment(tid);
+                    t.clock.update(&clock);
                     t.unblock();
```
jamesbornholt, line 91: I wonder if there's some way to make this API less error-prone by making `unblock` take a clock as an argument and do the required clock updates itself.

src/sync/condvar.rs
```diff
@@ -138,15 +141,16 @@ impl Condvar {
...
         match my_status {
-            CondvarWaitStatus::Broadcast => {
-                // Woken by a broadcast, so nothing to do
+            CondvarWaitStatus::Broadcast(clock) => {
+                // Woken by a broadcast, so nothing to do except update the clock
+                ExecutionState::with(|s| s.update_clock(&clock));
             }
             CondvarWaitStatus::Signal(mut epochs) => {
-                let epoch = epochs.pop_front().expect("should be a pending signal");
+                let (epoch, clock) = epochs.pop_front().expect("should be a pending signal");
```
jamesbornholt, line 149: Maybe we can think about this in a future PR, but: does having vector clocks subsume the need for epochs? We just want to know whether a thread can "see" a notify, and it seems like that's exactly what causality would tell us?

src/thread.rs
```diff
@@ -146,6 +159,19 @@ pub fn current() -> Thread {
...
+        let me = state.current();
+        state.get_clock(me.id()).clone()
+    })
+}
+
+/// Gets the clock for the thread with the given task_id
+pub fn get_clock(task_id: TaskId) -> VectorClock {
+    ExecutionState::with(|state| state.get_clock(task_id).clone())
+}
+
```
jamesbornholt, line 174: These aren't specific to threads (they'll work for futures too), and aren't part of the stdlib, so I feel like maybe they belong somewhere else. Bernhard recently added `shuttle::context_switches`, maybe we need to make some new module to contain all this "Shuttle metadata".

tests/basic/clocks.rs
```diff
@@ -0,0 +1,415 @@
...
+use shuttle::{check_dfs, check_pct, thread};
+use std::collections::HashSet;
+use std::sync::{atomic::Ordering, Arc};
+use test_env_log::test;
+
+pub fn me() -> usize {
+    usize::from(thread::current().id())
+}
+
+// TODO Maybe make this a macro so backtraces are more informative
```
jamesbornholt, line 15: Yeah, should do that

tests/basic/clocks.rs
```diff
@@ -0,0 +1,415 @@
+use shuttle::sync::{
+    atomic::AtomicBool,
+    mpsc::{channel, sync_channel},
+    Barrier, Condvar, Mutex, RwLock,
+};
```
jamesbornholt, line 5: nit:
```
use shuttle::sync::atomic::AtomcBool;
use shuttle::sync::mpsc::{channel, sync_channel};
use shuttle::sync::{Barrier, Condvar, Mutex, RwLock};
```

tests/basic/clocks.rs
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// TODO Maybe make this a macro so backtraces are more informative
+pub fn check_clock(f: impl Fn(usize, u32) -> bool) {
+    for (i, &c) in (&thread::my_clock()).into_iter().enumerate() {
+        assert!(f(i, c));
+    }
+}
+
+fn clock_mutex(num_threads: usize) {
```
jamesbornholt, line 22: wonder whether this file should have one submodule per sync primitive just to make the organization a bit clearer

tests/basic/clocks.rs
```diff
@@ -0,0 +1,415 @@
...
+            // But we know for sure that the thread that unblocked us should causally precede us.
+            check_clock(|i, c| (i != *guard) || (c > 0));
+        });
+    }
+}
+
+#[test]
+fn clock_condvar_notify_all_dfs() {
+    check_dfs(|| clock_condvar_notify_all(2), None);
+}
```
jamesbornholt, line 277: This test is really slow, is there any way to make it smaller?

Actually a bunch of the clock tests are really slow. Might be worth spending some time playing with the output of `cargo test -- -Zunstable-options --report-time` to pick off the slowest ones

tests/basic/clocks.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    {
+        let lock = Arc::clone(&lock);
+        let cond = Arc::clone(&cond);
+        thread::spawn(move || {
+            assert_eq!(me(), 1);
+            *lock.lock().unwrap() = me();
+            cond.notify_all();
+        });
+    }
```
jamesbornholt, line 256: Could move this after the waiters spawn and avoid spawning a new thread for it, which might make the DFS version of this test faster. But I guess we'd have to find a different way around the `me() == 1` thing.

tests/basic/clocks.rs
```diff
@@ -0,0 +1,415 @@
...
+    // As each thread joins, we get knowledge of its vector clock.
+    for handle in handles {
+        let k = handle.join().unwrap();
+        check_clock(move |i, c| (c > 0) == (i <= k));
+    }
+}
+
+#[test]
+fn clock_threads_dfs() {
+    check_dfs(|| clock_threads(3), None);
```
jamesbornholt, line 409: also really slow, maybe do 2 threads

tests/basic/mutex.rs
```diff
@@ -6,18 +6,21 @@ use test_env_log::test;
 
 #[test]
 fn basic_lock_test() {
```
jamesbornholt, line 8: is this change intentional?

tests/basic/pct.rs
```diff
@@ -191,8 +191,8 @@ fn figure1b_pct() {
 fn figure1b_pct_with_many_tasks() {
     // Spawn 48 busy threads, each taking 5 steps, plus 2 main threads with 10 steps, so k=260
     // n=50, k=260, d=2, so probability of finding the bug in one iteration is at least 1/(50*260)
-    // So probability of hitting the bug in 2000 iterations = 1 - (1 - 1/400)^2000 > 99.9%
-    let scheduler = PctScheduler::new(2, 2000);
+    // So probability of hitting the bug in 2000 iterations = 1 - (1 - 1/1300)^60_000 > 99.9%
+    let scheduler = PctScheduler::new(2, 60_000);
```
jamesbornholt, line 195: The comment is still out of sync with the code :-) ("2000 iterations", and 50*260 = 13,000)

More importantly, 60,000 iterations is a ton. Maybe we should drop the test to 20 threads (which is still > DEFAULT_INLINE_TASKS). Then the comment math works out as:
```
// Spawn 18 busy threads, each taking 5 steps, plus 2 main threads with 10 steps, so k=110
// n=20, k=110, d=2, so probability of finding the bug in one iteration is at least 1/(20*110)
// So probability of hitting the bug in 16,000 iterations = 1 - (1 - 1/1300)^16_000 > 99.9%
```

tests/basic/thread.rs
```diff
@@ -96,7 +96,6 @@ fn thread_identity() {
                     let me = thread::current();
                     assert_eq!(me.name(), Some("producer"));
                     let id = me.id();
-
```
jamesbornholt, line 99: ?

jamesbornholt (commented):
Super cool. A bunch of minor comments. I think my only high-level comment is whether there's a way we can make the API less clunky by doing the clock maintenance inside `block`/`unblock`?

src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
...
+        self.time[task_id.0] += 1;
+    }
+
+    // Update the clock of `self` with the clock from `other`
+    pub(crate) fn update(&mut self, other: &Self) {
+        let n1 = self.time.len();
+        let n2 = other.time.len();
+        for i in 0..std::cmp::min(n1, n2) {
+            self.time[i] = std::cmp::max(self.time[i], other.time[i])
+        }
```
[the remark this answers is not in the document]
jorajeev: Done

src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
         write!(f, "}}")
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct VectorClock {
```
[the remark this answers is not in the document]
jorajeev: Done.

src/runtime/execution.rs
```diff
@@ -234,19 +234,33 @@ impl ExecutionState {
...
         Self::with(|state| {
             let task_id = TaskId(state.tasks.len());
-            let task = Task::from_closure(f, stack_size, task_id, name);
+            let clock = if let Some(ref mut clock) = initial_clock {
+                clock
+            } else {
+                // Inherit the clock of the parent thread (which spawned this task)
+                state.increment_clock_mut()
+            };
+            clock.extend(task_id); // and extend it with an entry for the new thread
```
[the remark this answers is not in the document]
jorajeev: I think this is enough.  We update the parent's clock (incrementing its own entry and adding a 0-entry for the child), and this becomes the starting clock for the child.  The only causal knowledge the parent has about the child after a successful spawn is that the child started execution of its closure.  But the child's clock at that point is the same as the parent, and so the two are causally related.

src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
...
+
+    // Zero extend clock to accommodate `task_id` tasks.
+    pub(crate) fn extend(&mut self, task_id: TaskId) {
+        let d = 1 + task_id.0 - self.time.len();
+        self.time.extend_from_slice(vec![0u32; d].as_slice());
+    }
+
+    pub(crate) fn increment(&mut self, task_id: TaskId) {
+        assert!(task_id.0 < self.time.len());
+        self.time[task_id.0] += 1;
```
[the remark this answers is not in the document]
jorajeev: +1

src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
...
+    #[cfg(test)]
+    pub(crate) fn new_from(v: &[u32]) -> Self {
+        let time: SmallVec<[u32; DEFAULT_INLINE_TASKS]> = SmallVec::from(v);
+        Self { time }
+    }
+
+    // Zero extend clock to accommodate `task_id` tasks.
+    pub(crate) fn extend(&mut self, task_id: TaskId) {
+        let d = 1 + task_id.0 - self.time.len();
+        self.time.extend_from_slice(vec![0u32; d].as_slice());
```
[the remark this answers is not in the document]
jorajeev: Done.

src/runtime/execution.rs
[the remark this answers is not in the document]
jamesbornholt: What I was trying to ask (and not explaining very well) was whether we wanted to update the child’s view of its own clock to some non-zero value when first spawned. But it doesn’t matter because you can’t compare the clocks of different threads, so starting a threads own clock at 0 is indistinguishable from starting it at some non-zero value.

tests/basic/clocks.rs
jorajeev: Thing is that I want the assertion on line 264 that checks that the waiter does not have causal dependency on the notifier as long as the lock reads zero.  This is easy to do with the extra thread (with taskid 1), since we can just check that the vector clock for thread-1 must be zero.  If we use thread 0 to do the notify, we can't use the same check, because thread 0 spawns all the waiters, so their clocks inherit its clock.  I could save a thread spawn by making the main thread be one of the waiters, but that would result in some code duplication, and i thought we could just live with a smaller dfs test.

tests/basic/clocks.rs
jorajeev: Done.

tests/basic/pct.rs
jorajeev: Fixed :-)

tests/basic/thread.rs
jorajeev: Fixed.

jorajeev:
> Super cool. A bunch of minor comments. I think my only high-level comment is whether there's a way we can make the API less clunky by doing the clock maintenance inside `block`/`unblock`?

I think hijacking block/unblock won't work for the atomic registers, where the reader has to inherit the clock from the last writer, but there is no blocking.  We could try piggybacking clocks on yield's, but since we're likely to remove some yields in the future (since we know we have some redundant ones), I'm worried that change may break causality.  So I propose we live with this clunkiness for now, and revisit this later (maybe when we add causality to async).  I added a TODO in the spawn method to record this.

tests/basic/mutex.rs
jorajeev: Do you mean why did I change from `check` to `check_dfs` ?  I think I was trying out something in an early revision; can revert that back.

src/runtime/execution.rs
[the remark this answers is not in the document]
jorajeev: Yes, that was my thinking too.

src/runtime/task/mod.rs
jorajeev: Done.

src/runtime/task/mod.rs
```diff
@@ -265,3 +287,146 @@ impl Debug for TaskSet {
...
+        }
+        for i in n1..n2 {
+            // could be empty
+            self.time.push(other.time[i]);
+        }
+    }
+
+    pub fn get(&self, i: usize) -> u32 {
+        assert!(i < self.time.len());
+        self.time[i]
```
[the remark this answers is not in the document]
jorajeev: Removed.

tests/basic/mutex.rs
jamesbornholt: Yeah. We should make this change, but in a separate PR, because we can remove the round-robin scheduler at the same time.

src/sync/condvar.rs
jorajeev: I added a TODO to take a look.  We can also create an issue if we want to track it better.

jamesbornholt: Created #39 if you want to link it from the todo.

tests/basic/mutex.rs
jorajeev: Reverted back to the old version.

src/thread.rs
jorajeev: Since there are only 2 such functions for now, I moved them to `lib.rs` where `context_switches` lives.  We can create a module if we add more to the set?

src/runtime/task/mod.rs
jorajeev: Yes, much nicer; fixed.

src/runtime/task/mod.rs
jorajeev: Feels a bit overkill, since most of the time we're comparing and updating entire clocks.  We only look at individual timestamps in the VectorClock impl and in test code.

src/sync/barrier.rs
jorajeev: Problem is that there are places where we call unblock only in some situations.  Example: when a sender sends a message on an empty channel, if there's a waiting receiver, it unblocks the receiver.  If we passed in the sender clock to unblock, we'd update the receiver clock with it, which is fine.  But then the receiver code that removes the message will also update the clock from the message, which is redundant now. And we can't avoid updating the receiver with the clock, because that's needed in the case where the sender didn't wake the receiver.

Also in some places (e.g., Condvar::notify_one), we unblock a bunch of threads, but the first one blocks the others.  In this case, we don't want to update the clocks of anyone except the winner.

src/sync/condvar.rs
jorajeev: Done - thanks!

src/sync/barrier.rs
jamesbornholt: That makes sense. I think in practice these are likely to be pretty stable anyway, and new primitives are unlikely to fit neatly into some predefined clock pattern.

src/sync/atomic/mod.rs
```diff
@@ -172,18 +190,46 @@ impl<T: Copy + Eq> Atomic<T> {
...
         } else {
             Err(current)
         };
+        self.inhale_clock();
         thread::switch();
         ret
     }
 
     unsafe fn raw_load(&self) -> T {
+        self.exhale_clock();
```
jamesbornholt, line 207: I don't think we want to modify the clock here; the idea is that `raw_load` doesn't affect synchronization operations at all (so we can insert uses of it when debugging it to inspect an atomic's value without perturbing a replay). I realize now that I put a comment to that effect on `AtomicBool::raw_load` and friends but not on this method.

src/sync/atomic/mod.rs
```diff
@@ -159,6 +175,8 @@ impl<T: Copy + Eq> Atomic<T> {
         maybe_warn_about_ordering(order);
 
         thread::switch();
+        self.inhale_clock();
+        self.exhale_clock();
```
jamesbornholt, line 179: Why do these happen in the opposite order to in `fetch_update`?

src/runtime/task/mod.rs
```diff
@@ -1,14 +1,18 @@
 use crate::runtime::execution::ExecutionState;
+use crate::runtime::task::clock::VectorClock;
 use crate::runtime::thread;
 use crate::runtime::thread::continuation::{ContinuationPool, PooledContinuation};
 use bitvec::prelude::*;
 use bitvec::vec::BitVec;
 use futures::{task::Waker, Future};
 use std::cell::RefCell;
+use std::cmp::PartialOrd;
```
jamesbornholt, line 9: nit: remove

src/runtime/task/mod.rs
```diff
@@ -1,14 +1,18 @@
...
 use std::fmt::Debug;
+use std::iter::Iterator;
```
jamesbornholt, line 11: nit: remove

src/runtime/task/clock.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl PartialOrd for VectorClock {
+    // Compare vector clocks
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let n1 = self.time.len();
+        let n2 = other.time.len();
+        // if (n1<n2), then other can't have happened before self, similarly for (n1>n2)
+        let mut ord = n1.cmp(&n2);
+        for i in 0..std::cmp::min(n1, n2) {
```
jamesbornholt, line 75:
```suggestion
        for i in 0..n1.min(n2) {
```

src/runtime/task/clock.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl VectorClock {
+    pub(crate) fn new() -> Self {
+        Self { time: SmallVec::new() }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new_from(v: &[u32]) -> Self {
+        let time: SmallVec<[u32; DEFAULT_INLINE_TASKS]> = SmallVec::from(v);
+        Self { time }
```
jamesbornholt, line 18:
```suggestion
        Self { time: SmallVec::from(v) }
```

src/runtime/task/clock.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    // Zero extend clock to accommodate `task_id` tasks.
+    pub(crate) fn extend(&mut self, task_id: TaskId) {
+        let d = 1 + task_id.0 - self.time.len();
```
jamesbornholt, line 23: nit: can we pick a better variable name than `d`? :-)

src/thread.rs
```diff
@@ -145,7 +158,6 @@ pub fn current() -> Thread {
         name,
     }
 }
-
```
jamesbornholt, line 148: mega nit: don't remove this blank line

jamesbornholt (commented):
A few more nits.

src/sync/atomic/mod.rs
jorajeev: Ah, got it.  Removed.

src/sync/atomic/mod.rs
jorajeev: Nice catch.  It should be the other way around.  I also realized fetch_update was wrong -- it was not handling the error case properly. :-(

Fixed both and added comments to explain.

src/runtime/task/mod.rs
jorajeev: Fixed.  Why doesn't clippy complain?

src/runtime/task/mod.rs
jorajeev: Fixed.

src/runtime/task/mod.rs
jorajeev: Ah, because it's test code.  Guess I should run it with --cfg test.

src/runtime/task/clock.rs
jorajeev: Missed that; fixed.

src/runtime/task/clock.rs
jorajeev: Ah, yes, thanks.

src/runtime/task/clock.rs
jorajeev: How about `d1` ? ;-)

jorajeev: Renamed to `num_new_tasks`

src/thread.rs
jorajeev: Didn't mean to; restored.

src/sync/atomic/mod.rs
jorajeev: Also added a new test `clock_fetch_update` to reveal the fetch_update bug.  For the other one, I think the way I had it wasn't a bug, but reordering the steps is more consistent with translating the swap to load/store.

src/runtime/task/mod.rs
jorajeev: For the record, it's `cargo clippy --tests`.  TIL.

jamesbornholt (approved):

PR #40, opened 2021-07-06
Print a failing schedule even during double panics

jamesbornholt:
Double panics are a common problem in concurrent Rust code. For example,
a thread might panic while holding a lock, and then need to acquire that
lock again during stack unwinding, but the lock is already poisoned. In
these cases, Shuttle currently can't print the schedule that led to the
original panic, because the double panic aborts the process before our
`catch_unwind` has a chance to run.

This change adds a new panic hook that gets a chance to run before the
double panic happens. We use this hook to print the schedule, so that
even if we double panic in future at least the user gets some output
they can use to reproduce the problem. There's some trickiness here
(outlined in a module comment for `failure.rs`) around running different
panic handlers at different times, which makes the code a little more
complex than I would have hoped, but it gets the job done.

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -1,6 +1,7 @@
-use crate::runtime::failure::persist_failure;
+use crate::runtime::failure::{init_panic_hook, persist_failure, persist_task_failure};
 use crate::runtime::task::clock::VectorClock;
 use crate::runtime::task::{Task, TaskId, TaskState, DEFAULT_INLINE_TASKS};
+use crate::runtime::thread::continuation::PooledContinuation;
 use crate::scheduler::{Schedule, Scheduler};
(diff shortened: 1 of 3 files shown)


jamesbornholt:
> Can we grab the schedule generated by the test, and add another (also ignored) test that replays it? Just to make sure, for instance, that the printed schedule is really complete?

The schedule actually won't be complete in this double panic scenario, as it'll miss whatever steps were taken during unwinding. I don't think there's a good way to fix that; we'd have to intercept the second panic somehow. Instead, what happens is that the replay successfully makes it to the first panic, which it reproduces, and then panics again because the schedule ends early. Same double-panic effect, just that the second panic is different.

jorajeev (approved):

PR #42, opened 2021-07-12
Give a better error message when ExecutionState is inaccessible

jamesbornholt:
This happens most often if you try to access a Shuttle primitive from
outside a Shuttle test. Rather than spewing an opaque error message
about a thread local not being set, let's give a somewhat meaningful
error message about what probably went wrong.

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -227,11 +227,12 @@ impl ExecutionState {
     where
         F: FnOnce(&mut ExecutionState) -> T,
     {
-        EXECUTION_STATE.with(|cell| f(&mut *cell.borrow_mut()))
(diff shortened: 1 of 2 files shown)


jorajeev (approved):

src/runtime/execution.rs
```diff
@@ -227,11 +227,12 @@ impl ExecutionState {
     where
         F: FnOnce(&mut ExecutionState) -> T,
     {
-        EXECUTION_STATE.with(|cell| f(&mut *cell.borrow_mut()))
+        Self::try_with(f).expect("Shuttle internal error: cannot access ExecutionState. are you trying to access a Shuttle primitive from outside a Shuttle test?")
```
bkragl, line 230: I'm late to the party for this nit, but `are` at the beginning of the second sentence should be capitalized.

PR #43, opened 2021-07-13
Add LocalKey and thread local storage

jamesbornholt:
Thread local storage allows threads to create slots using `LocalKey`,
which is instantiated using the `thread_local` macro. The objects in
these slots have their destructors run when the corresponding thread
exits.

Rust's TLS implementation leaves a lot of decisions either "best effort"
or platform-defined, so we have some degrees of freedom here. We model
something roughly like the fast-path Unix implementation: we allow TLS
destructors to initialize other TLS slots, but prevent re-initialization
of any TLS slot that has already been destroyed. We do run TLS
destructors on the main thread before it exits.

This implementation is based on Loom's TLS implementation, but we have a
little richer support for destructors (shown in the `drop_init` test).

---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -394,3 +394,44 @@ pub fn my_clock() -> crate::runtime::task::clock::VectorClock {
 pub fn get_clock(task_id: crate::runtime::task::TaskId) -> crate::runtime::task::clock::VectorClock {
     crate::runtime::execution::ExecutionState::with(|state| state.get_clock(task_id).clone())
 }
+
+/// Declare a new thread local storage key of type [`LocalKey`](crate::thread::LocalKey).
+#[macro_export]
+macro_rules! thread_local {
+    // empty (base case for the recursion)
+    () => {};
+
+    // process multiple declarations with a const initializer
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty = const { $init:expr }; $($rest:tt)*) => (
+        $crate::__thread_local_inner!($(#[$attr])* $vis $name, $t, $init);
+        $crate::thread_local!($($rest)*);
+    );
+
+    // handle a single declaration with a const initializer
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty = const { $init:expr }) => (
+        $crate::__thread_local_inner!($(#[$attr])* $vis $name, $t, $init);
+    );
+
+    // process multiple declarations
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty = $init:expr; $($rest:tt)*) => (
+        $crate::__thread_local_inner!($(#[$attr])* $vis $name, $t, $init);
+        $crate::thread_local!($($rest)*);
+    );
+
+    // handle a single declaration
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty = $init:expr) => (
+        $crate::__thread_local_inner!($(#[$attr])* $vis $name, $t, $init);
+    );
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __thread_local_inner {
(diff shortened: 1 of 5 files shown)


JacobVanGeffen (approved):
LGTM

src/runtime/task/mod.rs
```diff
@@ -285,3 +324,55 @@ impl Debug for TaskSet {
...
+                        .downcast_ref::<T>()
+                        .expect("local value must downcast to expected type"))
+                })
+                .unwrap_or(Err(AlreadyDestructedError))
+        })
+    }
+
+    fn init<T: 'static>(&mut self, key: &'static LocalKey<T>, value: T) {
+        let key = LocalKeyId::new(key);
+        assert!(self.0.get(&key).is_none(), "cannot reinitialize a TLS slot");
```
jorajeev, line 361: Instead of doing a `get` followed by an `insert`, use the `entry()` api instead?

src/runtime/task/mod.rs
```diff
@@ -285,3 +324,55 @@ impl Debug for TaskSet {
...
+
+impl LocalKeyId {
+    fn new<T: 'static>(key: &'static LocalKey<T>) -> Self {
+        Self(key as *const _ as usize)
+    }
+}
+
+/// A map of thread-local storage values. Values are Option<_> because we need to be able to
+/// iteratively destruct them, as it's valid for TLS destructors to initialize new TLS slots.
+struct LocalMap(HashMap<LocalKeyId, Option<Box<dyn Any>>>);
```
jorajeev, line 340: Since `LocalKeyId` can be ordered (being a `usize`), how about using `BTreeMap` instead?  That way the drop order will be consistent across runs, in case we need reproducibility?

tests/basic/thread.rs
```diff
@@ -115,3 +115,236 @@ fn thread_identity() {
...
+    fn nested_with() {
+        shuttle::thread_local! {
+            static LOCAL1: RefCell<usize> = RefCell::new(1);
+            static LOCAL2: RefCell<usize> = RefCell::new(2);
+        }
+
+        check_dfs(
+            || {
+                LOCAL1.with(|local1| *local1.borrow_mut() = LOCAL2.with(|local2| *local2.borrow()));
+            },
```
jorajeev, line 175: Add a block here that reads `LOCAL` and asserts that its value is now 2 ?

src/lib.rs
```diff
@@ -394,3 +394,42 @@ pub fn my_clock() -> crate::runtime::task::clock::VectorClock {
...
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty = const { $init:expr }; $($rest:tt)*) => (
+        $crate::__thread_local_inner!($(#[$attr])* $vis $name, $t, $init);
+        $crate::thread_local!($($rest)*);
+    );
+
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty = const { $init:expr }) => (
+        $crate::__thread_local_inner!($(#[$attr])* $vis $name, $t, $init);
+    );
+
+    // process multiple declarations
```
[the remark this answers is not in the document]
jamesbornholt: The macro is copied and pasted from [Rust's implementation](https://github.com/rust-lang/rust/blob/59d92bd0173fb0436b93aa56ff2cb288d4717701/library/std/src/thread/local.rs#L132), but yeah, I'll add the comments.

src/runtime/task/mod.rs
jamesbornholt: Deterministic drop order is a really good point. Deriving `Ord` on `LocalKeyId` isn't quite enough because its values are computed from pointers, which aren't guaranteed to be deterministic across runs. I ended up just biting the bullet and making `pop` return slots in insertion order.

jorajeev (approved):

PR #45, opened 2021-07-23
Remove `Clone` bound from `PortfolioRunner::run`

bkragl:
`Runner::run` already needs to construct an `Arc` of the function to test. Should we avoid the construction of `Arc<Arc<_>>` via the `PortfolioRunner`? E.g., by creating a helper run function in `Runner` that can take an existing `Arc`.

---


jamesbornholt (approved):
I don't think avoiding the double-`Arc` is worth the effort, since it only happens once per `Runner` rather than once per test iteration or something.

PR #49, opened 2021-08-19
Implement `Once` and generalize thread-local storage map

jamesbornholt:
`Once` is a synchronization primitive often used in a static variable to
ensure some initialization logic runs at most once in the execution of a
program. To ensure that we support this pattern, we implement a new
global storage map that is empied across executions, and use it to
ensure that a `Once` cell's state resets across test executions. This
also means that multiple Shuttle tests running concurrently (e.g., under
the cargo test runner with a default config) will see independent
instances of a static Once cell and be properly isolated from each
other.

The storage map shares common logic with the TLS map we already have, so
this PR factors that map out into a single abstraction.

In a future PR we'll use Once to implement lazy_static, which will also
store its values into the global storage map.

---

### DIFF
--- a/src/sync/once.rs
+++ b/src/sync/once.rs
@@ -1,0 +1,170 @@
+use crate::runtime::execution::ExecutionState;
+use crate::runtime::storage::StorageKey;
+use crate::runtime::task::clock::VectorClock;
+use crate::sync::Mutex;
+use std::cell::RefCell;
+use std::rc::Rc;
+use tracing::trace;
+
+/// A synchronization primitive which can be used to run a one-time global initialization. Useful
+/// for one-time initialization for FFI or related functionality. This type can only be constructed
+/// with [`Once::new()`].
+#[derive(Debug)]
+pub struct Once {
+    // We use the address of the `Once` as an identifier, so it can't be zero-sized even though all
+    // its state is stored in ExecutionState storage
+    _dummy: usize,
+}
+
+/// A `Once` cell can either be `Running`, in which case a `Mutex` mediates racing threads trying to
+/// invoke `call_once`, or `Complete` once an initializer has completed, in which case the `Mutex`
+/// is no longer necessary.
+enum OnceInitState {
+    Running(Rc<Mutex<bool>>),
+    Complete(VectorClock),
+}
+
+impl std::fmt::Debug for OnceInitState {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Self::Running(_) => write!(f, "Running"),
+            Self::Complete(_) => write!(f, "Complete"),
+        }
+    }
+}
+
+impl Once {
+    /// Creates a new `Once` value.
+    pub const fn new() -> Self {
+        Self { _dummy: 0 }
+    }
+
+    /// Performs an initialization routine once and only once. The given closure will be executed
+    /// if this is the first time `call_once` has been called, and otherwise the routine will *not*
+    /// be invoked.
+    ///
+    /// This method will block the calling thread if another initialization routine is currently
+    /// running.
+    ///
+    /// When this function returns, it is guaranteed that some initialization has run and completed
+    /// (it may not be the closure specified).
+    pub fn call_once<F>(&self, f: F)
+    where
+        F: FnOnce(),
+    {
+        self.call_once_inner(|_state| f(), false);
+    }
+
+    /// Performs the same function as [`Once::call_once()`] except ignores poisoning.
+    ///
+    /// If the cell has previously been poisoned, this function will still attempt to call the given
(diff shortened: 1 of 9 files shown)


src/sync/once.rs
```diff
@@ -0,0 +1,147 @@
...
+        // If there's a lock, then we need to try racing on it to decide who gets to run their
+        // initialization closure.
+        if let Some(lock) = lock {
+            let mut flag = lock.lock().unwrap();
+            if *flag {
+                return;
+            }
+
+            trace!("won the call_once race for cell {:p}", self);
+            f();
```
jorajeev, line 85: The doc for `Once` says that if `f` panics here, the `Once` instance gets poisoned.  In this version, Shuttle will just report a panic?  Could you add a test for the "f panics" case?

src/sync/once.rs
```diff
@@ -0,0 +1,147 @@
...
+    // TODO we don't currently implement poisoning at all, so this just calls `call_once`
+    pub fn call_once_force<F>(&self, f: F)
+    where
+        F: FnOnce(&OnceState),
+    {
+        self.call_once(|| f(&OnceState));
+    }
+
+    /// Returns `true` if some [`Once::call_once()`] call has completed successfully.
+    pub fn is_completed(&self) -> bool {
```
jorajeev, line 111: In the case where the state is `OnceInitState::Complete(C)`, you need to update the clock of the caller with `C`.

tests/basic/clocks.rs
```diff
@@ -437,3 +440,37 @@ fn clock_fetch_update() {
...
+        .map(|_| {
+            let once = Arc::clone(&once);
+            let init = Arc::clone(&init);
+            thread::spawn(move || {
+                check_clock(|i, c| (c > 0) == (i == 0));
+                once.call_once(|| init.store(me(), std::sync::atomic::Ordering::SeqCst));
+                let who_inited = init.load(std::sync::atomic::Ordering::SeqCst);
+                // should have inhaled the clock of the thread that inited the Once, but might also
+                // have inhaled the clocks of threads that we were racing with for initialization
+                check_clock(|i, c| i != who_inited || c > 0);
```
jorajeev, line 458: I think this can be a stronger check.  You can assert that the clocks of the following threads are nonzero: {0, who_inited, me}, i.e.,
`check_clock(|i, c| !(i == who_inited || i == 0 || i == me()) || c>0)`

tests/basic/once.rs
```diff
@@ -0,0 +1,208 @@
...
+            });
+            O2.call_once(|| {
+                counter.fetch_add(8, Ordering::SeqCst);
+            });
+
+            thd.join().unwrap();
+
+            let counter = counter.load(Ordering::SeqCst);
+            // lower two bits for C1, upper two bits for C2
+            assert_ne!(counter & (1 + 2), 0);
```
jorajeev, line 149: Also add `assert_ne!(counter & (1+2), 3)` ?  Only one initializer runs, etc?

src/sync/once.rs
jamesbornholt: I ended up just implementing poisoning -- it's actually really easy because our `Once` uses a `Mutex` under the hood, and `Mutex` already supports poisoning.

tests/basic/clocks.rs
```diff
@@ -437,3 +440,37 @@ fn clock_fetch_update() {
...
+    let threads = (0..num_threads)
+        .map(|_| {
+            let once = Arc::clone(&once);
+            let init = Arc::clone(&init);
+            thread::spawn(move || {
+                check_clock(|i, c| (c > 0) == (i == 0));
+                once.call_once(|| init.store(me(), std::sync::atomic::Ordering::SeqCst));
+                let who_inited = init.load(std::sync::atomic::Ordering::SeqCst);
+                // should have inhaled the clock of the thread that inited the Once, but might also
+                // have inhaled the clocks of threads that we were racing with for initialization
```
jorajeev, line 457: I don't understand the second part of this comment.  We are asserting here that this thread's clock is nonzero (causally depends on) only for (1) the thread that inited the Once, (2) itself, (3) the main thread.
Maybe just remove the second part of the comment?

jorajeev (changes requested):

tests/basic/clocks.rs
jamesbornholt: We're asserting that the clock is non-zero for *at least* those three threads; we're saying nothing about the others. The comment is just to say why the check isn't stronger -- we might have learned something about threads other than those three depending on whether they raced for initialization or not.

jorajeev: Ah, sorry.  For some reason I was thinking of an equivalence; but it's just an implication.

jorajeev (approved):

PR #50, opened 2021-08-19
Implement `Drop` for async task `JoinHandle`

JacobVanGeffen:
When an async task's `JoinHandle` is dropped, we mark the task as dropped. Dropped tasks are not allowed to force a deadlock, meaning that execution can end successfully when all tasks are either finished or dropped. At the same time, dropped tasks should still be runnable by the executor so long as some non-dropped task has not finished.

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -95,15 +95,18 @@ impl Execution {
                     let task_states = state
                         .tasks
                         .iter()
-                        .map(|t| (t.id, t.state))
+                        .map(|t| (t.id, t.state, t.detached))
                         .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
-                    if task_states.iter().any(|(_, s)| *s == TaskState::Blocked) {
+                    if task_states
+                        .iter()
+                        .any(|(_, s, detached)| !detached && *s == TaskState::Blocked)
+                    {
                         NextStep::Failure(
                             format!("deadlock! runnable tasks: {:?}", task_states),
                             state.current_schedule.clone(),
                         )
                     } else {
-                        debug_assert!(state.tasks.iter().all(|t| t.finished()));
+                        debug_assert!(state.tasks.iter().all(|t| t.detached || t.finished()));
                         NextStep::Finished
                     }
                 }
@@ -309,7 +312,7 @@ impl ExecutionState {
 
         for task in tasks.drain(..) {
             assert!(
-                final_state == ScheduledTask::Stopped || task.finished(),
+                final_state == ScheduledTask::Stopped || task.finished() || task.detached,
                 "execution finished but task is not"
             );
             Rc::try_unwrap(task.continuation)
@@ -319,6 +322,11 @@ impl ExecutionState {
 
         #[cfg(debug_assertions)]
         Self::with(|state| state.has_cleaned_up = true);
+    }
+
+    /// Determine whether the execution has finished.
+    pub(crate) fn is_finished(&self) -> bool {
+        self.current_task == ScheduledTask::Stopped || self.current_task == ScheduledTask::Finished
     }
 
     /// Invoke the scheduler to decide which task to schedule next. Returns true if the chosen task
@@ -477,14 +485,21 @@ impl ExecutionState {
             _ => {}
         }
 
+        let mut blocked_attached = false;
         let runnable = self
             .tasks
             .iter()
+            .inspect(|t| blocked_attached = blocked_attached || (t.blocked() && !t.detached))
             .filter(|t| t.runnable())
             .map(|t| t.id)
             .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
 
-        if runnable.is_empty() {
+        // We should finish execution when either
+        // (1) There are no runnable tasks, or
+        // (2) All runnable tasks have been detached AND there are no blocked attached tasks
+        // If there are some blocked attached tasks and all runnable tasks are detached,
+        // we must run some detached task so that blocked attached tasks may become unblocked.
+        if runnable.is_empty() || (!blocked_attached && runnable.iter().all(|id| self.get(*id).detached)) {
             self.next_task = ScheduledTask::Finished;
             return Ok(());
         }
--- a/src/runtime/task/waker.rs
+++ b/src/runtime/task/waker.rs
@@ -28,6 +28,10 @@ unsafe fn raw_waker_clone(data: *const ()) -> RawWaker {
 unsafe fn raw_waker_wake(data: *const ()) {
     let task_id = TaskId::from(data as usize);
     ExecutionState::with(|state| {
+        if state.is_finished() {
+            return;
+        }
+
         let waiter = state.get_mut(task_id);
 
         if waiter.finished() {
--- a/tests/asynch/basic.rs
+++ b/tests/asynch/basic.rs
@@ -1,3 +1,4 @@
+use futures::try_join;
 use shuttle::sync::Mutex;
 use shuttle::{asynch, check_dfs, check_random, scheduler::PctScheduler, thread, Runner};
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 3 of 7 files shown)


src/runtime/execution.rs
```diff
@@ -307,18 +308,24 @@ impl ExecutionState {
             (std::mem::replace(&mut state.tasks, SmallVec::new()), state.current_task)
         });
 
+        // NOTE: This needs to be set before tasks are dropped
+        Self::with(|state| state.has_cleaned_up = true);
```
jamesbornholt, line 312: I think rather than abusing this field we should just factor out a method like
```
pub(crate) fn is_finished() -> bool {
    state.current_task == ScheduledTask::Stopped || state.current_task == ScheduledTask::Finished
}
```

src/runtime/execution.rs
```diff
@@ -484,7 +491,9 @@ impl ExecutionState {
             .map(|t| t.id)
             .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
 
-        if runnable.is_empty() {
+        let any_blocked = self.tasks.iter().any(|t| t.blocked() && !t.dropped);
+
+        if runnable.is_empty() || (!any_blocked && runnable.iter().all(|id| self.get(*id).dropped)) {
```
jamesbornholt, line 496: I think this logic deserves a comment, it's a bit tricky to follow the cases that count as "Finished" now.

Also, to avoid iterating over `self.tasks` twice, we could probably do something like
```
let mut blocked_attached = false;
let runnable = self
    .tasks
    .iter()
    .inspect(|t| blocked_attached = blocked_attached || (t.blocked() && !t.dropped))
    .filter(|t| t.runnable())
    .map(|t| t.id)
    .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
```

tests/asynch/basic.rs
```diff
@@ -1,9 +1,103 @@
+use futures::try_join;
 use shuttle::sync::Mutex;
 use shuttle::{asynch, check_dfs, check_random, scheduler::PctScheduler, thread, Runner};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use test_env_log::test;
 
+async fn do_err(e: bool) -> Result<(), ()> {
```
jamesbornholt, line 8: nitpick: new code goes at the end of the file

tests/asynch/basic.rs
```diff
@@ -1,9 +1,103 @@
...
+    if e {
+        Err(())
+    } else {
+        Ok(())
+    }
+}
+
+// Check that try_join on shuttle futures works as expected
```
jamesbornholt, line 16: `Shuttle`

tests/asynch/basic.rs
```diff
@@ -1,9 +1,103 @@
...
+    );
+}
+
+// Check that when we block on a future, it gets executed in both executions
+// (block_on task finishing first and main task finishing first).
+#[test]
+fn block_shuttle_future() {
+    let orderings_orig = Arc::new(AtomicUsize::new(0));
+    let async_accesses_orig = Arc::new(AtomicUsize::new(0));
+    let async_accesses = async_accesses_orig.clone();
```
jamesbornholt, line 36: nit: make the order of these declarations consistent, and we usually use `foo` and `foo_clone`

tests/asynch/basic.rs
```diff
@@ -1,9 +1,103 @@
...
+            let f2 = do_err(true);
+            let f1 = do_err(false);
+            let res = asynch::block_on(async { try_join!(f1, f2) });
+            assert!(res.is_err());
+        },
+        None,
+    );
+}
+
+// Check that when we block on a future, it gets executed in both executions
```
jamesbornholt, line 30: I don't see how this test actually checks the *ordering* -- it just checks that the test ran twice and the blocked-on future ran twice.

tests/asynch/basic.rs
```diff
@@ -1,9 +1,103 @@
...
+            });
+        },
+        None,
+    );
+
+    assert_eq!(2, orderings_orig.load(Ordering::SeqCst));
+    assert_eq!(2, async_accesses_orig.load(Ordering::SeqCst));
+}
+
+// Check that when we spawn and drop a future, it gets executed exactly one of two possible executions.
```
jamesbornholt, line 54: maybe
```
// Check that a task may not run if its JoinHandle is dropped
```

tests/asynch/basic.rs
```diff
@@ -1,9 +1,103 @@
...
+            });
+        },
+        None,
+    );
+
+    assert_eq!(2, orderings_orig.load(Ordering::SeqCst));
+    assert_eq!(1, async_accesses_orig.load(Ordering::SeqCst));
+}
+
+// Check that when we spawn and drop a future that immediately yields, it gets executed exactly one of two possible executions.
```
jamesbornholt, line 77: maybe
```
// Same as `drop_shuttle_future`, but the inner task yields first, and might be cancelled part way through
``` 

and then it would be good to actually check that the "part way through" part is true -- there should be two executions that reach the `yield_now()`, but only one execution that reaches beyond there.

tests/asynch/channel.rs
```diff
@@ -12,19 +12,41 @@ use std::collections::HashSet;
...
+            let send_task = asynch::spawn(async move {
                 tx.send(42u32).unwrap();
             });
 
             asynch::block_on(async {
                 let x = rx.await.unwrap();
                 assert_eq!(x, 42u32);
             });
+
+            // Don't drop send_task (causing it to finish)
```
jamesbornholt, line 29: Not sure I follow why we have to do this. In the original version:
```
asynch::spawn(async move {
    tx.send(42u32).unwrap();
});

asynch::block_on(async {
    let x = rx.await.unwrap();
    assert_eq!(x, 42u32);
});
```
the sender handle gets dropped before `block_on`, but once we reach the `block_on`, we should be able to run the (now detached) sender task. So that test should work fine -- there's no deadlock because the sender task is still runnable.

jamesbornholt:
A few notes, mostly about comments. We do also have to think about how this interacts with our somewhat sketchy story around cleaning up incomplete tasks, but we can do that seperately.

src/runtime/task/waker.rs
```diff
@@ -28,6 +28,10 @@ unsafe fn raw_waker_clone(data: *const ()) -> RawWaker {
 unsafe fn raw_waker_wake(data: *const ()) {
     let task_id = TaskId::from(data as usize);
     ExecutionState::with(|state| {
+        if state.has_cleaned_up() {
```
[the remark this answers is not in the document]
JacobVanGeffen: Yes, but specifically because dropping a channel from the futures crate can call a waker. In general though, a waker could be called by a drop implementation of any library afaik.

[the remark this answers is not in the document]
JacobVanGeffen: If you comment out this if statement, the test `asynch::channel::oneshot_once` will fail. In the stack trace, you can see where `futures::channel::oneshot::drop_tx` calls the waker.

tests/asynch/basic.rs
JacobVanGeffen: Do you think it would be better if the test showed that in exactly one of the two orderings, the blocked on task finishes first?

tests/asynch/channel.rs
JacobVanGeffen: Yep totally right.

tests/asynch/basic.rs
JacobVanGeffen: Oh yeah, I like this idea.

src/runtime/execution.rs
JacobVanGeffen: Yeah it's a bit tricky. I'll add a comment.

src/runtime/execution.rs
```diff
@@ -321,6 +324,12 @@ impl ExecutionState {
         Self::with(|state| state.has_cleaned_up = true);
     }
 
+    /// Determine whether the execution has finished. This is used in the Drop
+    /// implementation of JoinHandle to determine if the task should be marked as detached.
```
jamesbornholt, line 328: nit: I'd just drop the second sentence from this comment

src/runtime/execution.rs
```diff
@@ -477,14 +486,21 @@ impl ExecutionState {
...
             .map(|t| t.id)
             .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
 
-        if runnable.is_empty() {
+        // We should finish execution when either
+        // (1) There are no runnable tasks, or
+        // (2) All runnable tasks have been detached AND there are no blocked attached tasks
+        // If there are some blocked attached tasks and all runnable tasks are detached,
+        // we must run some detached task so that blocked attached tasks may become unblocked.
+        if runnable.is_empty() || (!blocked_attached && runnable.iter().all(|id| self.get(*id).detached)) {
```
jamesbornholt, line 503: Now that I think about it some more, does this logic go away once we implement detached threads? I'm thinking that once that happens, the logic for "finished execution" is just "main thread is finished", because every other thread has either been joined on by the main thread (perhaps transitively), or has been detached.

jamesbornholt (commented):
Looks good! Two more tiny things.

tests/asynch/basic.rs
jamesbornholt: No, I think just checking both executions is good enough, the comment was just a little misleading.

src/runtime/execution.rs
JacobVanGeffen: I see, the (2) case would just boil down to the main thread being finished I think. Maybe we keep this here for now and change the logic in a separate "detached threads" PR?

jamesbornholt: Yeah, we can't make it work that way without doing the detached threads thing, so it makes sense to defer it to later.

jamesbornholt (approved):

PR #51, opened 2021-08-27
Fix how waiters are set and woken

JacobVanGeffen:
Currently, Shuttle does not allow a waiter to be set for a task if a waiter already exists. This may cause a problem when the waiting task wakes itself.
Additionally, it's possible for a waiter to be set for a task, then have the waiter finish before the other task resumes. This happens whenever the waiter wakes itself up. This PR fixes that issue by only waking async waiting tasks when they are blocked.

---

### DIFF
--- a/src/asynch.rs
+++ b/src/asynch.rs
@@ -120,10 +120,12 @@ where
 
                 *self.result.lock().unwrap() = Some(Ok(result));
 
-                // Unblock our waiter if we have one
+                // Unblock our waiter if we have one and it's still alive
                 ExecutionState::with(|state| {
                     if let Some(waiter) = state.current_mut().take_waiter() {
-                        state.get_mut(waiter).unblock();
+                        if !state.get_mut(waiter).finished() {
+                            state.get_mut(waiter).unblock();
+                        }
                     }
                 });
 
--- a/src/runtime/task/mod.rs
+++ b/src/runtime/task/mod.rs
@@ -197,7 +197,10 @@ impl Task {
     /// waiter should block or not. If false, this task has already finished, and so the waiter need
     /// not block.
     pub(crate) fn set_waiter(&mut self, waiter: TaskId) -> bool {
-        assert!(self.waiter.is_none());
+        assert!(
+            self.waiter.is_none() || self.waiter == Some(waiter),
+            "Task cannot have more than one waiter"
+        );
         if self.finished() {
             false
         } else {
--- a/tests/asynch/basic.rs
+++ b/tests/asynch/basic.rs
@@ -1,8 +1,10 @@
-use futures::try_join;
+use futures::{try_join, Future};
 use shuttle::sync::Mutex;
 use shuttle::{asynch, check_dfs, check_random, scheduler::PctScheduler, thread, Runner};
+use std::pin::Pin;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use std::task::{Context, Poll};
 use test_env_log::test;
 
 async fn add(a: u32, b: u32) -> u32 {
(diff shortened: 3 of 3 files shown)


src/asynch.rs
```diff
@@ -123,7 +123,10 @@ where
                 // Unblock our waiter if we have one
                 ExecutionState::with(|state| {
                     if let Some(waiter) = state.current_mut().take_waiter() {
-                        state.get_mut(waiter).unblock();
+                        // NOTE: This could happen if the waiter returns Poll::Ready before the current task
```
jamesbornholt, line 126: rather than a new comment here, just extend the one two lines up:
```
// Unblock our waiter if we have one and it's still alive
```

tests/asynch/basic.rs
```diff
@@ -311,3 +311,58 @@ fn drop_shuttle_yield_future() {
...
+
+// This test checks two behaviors. First, it checks that a future can poll another future
+// and wake itself in the same poll() call. Second, it checks that a waiter can finish
+// before the waitee task.
+#[test]
+fn wake_self_on_join_handle() {
+    use futures::{Future, FutureExt};
+    use shuttle::asynch::JoinHandle;
+    use std::pin::Pin;
+    use std::task::{Context, Poll};
```
jamesbornholt, line 323: move these to the module level to match the rest of this file

tests/asynch/basic.rs
```diff
@@ -311,3 +311,58 @@ fn drop_shuttle_yield_future() {
     assert_eq!(2, async_accesses_clone.load(Ordering::SeqCst));
     assert_eq!(1, post_yield_accesses_clone.load(Ordering::SeqCst));
 }
+
+// This test checks two behaviors. First, it checks that a future can poll another future
+// and wake itself in the same poll() call. Second, it checks that a waiter can finish
+// before the waitee task.
```
jamesbornholt, line 317: If I understand right, the first behavior is actually that a task can be polled twice without needing to wake up any task waiting on it in between.

tests/asynch/basic.rs
```diff
@@ -311,3 +311,58 @@ fn drop_shuttle_yield_future() {
...
+    check_dfs(
+        || {
+            let yielder = asynch::spawn(async move {
+                asynch::yield_now().await;
+            });
+
+            struct Timeout {
+                inner: JoinHandle<()>,
+                counter: u8,
+            }
```
jamesbornholt, line 334: I'd write this as
```
struct Timeout<F: Future<Output = ()>> {
    inner: F,
    counter: u8,
}
```
rather than explicitly naming the JoinHandle type

src/runtime/task/mod.rs
```diff
@@ -197,7 +197,6 @@ impl Task {
     /// waiter should block or not. If false, this task has already finished, and so the waiter need
     /// not block.
     pub(crate) fn set_waiter(&mut self, waiter: TaskId) -> bool {
-        assert!(self.waiter.is_none());
```
jamesbornholt, line 200: Rather than removing this assertion entirely, can we weaken it to:
```
assert!(self.waiter.is_none() || self.waiter == Some(waiter));
```
The point being that Shuttle can't currently handle multiple tasks waiting on the same other task. This way we'll still detect that situation happening rather than just silently missing behaviors.

jamesbornholt (commented):
Nice find!

src/runtime/task/mod.rs
JacobVanGeffen: Ah I didn't realize that. Maybe we can add a message to this assert like "Task cannot have more than one waiter" to make this easier to debug?

tests/asynch/basic.rs
JacobVanGeffen: Sounds good, changing to
```
struct Timeout<F: Future> {
    inner: Pin<Box<F>>,
    counter: u8,
}
```

tests/asynch/basic.rs
JacobVanGeffen: Yes, exactly. I'll change the comment to something like this because maybe there are other possible causes besides this specific scenario.

src/runtime/task/mod.rs
jamesbornholt: yep, that message looks good to me

jamesbornholt (approved):

PR #52, opened 2021-09-04
Make `asynch::block_on` run on the current thread

jamesbornholt:
`block_on` is supposed to run a future to completion *on the current
thread*. This is visible in the type signature for `block_on` in other
futures executors, which do not require the future to be `Send`. We were
instead implementing `block_on` using `spawn`, and so required the
future to be `Send`.

This change makes `block_on` truly run the future to completion on the
current thread, by just running poll in a loop (with the appropriate
yield points inserted).

---

### DIFF
--- a/src/asynch.rs
+++ b/src/asynch.rs
@@ -137,29 +137,25 @@ where
 }
 
 /// Run a future to completion on the current thread.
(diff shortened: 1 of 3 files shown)


jorajeev (approved):
Nice!

PR #54, opened 2021-09-09
Fix waker behavior when invoked before a `poll` finishes

jamesbornholt:
It's possible for a task's waker to be invoked in the middle of a call
to that task's `poll` by the executor. We had accounted for that
possibility if the task called *its own* waker, but that's not good
enough: the waker can escape to other threads that can invoke it before
`poll` finishes (e.g., if the task blocks to acquire a lock).

This change fixes the waker behavior by clarifying the semantics of a
call to `wake`: a task whose waker is invoked should not be blocked
*when it next returns Pending to the executor*, and should be woken if
that has already happened. To do this, we introduce a new `Sleeping`
state for tasks, that has the same semantics as `Blocked` but that is
recognized by waker invocations, which will only unblock a task in
`Sleeping` state. This also removes the special case "woken by self"
behavior -- being woken by *any* thread should be enough to trigger this
sleep logic.

---

### DIFF
--- a/tests/asynch/waker.rs
+++ b/tests/asynch/waker.rs
@@ -1,8 +1,11 @@
+use futures::future::poll_fn;
+use shuttle::sync::atomic::{AtomicBool, Ordering};
+use shuttle::sync::Mutex;
+use shuttle::{asynch, check_dfs, thread};
 use std::future::Future;
 use std::pin::Pin;
+use std::sync::Arc;
 use std::task::{Context, Poll, Waker};
-
-use shuttle::{asynch, check_dfs};
 use test_env_log::test;
 
 #[test]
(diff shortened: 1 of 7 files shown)


tests/asynch/waker.rs
```diff
@@ -49,3 +52,78 @@ fn wake_after_finish() {
...
+            });
+
+            // If this `wake()` invocation happens while the thread above holds the `counter` lock
+            // and the `block_on` task below is blocked waiting to acquire that same lock, then
+            // `wake` must not unblock the `block_on` task. That is, `wake` should prevent the task
+            // from being blocked *the next time it returns Pending*, not just any time it is
+            // blocked.
+            thread::spawn(move || {
+                if let Some(waker) = waker_clone.lock().unwrap().take() {
+                    waker.wake();
```
jorajeev, line 114: Add a check (global var outside Shuttle) to ensure that we explore a run in which this line is executed?

jorajeev (commented):
This PR reminds me -- an interesting use of Shuttle could be to warn if your `poll` function could ever lead to blocking as this may be undesirable.  Maybe we should think about adding an option to generate such a warning?

jamesbornholt:
> This PR reminds me -- an interesting use of Shuttle could be to warn if your `poll` function could ever lead to blocking as this may be undesirable.

I like this idea a whole bunch. Feels related to some of the recent clippy lints around holding a lock inside a future, e.g. https://github.com/rust-lang/rust-clippy/pull/5439.

jorajeev (approved):

PR #56, opened 2021-09-20
Move clock functions into their own module

jamesbornholt:
Just an effort to not pollute the top-level namespace too much. Putting
these into their own module gives us a place to put future functions
that also reveal information about the current execution.

---

### DIFF
--- a/src/current.rs
+++ b/src/current.rs
@@ -1,0 +1,34 @@
+//! Information about the current thread and current Shuttle execution.
+//!
+//! This module provides access to information about the current Shuttle execution. It is useful for
+//! building tools that need to exploit Shuttle's total ordering of concurrent operations; for
+//! example, a tool that wants to check linearizability might want access to a global timestamp for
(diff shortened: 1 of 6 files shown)


src/current.rs
```diff
@@ -0,0 +1,34 @@
...
+/// Get the current thread's vector clock
+pub fn clock() -> VectorClock {
+    crate::runtime::execution::ExecutionState::with(|state| {
+        let me = state.current();
+        state.get_clock(me.id()).clone()
+    })
+}
+
+/// Gets the clock for the thread with the given task ID
+pub fn clock_for(task_id: TaskId) -> VectorClock {
```
jorajeev, line 32: Is this function being used anywhere?

jorajeev (approved):

src/current.rs
jamesbornholt: I don't think so, but I don't think it's doing any harm being there.

bkragl (commented):
LGTM

PR #59, opened 2021-11-16
Implement `Condvar::wait_while` and `Condvar::wait_timeout_while`

jamesbornholt:
The timeout variant has the same caveat as `Condvar::wait_timeout`: we
don't currently test the timeout case, so these methods will always
succeed.

---

### DIFF
--- a/tests/basic/condvar.rs
+++ b/tests/basic/condvar.rs
@@ -32,6 +32,31 @@ fn notify_one() {
     )
 }
 
+#[test]
+fn notify_one_while() {
+    check_dfs(
+        || {
+            let lock = Arc::new(Mutex::new(false));
+            let cond = Arc::new(Condvar::new());
+
+            {
+                let lock = Arc::clone(&lock);
+                let cond = Arc::clone(&cond);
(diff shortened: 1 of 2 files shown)


tests/basic/condvar.rs
```diff
@@ -32,6 +32,30 @@ fn notify_one() {
...
+            let lock = Arc::new(Mutex::new(false));
+            let cond = Arc::new(Condvar::new());
+
+            {
+                let lock = Arc::clone(&lock);
+                let cond = Arc::clone(&cond);
+                thread::spawn(move || {
+                    let guard = lock.lock().unwrap();
+                    cond.wait_while(guard, |flag| !*flag).unwrap();
+                });
```
jorajeev, line 48: Maybe add an assertion here to check that the boolean is now true. (Also in the test below.)

jorajeev (approved):

tests/basic/condvar.rs
jamesbornholt: Yes, good idea, done.

jorajeev (approved):

PR #62, opened 2022-01-23
Remove implicit `Sized` bounds on `Mutex` and `RwLock`

jamesbornholt:
This is to match the standard library, which doesn't require the
contents of a lock to be `Sized`.

---


jamesbornholt:
The test failures are just the new warning in #61.

jorajeev (approved):

PR #66, opened 2022-04-30
Friendlier error messages for re-entrancy failures

jamesbornholt:
Currently, if code tries to re-acquire a lock it already holds, it
correctly panics but with an unhelpful error message about Shuttle's
internals. Instead, let's provide a nicer error message about why the
calling code is buggy.

This came up while trying out the super neat example from a
[fasterthanlime blog post], which Shuttle has no trouble finding the
deadlock in, but wasn't giving a helpful error message for.

[fasterthanlime blog post]: https://fasterthanli.me/articles/a-rust-match-made-in-hell

---

### DIFF
--- a/src/sync/rwlock.rs
+++ b/src/sync/rwlock.rs
@@ -147,11 +147,15 @@ impl<T: ?Sized> RwLock<T> {
         // Block if the lock is in a state where we can't acquire it immediately
         match &state.holder {
             RwLockHolder::Write(writer) => {
-                assert_ne!(*writer, me);
+                if *writer == me {
+                    panic!("deadlock! task {:?} tried to acquire a RwLock it already holds", me);
+                }
                 ExecutionState::with(|s| s.current_mut().block());
             }
             RwLockHolder::Read(readers) => {
-                assert!(!readers.contains(me));
+                if readers.contains(me) {
+                    panic!("deadlock! task {:?} tried to acquire a RwLock it already holds", me);
+                }
                 if typ == RwLockType::Write {
                     ExecutionState::with(|s| s.current_mut().block());
                 }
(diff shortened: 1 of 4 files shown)


src/sync/rwlock.rs
```diff
@@ -147,11 +147,15 @@ impl<T: ?Sized> RwLock<T> {
         // Block if the lock is in a state where we can't acquire it immediately
         match &state.holder {
             RwLockHolder::Write(writer) => {
-                assert_ne!(*writer, me);
+                if *writer == me {
+                    panic!("deadlock! task {:?} tried to acquire a RwLock it already holds", me);
+                }
```
bkragl, line 152: nit: Should be *an* `RwLock`, like in the [doc](https://doc.rust-lang.org/std/sync/struct.RwLock.html). Same below.

bkragl (approved):
Neat example and usability improvement!

One thing I was thinking about is whether it makes sense to have a feature in Shuttle that does precise on-the-fly deadlock detection. Right now there might be a deadlock among some tasks early on in an execution, but we need to wait until the end of the execution to detect it. It might be helpful for debugging to stop the execution right when the deadlock happens.

The nice thing about the current check is that it is completely generic. What I'm proposing would require knowledge about the synchronization primitives (e.g., a combined resource allocation graph for both `Mutex` and `RwLock`).

jamesbornholt:
> One thing I was thinking about is whether it makes sense to have a feature in Shuttle that does precise on-the-fly deadlock detection. Right now there might be a deadlock among some tasks early on in an execution, but we need to wait until the end of the execution to detect it. It might be helpful for debugging to stop the execution right when the deadlock happens.
> 
> The nice thing about the current check is that it is completely generic. What I'm proposing would require knowledge about the synchronization primitives (e.g., a combined resource allocation graph for both `Mutex` and `RwLock`).

I bet we could get some of the same effect by just having the ability to print a stacktrace of all the deadlocked threads. I'll open a separate issue for that.

PR #68, opened 2022-05-17
Determinism Check Scheduler

jdeans289:
Added a DeterminismCheckScheduler to check functions for determinism by recording runnable tasks at each step and comparing on future iterations of the scheduler. Wraps an inner scheduler, which can be Random, RoundRobin, or PCT. Added unit tests as well.

---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -334,6 +334,24 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a scheduler that checks for determinism.
+/// iterations: total number of times to run the inner scheduler
+/// inner_iterations: number of times to test a recorded schedule
+pub fn check_determinism<F>(f: F, iterations: usize, inner_iterations: usize)
+where
+    F: Fn() + Send + Sync + 'static,
+{
+    use crate::scheduler::DeterminismCheckScheduler;
+    use crate::scheduler::RandomScheduler;
+
+    let inner = RandomScheduler::new(iterations);
+
+    let scheduler = DeterminismCheckScheduler::new(inner_iterations, inner);
+
+    let runner = Runner::new(scheduler, Default::default());
+    runner.run(f);
+}
+
 /// Run the given function according to a given encoded schedule, usually produced as the output of
 /// a failing Shuttle test case.
 ///
--- a/src/scheduler/determinism_check.rs
+++ b/src/scheduler/determinism_check.rs
@@ -4,25 +4,6 @@ use std::iter::FromIterator;
 use crate::runtime::task::TaskId;
 use crate::scheduler::{Schedule, Scheduler,ScheduleStep, ScheduleRecord};
 
-// #[derive(Clone, Debug, PartialEq, Eq)]
-// pub struct ScheduleRecordStep {
-//     step: ScheduleStep,
-//     options: HashSet<TaskId>
-// }
-
-// impl ScheduleRecordStep {
-//     /// Create a record step from the chosen task and options
-//     fn new(step: ScheduleStep, options_vec: &[TaskId]) -> Self {
-//         let mut new_record = Self { step, options: HashSet::new() };
-
-// 		for task_id in options_vec {
-// 			new_record.options.insert(task_id.clone());
-// 		}
-		
-// 		new_record
-//     }
-// }
-
 /// A `DeterminismCheckScheduler` wraps an inner `Scheduler`, and when given a program,
 /// decides whether the scheduling decisions can be deterministic or not.
 /// Inner scheduler can be RoundRobin, Random, or PCT
@@ -37,6 +18,7 @@ pub struct DeterminismCheckScheduler<S: ?Sized + Scheduler> {
 }
 
 impl<S: Scheduler> DeterminismCheckScheduler<S> {
+
     /// Create a new `DeterminismCheckScheduler` by wrapping the given `Scheduler` implementation.
     pub fn new(inner_iterations: usize,  inner: S) -> Self {
         Self {
@@ -48,12 +30,6 @@ impl<S: Scheduler> DeterminismCheckScheduler<S> {
 			current_step: 0
         }
     }
-
-	/// Record an instance of non-determinism
-	/// Should probably do something more informative here
-	pub fn log_error(&mut self, message: &str) {
-		assert!(false, "Found nondeterministic function! {}", message);
-	}
 }
 
 
@@ -65,6 +41,8 @@ impl<S: Scheduler> Scheduler for DeterminismCheckScheduler<S> {
 			self.recording = true;
 			self.original_schedule.clear();
 		} else {
+			// Create a new execution to test against prior recording
+			
 			if self.recording {
 				self.recording = false;
 			}
@@ -89,6 +67,7 @@ impl<S: Scheduler> Scheduler for DeterminismCheckScheduler<S> {
     ) -> Option<TaskId> {
 
 		if self.recording {
+
 			// Recording a schedule
 			let choice = self.inner.next_task(runnable_tasks, current_task, is_yielding).unwrap();
 			self.original_schedule.push(ScheduleRecord::new(ScheduleStep::Task(choice), runnable_tasks));
@@ -101,12 +80,11 @@ impl<S: Scheduler> Scheduler for DeterminismCheckScheduler<S> {
 			// Determine whether the state is the same
 
 			if self.current_step >= self.original_schedule.len() {
-				// Current run is longer than original recording
 				panic!("Current execution longer than expected. \n\tOriginal length: {}, acutal length: {}",
 															self.original_schedule.len(), self.current_step);
 			}
 
-			let expected = self.original_schedule.get(self.current_step).unwrap().clone();
+			let expected = self.original_schedule[self.current_step].clone();
 			let expected_step = expected.step;
 			let expected_options = expected.runnable_tasks;
 			let actual_options: HashSet<TaskId> = HashSet::from_iter(runnable_tasks.iter().cloned());
@@ -114,13 +92,13 @@ impl<S: Scheduler> Scheduler for DeterminismCheckScheduler<S> {
 			match expected_step {
 				ScheduleStep::Task(id) => { 
 					if !expected_options.contains(&id) {
-						panic!("\nTask chosen in recording is not currently runnable.\n\tExpected Task: {:?} \n\tActual options: {:?}\n",
-																		id, actual_options);
+						panic!("\nTask chosen in recording is not currently runnable.
+						\n\tExpected Task: {:?} \n\tActual options: {:?}\n", id, actual_options);
 					}
 
 					if expected_options != actual_options {
-						panic!("\nSet of runnable tasks is different than expected.\n\tExpected options: {:?} \n\tActual options: {:?}\n",
-																		expected_options, actual_options);
+						panic!("\nSet of runnable tasks is different than expected.
+						\n\tExpected options: {:?} \n\tActual options: {:?}\n", expected_options, actual_options);
 					}
 
 					self.current_step += 1;
@@ -146,6 +124,5 @@ impl<S: Scheduler> Scheduler for DeterminismCheckScheduler<S> {
 
 		self.current_step += 1;
 		self.inner.next_u64()
-
     }
 }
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -334,6 +334,24 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a scheduler that checks for determinism.
+/// iterations: total number of times to run the inner scheduler
+/// inner_iterations: number of times to test a recorded schedule
```
jamesbornholt, line 339:
```suggestion
/// Run the given function under a scheduler that checks whether the function is deterministic.
/// Each `iteration` will check a different random schedule and replay that schedule
/// `inner_iterations` times to validate determinism.
```

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
+use std::collections::HashSet;
+use std::iter::FromIterator;
+
+use crate::runtime::task::TaskId;
+use crate::scheduler::{Schedule, Scheduler,ScheduleStep, ScheduleRecord};
+
+/// A `DeterminismCheckScheduler` wraps an inner `Scheduler`, and when given a program,
+/// decides whether the scheduling decisions can be deterministic or not.
+/// Inner scheduler can be RoundRobin, Random, or PCT
```
jamesbornholt, line 9:
```suggestion
/// A `DeterminismCheckScheduler` checks whether a given program is deterministic
/// by wrapping an inner `Scheduler` and, for each schedule generated by that scheduler,
/// replaying the schedule a chosen number of times. On each replay, we check that the 
/// schedule is still valid and that the set of runnable tasks is the same at each step.
/// Violations of checks checks mean that the program is not deterministic.
```

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+    inner: Box<S>,
+	iterations: usize,
+	inner_iterations: usize,
+	recording: bool,
+	original_schedule: Vec<ScheduleRecord>,
+	current_step: usize
+}
+
+impl<S: Scheduler> DeterminismCheckScheduler<S> {
+
```
jamesbornholt, line 21: nit: shouldn't need this new line (`cargo fmt` should detect this)

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+impl<S: Scheduler> Scheduler for DeterminismCheckScheduler<S> {
+    fn new_execution(&mut self) -> Option<Schedule> {
+
+		if self.iterations % self.inner_iterations == 0 {
+			// Start a new recording
+			self.recording = true;
+			self.original_schedule.clear();
+		} else {
+			// Create a new execution to test against prior recording
+
```
jamesbornholt, line 45: nit: no blank line

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+		} else {
+			// Create a new execution to test against prior recording
+			
+			if self.recording {
+				self.recording = false;
+			}
+
+			if self.current_step != self.original_schedule.len() {
+				panic!("Current execution ended earlier than original execution\n\tOriginal length: {}, acutal length: {}",
+																self.original_schedule.len(), self.current_step);
```
jamesbornholt, line 52: I think our style is lowercase, and to not include newlines in panics, so:
```suggestion
				panic!("possible nondeterminism: current execution ended earlier than expected (expected length {} but ended after {})", self.original_schedule.len(), self.current_step);
```

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+			self.current_step += 1;
+
+			Some(choice)
+				
+		} else {
+			// Determine whether the state is the same
+
+			if self.current_step >= self.original_schedule.len() {
+				panic!("Current execution longer than expected. \n\tOriginal length: {}, acutal length: {}",
+															self.original_schedule.len(), self.current_step);
```
jamesbornholt, line 84: same as above w.r.t. style, so maybe something like::
```suggestion
				panic!("possible nondeterminism: current execution should have ended after {} steps", self.original_schedule.len(), self.current_step);
```

src/scheduler/mod.rs
```diff
@@ -27,12 +29,33 @@ pub struct Schedule {
...
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 enum ScheduleStep {
     Task(TaskId),
     Random,
 }
 
+/// A `ScheduleRecord` can be used to record both the step
+/// taken and all runnable tasks
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ScheduleRecord {
```
jamesbornholt, line 41: I think we should move this into `determinism_check.rs`, since it's the only scheduler that's going to use it.

src/scheduler/mod.rs
```diff
@@ -27,12 +29,33 @@ pub struct Schedule {
...
+    step: ScheduleStep,
+    runnable_tasks: HashSet<TaskId>
```
jamesbornholt, line 43: Instead of a `HashSet`, could we just make this a `SmallVec<[TaskId; DEFAULT_INLINE_TASKS]>`? I think the invariant is that the `runnable` list we pass to schedulers is in a fixed order, so we can just compare them element by element to check if they're equal. (We need to document that invariant, though; I'll take care of that).

src/scheduler/mod.rs
```diff
@@ -27,12 +29,33 @@ pub struct Schedule {
...
+    /// Create a record step from the chosen task and options
+    fn new(step: ScheduleStep, options_vec: &[TaskId]) -> Self {
+        let mut new_record = Self { step, runnable_tasks: HashSet::new() };
+
+		for task_id in options_vec {
+			new_record.runnable_tasks.insert(task_id.clone());
+		}
+		
+		new_record
+    }
```
jamesbornholt, line 56: Better variable name, and a neat Rust trick:
```rust
    fn new(step: ScheduleStep, runnable_tasks: &[TaskId]) -> Self {
        let runnable_tasks = runnable_tasks.iter().cloned().collect::<HashSet<_>>();
        Self {
            step,
            runnable_tasks,
        }
    }
```

src/scheduler/round_robin.rs
```diff
@@ -16,14 +17,25 @@ impl RoundRobinScheduler {
...
         Self {
             iterations: 0,
+            max_iterations: 1,
+            data_source: RandomDataSource::initialize(0),
+        }
+    }
+
+    /// Construct a new `RoundRobinScheduler` that will execute the test up to max_iteration times, scheduling its
+    /// tasks in a round-robin fashion.
+    pub fn new_multi_run(max_iterations : usize) -> Self {
```
jamesbornholt, line 27: I think we can just drop these changes—we've been meaning to remove the RoundRobinScheduler for a while, and I don't think you're using it anywhere?

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,94 @@
...
+        let threads: Vec<_> = (0..NUM_THREADS)
+        .map(|_| {
+
+            let my_lock = lock.clone();
+
+            thread::spawn(move || {
+                let x = thread_rng().gen::<u64>();
+
+				// Fail every x threads
+                if x % 1000 == 0 {
```
jamesbornholt, line 26: Is this test going to be reliable enough to fail all the time? Wonder if we should just make this `x % 10` or something so it's always going to fail.

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,94 @@
...
+use shuttle::rand::{thread_rng, Rng};
+// use shuttle::scheduler::RandomScheduler;
+use shuttle::sync::{Mutex, Arc};
+use shuttle::{check_determinism, thread};
+use test_log::test;
+
+
+#[test]
+#[should_panic]
+fn shuttle_determinism_test1() {
```
jamesbornholt, line 11: Can we give these tests slightly more suggestive names? This one fails because it randomly acquires a lock, the next one fails because it spawns different numbers of threads.

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,94 @@
...
+#[test]
+fn shuttle_determinism_test3() {
+    shuttle::check_determinism(|| {
+        // Should pass
+        let num_threads: u64 = 100;
+        let lock = Arc::new(Mutex::new(0u64));
+        let threads: Vec<_> = (0..num_threads)
+        .map(|_| {
+            // To share the same cache across the threads, clone it.
+            // This is a cheap operation.
```
jamesbornholt, line 77: nit: think this comment is left over from your other demo

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+			}
+
+			let expected = self.original_schedule[self.current_step].clone();
+			let expected_step = expected.step;
+			let expected_options = expected.runnable_tasks;
+			let actual_options: HashSet<TaskId> = HashSet::from_iter(runnable_tasks.iter().cloned());
```
jamesbornholt, line 90: Think you can drop these 3 lines and just add them below where you do the actual checks.

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+				
+		} else {
+			// Determine whether the state is the same
+
+			if self.current_step >= self.original_schedule.len() {
+				panic!("Current execution longer than expected. \n\tOriginal length: {}, acutal length: {}",
+															self.original_schedule.len(), self.current_step);
+			}
+
+			let expected = self.original_schedule[self.current_step].clone();
```
jamesbornholt, line 87: Shouldn't need to clone here.

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+			let actual_options: HashSet<TaskId> = HashSet::from_iter(runnable_tasks.iter().cloned());
+
+			match expected_step {
+				ScheduleStep::Task(id) => { 
+					if !expected_options.contains(&id) {
+						panic!("\nTask chosen in recording is not currently runnable.
+						\n\tExpected Task: {:?} \n\tActual options: {:?}\n", id, actual_options);
+					}
+
+					if expected_options != actual_options {
```
jamesbornholt, line 99: If you do the thing I mentioned where `ScheduleRecord` holds a `SmallVec`, this can be:
```rust
if !expected.runnable_tasks.iter().eq(runnable_tasks.iter()) {
```

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+			let expected = self.original_schedule[self.current_step].clone();
+			let expected_step = expected.step;
+			let expected_options = expected.runnable_tasks;
+			let actual_options: HashSet<TaskId> = HashSet::from_iter(runnable_tasks.iter().cloned());
+
+			match expected_step {
+				ScheduleStep::Task(id) => { 
+					if !expected_options.contains(&id) {
+						panic!("\nTask chosen in recording is not currently runnable.
+						\n\tExpected Task: {:?} \n\tActual options: {:?}\n", id, actual_options);
```
jamesbornholt, line 96:
```suggestion
						panic!("possible nondeterminism: expected next task is not runnable (expected to run {:?} but runnable tasks were {:?}", id, runnable_tasks);
```

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,128 @@
...
+					}
+
+					if expected_options != actual_options {
+						panic!("\nSet of runnable tasks is different than expected.
+						\n\tExpected options: {:?} \n\tActual options: {:?}\n", expected_options, actual_options);
```
jamesbornholt, line 101:
```suggestion
						panic!("possible nondeterminism: set of runnable tasks is different than expected (expected {:?} but got {:?}), &expected.runnable_tasks[..], runnable_tasks);
```

jamesbornholt (commented):
Thanks for doing this, it's awesome! I have a bunch of minor style/writing comments, but overall it looks great.

The checks are complaining about (1) a clippy warning that should be easy to fix, and (2) some formatting stuff—you can run `cargo fmt` to apply all the formatting changes automatically.

jorajeev:
Work was taken over by @sarsko and merged in PR #96

PR #71, opened 2022-05-25
Implement `Mutex::try_lock`

bkragl:
Without `try_lock` it was easier to justify context switches, because
acquire was a right mover (we only needed a context switch before) and
release was a left mover (we only needed a context switch after).
However, with `try_lock` that is not the case anymore. This commit
argues why we need a context switch at the end of `lock` and `try_lock`
(both in the success and failure case), and why we do not need a context
switch at the beginning of `try_lock` and `MutexGuard::drop`.

---

### DIFF
--- a/src/sync/mutex.rs
+++ b/src/sync/mutex.rs
@@ -61,13 +61,18 @@ impl<T: ?Sized> Mutex<T> {
                 panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
             }
             ExecutionState::with(|s| s.current_mut().block());
+            drop(state);
+
+            // Note that we only need a context switch when we are blocked, but not if the lock is
+            // available. Consider that there is another thread `t` that also wants to acquire the
+            // lock. At the last context switch (where we were chosen), `t` must have been already
+            // runnable and could have been chosen by the scheduler instead. Also, if we want to
+            // re-acquire the lock immediately after having it released, we know that the release
+            // had a context switch that allowed other threads to acquire in between.
+            thread::switch();
+            state = self.state.borrow_mut();
         }
-        drop(state);
 
-        // Acquiring a lock is a yield point
-        thread::switch();
-
-        let mut state = self.state.borrow_mut();
         // Once the scheduler has resumed this thread, we are clear to become its holder.
         assert!(state.waiters.remove(me));
         assert!(state.holder.is_none());
@@ -81,10 +86,13 @@ impl<T: ?Sized> Mutex<T> {
         }
         // Update acquiring thread's clock with the clock stored in the Mutex
         ExecutionState::with(|s| s.update_clock(&state.clock));
+        // Update the vector clock stored in the Mutex with this threads clock.
+        // Future threads that fail a `try_lock` have a causal dependency on this thread's acquire.
+        ExecutionState::with(|s| state.clock.update(s.get_clock(me)));
         drop(state);
 
         // Grab a `MutexGuard` from the inner lock, which we must be able to acquire here
-        match self.inner.try_lock() {
+        let result = match self.inner.try_lock() {
             Ok(guard) => Ok(MutexGuard {
                 inner: Some(guard),
                 mutex: self,
@@ -94,7 +102,15 @@ impl<T: ?Sized> Mutex<T> {
                 mutex: self,
             })),
             Err(TryLockError::WouldBlock) => panic!("mutex state out of sync"),
-        }
+        };
+
+        // We need to let other threads in here so they may fail a `try_lock`. This is the case
+        // because the current thread holding the lock might not have any further context switches
+        // until after releasing the lock. The `concurrent_lock_try_lock` test illustrates this
+        // scenario and would fail if this context switch is not here.
+        thread::switch();
+
+        result
     }
 
     /// Attempts to acquire this lock.
@@ -102,7 +118,68 @@ impl<T: ?Sized> Mutex<T> {
     /// If the lock could not be acquired at this time, then Err is returned. This function does not
     /// block.
     pub fn try_lock(&self) -> TryLockResult<MutexGuard<T>> {
-        unimplemented!()
+        let me = ExecutionState::me();
+
+        let mut state = self.state.borrow_mut();
+        trace!(holder=?state.holder, waiters=?state.waiters, "trying to acquire mutex {:p}", self);
+
+        // We don't need a context switch here. There are two cases to analyze.
+        // * Consider that `state.holder == None` so that we manage to acquire the lock, but that
+        //   there is another thread `t` that also wants to acquire. At the last context switch
+        //   (where we were chosen), `t` must have been already runnable and could have been chosen
+        //   by the scheduler instead. Then `t`'s acquire has a context switch that allows us to
+        //   run into the `WouldBlock` case.
+        // * Consider that `state.holder == Some(t)` so that we run into the `WouldBlock` case,
+        //   but that `t` wants to release. At the last context switch (where we were chosen), `t`
+        //   must have been already runnable and could have been chosen by the scheduler instead.
+        //   Then `t`'s release has a context switch that allows us to acquire the lock.
+
+        let result = if let Some(holder) = state.holder {
+            trace!("`try_lock` failed to acquire mutex {:p} held by {:?}", self, holder);
+            Err(TryLockError::WouldBlock)
+        } else {
+            state.holder = Some(me);
+
+            trace!("acquired mutex {:p}", self);
+
+            // Re-block all other waiting threads, since we won the race to take this lock
+            for tid in state.waiters.iter() {
+                ExecutionState::with(|s| s.get_mut(tid).block());
+            }
+
+            // Grab a `MutexGuard` from the inner lock, which we must be able to acquire here
+            match self.inner.try_lock() {
+                Ok(guard) => Ok(MutexGuard {
+                    inner: Some(guard),
+                    mutex: self,
+                }),
+                Err(TryLockError::Poisoned(guard)) => Err(TryLockError::Poisoned(PoisonError::new(MutexGuard {
+                    inner: Some(guard.into_inner()),
+                    mutex: self,
+                }))),
+                Err(TryLockError::WouldBlock) => panic!("mutex state out of sync"),
+            }
+        };
+
+        // Update the vector clock stored in the Mutex with this threads clock.
+        // Future threads that manage to acquire have a causal dependency on this thread's failed `try_lock`.
+        // Future threads that fail a `try_lock` have a causal dependency on this thread's successful `try_lock`.
+        ExecutionState::with(|s| state.clock.update(s.get_clock(me)));
+
+        // Update this thread's clock with the clock stored in the Mutex.
+        // We need to do the vector clock update even in the failing case, because there's a causal
+        // dependency: if the `try_lock` fails, the current thread `t1` knows that the thread `t2`
+        // that owns the lock is in its critical section, and therefore `t1` has a causal dependency
+        // on everything that happened before in `t2` (which is recorded in the Mutex's clock).
+        ExecutionState::with(|s| s.update_clock(&state.clock));
+        drop(state);
+
+        // We need to let other threads in here so they
+        // (a) may fail a `try_lock` (in case we acquired), or
+        // (b) may release the lock (in case we failed to acquire) so we can succeed in a subsequent `try_lock`.
+        thread::switch();
+
+        result
     }
 
     /// Consumes this mutex, returning the underlying data.
@@ -153,6 +230,12 @@ impl<'a, T: ?Sized> MutexGuard<'a, T> {
 
 impl<'a, T: ?Sized> Drop for MutexGuard<'a, T> {
     fn drop(&mut self) {
+        // We don't need a context switch here *before* releasing the lock. There are two cases to analyze.
+        // * Other threads that want to `lock` are still blocked at this point.
+        // * Other threads that want to `try_lock` and would fail at this point (but not after we release)
+        //   were already runnable at the last context switch (which could have been right after we acquired)
+        //   and could have been scheduled then to fail the `try_lock`.
+
         self.inner = None;
 
         let mut state = self.mutex.state.borrow_mut();
--- a/src/sync/mutex.rs
+++ b/src/sync/mutex.rs
@@ -53,42 +53,48 @@ impl<T: ?Sized> Mutex<T> {
         let mut state = self.state.borrow_mut();
         trace!(holder=?state.holder, waiters=?state.waiters, "waiting to acquire mutex {:p}", self);
 
-        // We are waiting for the lock
-        state.waiters.insert(me);
         // If the lock is already held, then we are blocked
         if let Some(holder) = state.holder {
             if holder == me {
                 panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
             }
-            ExecutionState::with(|s| s.current_mut().block());
+
+            state.waiters.insert(me);
             drop(state);
 
             // Note that we only need a context switch when we are blocked, but not if the lock is
             // available. Consider that there is another thread `t` that also wants to acquire the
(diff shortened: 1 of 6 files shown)


tests/basic/mutex.rs
```diff
@@ -114,6 +116,125 @@ fn concurrent_increment() {
     });
 }
 
+#[test]
+/// Two concurrent threads trying to do an atomic increment using `try_lock`.
+/// One `try_lock` must succeed, while the other may or may not succeed.
+/// Thus we expect to see final values 1 and 2.
+fn concurrent_try_increment() {
+    let final_value_tracker = Arc::new(
```
jorajeev, line 124: Feels a bit heavyweight (and harder to read) to use a HashMap here.  Maybe use a HashSet similar to what's done in the tests for atomics?

tests/basic/mutex.rs
```diff
@@ -114,6 +116,125 @@ fn concurrent_increment() {
...
+
+#[test]
+/// Two concurrent threads, one doing an atomic increment by 1 using `lock`,
+/// the other trying to do an atomic increment by 1 followed by trying to do
+/// an atomic increment by 2 using `try_lock`. The `lock` must succeed, while
+/// both `try_lock`s may or may not succeed. Thus we expect to see final value
+/// 1 (both `try_lock`s fail), 2 (second `try_lock` fails), 3 (first `try_lock`
+/// fails), and 4 (both `try_lock`s succeed).
+fn concurrent_lock_try_lock() {
+    let final_value_tracker = Arc::new(
```
jorajeev, line 179: See comment above about using HashSets

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
                 mutex: self,
             })),
             Err(TryLockError::WouldBlock) => panic!("mutex state out of sync"),
-        }
+        };
+
+        // We need to let other threads in here so they may fail a `try_lock`
```
jorajeev, line 99: You could explain this a bit more, maybe even pointing to the `concurrent_lock_try_lock` test and saying that test doesn't fail the `try_lock`s if this context switch is not here.

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
...
+        trace!(holder=?state.holder, waiters=?state.waiters, "trying to acquire mutex {:p}", self);
+
+        if let Some(holder) = state.holder {
+            if holder == me {
+                panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
+            }
+        }
+
+        // We don't need a context switch here. There are two cases to analyze.
+        // * Consider that we manage to acquire the lock, but that there is another thread that
```
jorajeev, line 122: Super nit: I think these comments would be easier to read if you started with `Consider that state_holder == None` and `Consider that state_holder == Some(th) and th wants to release`.  Maps more directly to the code structure, and gives a name to `the other thread`.

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
...
+        //   thread must have been already runnable and could have been chosen by the scheduler.
+        //   Then that other thread's acquire has a context switch that allows us to run into the
+        //   `WouldBlock` case.
+        // * Consider that we run into the `WouldBlock` case, but that there is another thread
+        //   that wants to release. At the last context switch (where we were chosen), the other
+        //   thread must have been already runnable and could have been chosen by the scheduler.
+        //   Then that other thread's release has a context switch that allows us to acquire the
+        //   lock.
+
+        let result = if state.holder.is_none() {
```
jorajeev, line 133: Nit: could you rewrite this using a `match` and in the `trace!` on line 158, name the thread that's currently holding the lock?

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
...
+            state.holder = Some(me);
+
+            trace!("acquired mutex {:p}", self);
+
+            // Re-block all other waiting threads, since we won the race to take this lock
+            for tid in state.waiters.iter() {
+                ExecutionState::with(|s| s.get_mut(tid).block());
+            }
+            // Update acquiring thread's clock with the clock stored in the Mutex
```
jorajeev, line 142: You need to do the vector clock update even in the failing case, because there's a causal dependency: if the `try_lock` fails, the current thread T1 *knows* that the thread T2 that owns the lock is in its critical section, and therefore T1 has a causal dependency on everything that happened before in T2 (which is recorded in the lock's clock, at line 83).

src/sync/mutex.rs
jamesbornholt: +1 to basically writing this comment as a case analysis over `state.holder`

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
...
+                panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
+            }
+        }
+
+        // We don't need a context switch here. There are two cases to analyze.
+        // * Consider that we manage to acquire the lock, but that there is another thread that
+        //   also wants to acquire. At the last context switch (where we were chosen), the other
+        //   thread must have been already runnable and could have been chosen by the scheduler.
+        //   Then that other thread's acquire has a context switch that allows us to run into the
+        //   `WouldBlock` case.
```
jamesbornholt, line 126: Does this same argument also work for the existing `thread::switch` on line 68 of the regular `lock` path? Basically, if you ever enter `lock`, there must have been an earlier context switch that chose to run you, and could instead have chosen to run someone else to win the race for the lock.

(I think the answer is no, because I just tried it, but I can't convince myself why)

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
...
     pub fn try_lock(&self) -> TryLockResult<MutexGuard<T>> {
-        unimplemented!()
+        let me = ExecutionState::me();
+
+        let mut state = self.state.borrow_mut();
+        trace!(holder=?state.holder, waiters=?state.waiters, "trying to acquire mutex {:p}", self);
+
+        if let Some(holder) = state.holder {
+            if holder == me {
+                panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
```
jamesbornholt, line 117: Is this what the real `try_lock` does? The docs suggest that it might instead just return `WouldBlock` here, but I haven't tested that.

src/sync/mutex.rs
jamesbornholt: Can we refactor this code path into a common `post_acquire` function? It looks very similar to the `lock` code; refactoring might make any differences more explicit.

src/sync/mutex.rs
```diff
@@ -94,15 +94,79 @@ impl<T: ?Sized> Mutex<T> {
...
+                    mutex: self,
+                }),
+                Err(TryLockError::Poisoned(guard)) => Err(TryLockError::Poisoned(PoisonError::new(MutexGuard {
+                    inner: Some(guard.into_inner()),
+                    mutex: self,
+                }))),
+                Err(TryLockError::WouldBlock) => panic!("mutex state out of sync"),
+            }
+        } else {
+            trace!("failed to acquired mutex {:p}", self);
```
jamesbornholt, line 158: maybe just to be more clear that this is a "good" failure, not a bug:
```suggestion
            trace!("`try_lock` failed to acquire mutex {:p} held by {:?}", self, state.holder.unwrap());
```

tests/basic/mutex.rs
jamesbornholt: Yeah, I think the HashSet approach might be a little clearer in both these tests.

tests/basic/mutex.rs
```diff
@@ -114,6 +116,125 @@ fn concurrent_increment() {
...
+            let threads = (0..2)
+                .map(|_| {
+                    let lock = Arc::clone(&lock);
+                    thread::spawn(move || match lock.try_lock() {
+                        Ok(mut guard) => {
+                            *guard += 1;
+                        }
+                        Err(TryLockError::WouldBlock) => (),
+                        Err(_) => panic!("unexpected TryLockError"),
+                    })
```
jamesbornholt, line 144: I mildly dislike what rustfmt does here; adding a semicolon at the end of the closure body makes it more palatable to me:
```suggestion
                    thread::spawn(move || {
                        match lock.try_lock() {
                            Ok(mut guard) => {
                                *guard += 1;
                            }
                            Err(TryLockError::WouldBlock) => (),
                            Err(_) => panic!("unexpected TryLockError"),
                        };
                    })
```

tests/basic/mutex.rs
```diff
@@ -114,6 +116,125 @@ fn concurrent_increment() {
...
+                        Err(TryLockError::WouldBlock) => (),
+                        Err(_) => panic!("unexpected TryLockError"),
+                    };
+                    match lock.try_lock() {
+                        Ok(mut guard) => {
+                            *guard += 2;
+                        }
+                        Err(TryLockError::WouldBlock) => (),
+                        Err(_) => panic!("unexpected TryLockError"),
+                    }
```
jamesbornholt, line 212:
```suggestion
                    for n in 1..3 {
                        match lock.try_lock() {
                            Ok(mut guard) => {
                                *guard += n;
                            }
                            Err(TryLockError::WouldBlock) => (),
                            Err(_) => panic!("unexpected TryLockError"),
                        };
                    }
```

tests/basic/mutex.rs
```diff
@@ -114,6 +116,125 @@ fn concurrent_increment() {
...
+    }
+}
+
+#[test]
+/// Two concurrent threads, one doing an atomic increment by 1 using `lock`,
+/// the other trying to do an atomic increment by 1 followed by trying to do
+/// an atomic increment by 2 using `try_lock`. The `lock` must succeed, while
+/// both `try_lock`s may or may not succeed. Thus we expect to see final value
+/// 1 (both `try_lock`s fail), 2 (second `try_lock` fails), 3 (first `try_lock`
+/// fails), and 4 (both `try_lock`s succeed).
```
jamesbornholt, line 177: nit: maybe list these outcomes as bullet points

jamesbornholt (commented):
I *think* all the broken tests are because we have an extra yield point now, so existing schedules don't replay (in which case we should bump the minor version on the next release) and some of the tests that count context switches are now wrong.

This extra yield point also makes the `clock_condvar_notify_all_dfs` test extremely slow; wonder if we can do something about that.

src/sync/mutex.rs
bkragl: Good catch. It indeed returns `WouldBlock` in this case.

src/sync/mutex.rs
bkragl: For regular `lock` we need to implement blocking. If the previous context switch decides who wins the race, then the other threads stay runnable and can be scheduled while the lock is still held. I think what that means is that we can do the context switch *only* in the case where we find the lock held by another thread.

src/sync/mutex.rs
jorajeev: Actually, there's even more to it.  In the failing try_lock case, you also have to update the vector clock stored in the Mutex, because future threads that manage to acquire the lock have a causal dependency on this failed try_lock.

src/sync/mutex.rs
bkragl: With the change in vector clock updates, I think a refactoring does not really help.

bkragl:
> I _think_ all the broken tests are because we have an extra yield point now, so existing schedules don't replay (in which case we should bump the minor version on the next release) and some of the tests that count context switches are now wrong.

I fixed all tests.

> This extra yield point also makes the `clock_condvar_notify_all_dfs` test extremely slow; wonder if we can do something about that.

I just did an experiment on my local machine. On the current main branch, `clock_condvar_notify_all_dfs` takes 60 seconds. On this PR (including the revision of doing a context switch *before* `lock` only in case we need to block) it takes "only" 40 seconds.

src/sync/mutex.rs
```diff
@@ -61,13 +61,16 @@ impl<T: ?Sized> Mutex<T> {
                 panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
             }
             ExecutionState::with(|s| s.current_mut().block());
+            drop(state);
+
+            // Note that we only need a context switch when we are blocked, but not if the lock is
+            // available. Consider that there is another thread `t` that also wants to acquire the
+            // lock. At the last context switch (where we were chosen), `t` must have been already
+            // runnable and could have been chosen by the scheduler instead.
```
jorajeev, line 69: I think this is assuming that there is a context switch when a lock is dropped.  Because without a yield in the drop, a thread that releases a lock and then immediately tries to acquire it again will always succeed, which misses a potential interleaving.

bkragl: There *is* a context switch in drop (after doing the release):
https://github.com/bkragl/shuttle/blob/e47be262a3afe5590423f5e2b1b0927e698f4063/src/sync/mutex.rs#L266-L267

bkragl: Oh, I guess your point is that this should be part of the argument.

jorajeev: Yes, was asking you to update the comment.

jamesbornholt: Would be good to add a mutex unit test for the scenario Rajeev's describing (if we don't already have one somewhere).

bkragl: Updated the comment and added test `unlock_yields`.

tests/basic/dfs.rs
```diff
@@ -48,6 +48,84 @@ fn trivial_two_threads() {
...
+/// Additionally, T0 and T1 may block before acquiring the lock, which we denote by B and b,
+/// respectively.
+///
+/// We have a total of 16 interleavings: 8 where the critical sections do not overlap (4 where
+/// T0 goes first and 4 symmetric ones where T1 goes first), and 8 where the critical sections
+/// overlap (4 where T0 goes first and T1 blocks, and 4 symmetric ones where T1 goes first and
+/// T0 blocks). The following computation tree illustrates all interleavings.
+///
+/// ```
+/// digraph G {
```
jorajeev, line 64: Nice!

tests/basic/mutex.rs
```diff
@@ -114,6 +115,230 @@ fn concurrent_increment() {
...
+            }
+
+            let value = *lock.try_lock().unwrap();
+            observed_values_clone.lock().unwrap().insert(value);
+        },
+        None,
+    );
+
+    let observed_values = Arc::try_unwrap(observed_values).unwrap().into_inner().unwrap();
+    assert_eq!(observed_values.len(), 2);
```
jorajeev, line 263: Think you can just do `assert_eq!(observed_values, HashSet::from([1, 2]))` here, and similarly in other tests below.

src/sync/mutex.rs
```diff
@@ -94,15 +99,84 @@ impl<T: ?Sized> Mutex<T> {
...
+        //   but that `t` wants to release. At the last context switch (where we were chosen), `t`
+        //   must have been already runnable and could have been chosen by the scheduler instead.
+        //   Then `t`'s release has a context switch that allows us to acquire the lock.
+
+        let result = if let Some(holder) = state.holder {
+            trace!("`try_lock` failed to acquire mutex {:p} held by {:?}", self, holder);
+
+            // Update the vector clock stored in the Mutex, because future threads that manage to
+            // acquire the lock have a causal dependency on this failed `try_lock`.
+            ExecutionState::with(|s| state.clock.update(s.get_clock(me)));
```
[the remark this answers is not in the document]
bkragl: In `try_lock` that update was already done in the successful case, so I factored it out from the if branch to be done in both cases.

tests/basic/mutex.rs
bkragl: Right, that'll give a more comprehensive error message.

jamesbornholt (approved):

PR #72, opened 2022-06-30
Implement `RwLock::{try_read, try_write}`

jamesbornholt:
This is a little tricky because `std` is unclear about whether a thread
can acquire the same read lock multiple times. For `read` it says:

> This function might panic when called if the lock is already held by
> the current thread.

So acquiring a second read lock _might_ fail. But for `try_read` it says:

> This function will return the WouldBlock error if the RwLock could not
> be acquired because it was already locked exclusively.

suggesting that `try_read` _must_ succeed the second time (the lock is
not held exclusively).

We resolve this ambiguity by choosing a conservative semantics that
always forbids a thread acquiring the read lock twice. This helps us
catch deadlocks, especially in async programs where a task might
nondeterministically migrate between threads and only deadlock if that
migration didn't happen.

Another difficulty with this change is that causality is pretty hairy
for the read side of a `RwLock`. In principle, concurrent readers
shouldn't inherit each other's causality, as they don't affect whether
the lock was readable or not. But that's really hard to implement,
especially with `try_write` in the picture too. So again we choose a
conservative implementation for our vector clocks that just always
inherits causality from all prior lock holders. This is likely too
strong, so we'd explore unnecessary symmetries, but I can't convince
myself of the correctness of a weaker implementation.

---

### DIFF
--- a/tests/basic/clocks.rs
+++ b/tests/basic/clocks.rs
@@ -1,7 +1,7 @@
 use shuttle::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use shuttle::sync::mpsc::{channel, sync_channel};
 use shuttle::sync::{Barrier, Condvar, Mutex, Once, RwLock};
-use shuttle::{check_dfs, check_pct, current, thread};
+use shuttle::{check_dfs, check_pct, check_random, current, thread};
 use std::collections::HashSet;
 use std::sync::Arc;
 use test_log::test;
@@ -10,7 +10,7 @@ pub fn me() -> usize {
     usize::from(thread::current().id())
 }
 
-// TODO Maybe make this a macro so backtraces are more informative
+#[track_caller]
 pub fn check_clock(f: impl Fn(usize, u32) -> bool) {
     for (i, &c) in current::clock().iter().enumerate() {
         assert!(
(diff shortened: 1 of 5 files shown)


tests/basic/rwlock.rs
```diff
@@ -220,3 +221,286 @@ fn rwlock_into_inner() {
...
+        5000,
+    );
+}
+
+/// Three concurrent threads, one doing an atomic increment by 1 using `write`, one trying to do an
+/// atomic increment by 1 followed by trying to do an atomic increment by 2 using `try_write`, and a
+/// third that peeks at the value using `try_read.` The `write` must succeed, while each `try_write`
+/// may or may not succeed. Thus we expect to see final values:
+/// * 0 (`try_read` ran first)
+/// * 1 (both `try_writes`s fail),
```
jorajeev, line 349: But the reader may see a final value of 1 even if one (or both) try_writes were successful (if the reader went before the try_writes).  So this comment is misleading.

One thing you could do is to read the value again after the joins, and then look at all the pairs of values (value_read_by_try_read, value_read_after_joins).

tests/basic/rwlock.rs
```diff
@@ -220,3 +221,286 @@ fn rwlock_into_inner() {
...
+        None,
+    )
+}
+
+/// This behavior isn't consistent with `std`, which seems to suggest that `try_read` succeeds if
+/// the current thread already holds a read lock. We assume it always fails so that we can more
+/// easily diagnose potential deadlocks, especially with async tasks that might migrate across
+/// threads in real implementations.
+#[test]
+fn double_try_read() {
```
jorajeev, line 461: There's a variant of this where a thread does a `read()` and then a `try_read()`

tests/basic/clocks.rs
```diff
@@ -76,32 +76,34 @@ fn clock_mutex_pct() {
...
             thread::spawn(move || {
                 check_clock(|i, c| (c > 0) == (i == 0));
-                let mut set = set.write().unwrap();
+                let _guard = lock.write().unwrap();
+                let mut set = set.lock().unwrap();
                 set.insert(me());
-                // Check that the only nonzero clock entries are for the threads in the set
-                check_clock(|i, c| (c > 0) == set.contains(&i));
+                assert!(!set.contains(&1)); // dummy thread is never in the set
+                check_clock(|i, c| !set.contains(&i) || (c > 0));
```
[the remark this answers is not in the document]
jamesbornholt: It's to handle race condition described at the top of the test—readers can have a non-zero clock without being in the set.

tests/basic/rwlock.rs
jamesbornholt: Good point. I updated the test to check all the possible pairs of values.

tests/basic/rwlock.rs
jamesbornholt: I figured this would be the same as if the first `try_read` succeeds (assuming that a successful `try_read` and a `read` are equivalent), but may as well add the explicit test anyway.

tests/basic/clocks.rs
```diff
@@ -336,7 +345,7 @@ fn clock_mpsc_bounded() {
                 // The sender has sent a message, so its clock is nonzero
                 let c1 = current::clock().get(1);
                 assert!(c1 > 0);
-                let _ = rx.recv().unwrap();
+                rx.recv().unwrap();
```
jamesbornholt, line 348: This is a Clippy 1.62 thing.

jorajeev (approved):

Issue #75, opened 2022-08-16
Implement thread park(), unpark(), and panicking()


kvark:
Already a TODO in code: https://github.com/awslabs/shuttle/blob/4a00174db03b2ef9dc4a1e0707b94827bac15f7a/src/thread.rs#L173
This would help making `shuttle::thread` to be an in-place substitute to `std::thread`.

jamesbornholt:
#77 will add park/unpark, and #78 will add `hint::spin_loop`, which I think you might also need. Those two PRs plus this diff to choir was enough to get your `queue::shuttle` test running (and passing!):

```diff
diff --git a/src/queue.rs b/src/queue.rs
index e9b677d..cfcbb42 100644
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -1,8 +1,8 @@
 #[cfg(shuttle)]
-use shuttle::{sync, thread};
-use std::{cell::UnsafeCell, hint, mem};
+use shuttle::{sync, hint, thread};
+use std::{cell::UnsafeCell, mem};
 #[cfg(not(shuttle))]
-use std::{sync, thread};
+use std::{sync, hint, thread};
 
 use self::sync::atomic::{AtomicUsize, Ordering};
 
@@ -194,7 +194,7 @@ impl<T> Drop for Queue<T> {
 #[cfg(shuttle)]
 #[test]
 fn shuttle() {
-    shuttle::check_random(test_barrage, 10000);
+    shuttle::check_pct(test_barrage, 3, 10000);
 }
 
 #[test]
@@ -219,7 +219,7 @@ fn test_smoke() {
 #[cfg_attr(not(shuttle), test)]
 fn test_barrage() {
     const NUM_THREADS: usize = if cfg!(miri) { 2 } else { 8 };
-    const NUM_ELEMENTS: usize = if cfg!(miri) { 1 << 7 } else { 1 << 16 };
+    const NUM_ELEMENTS: usize = if cfg!(miri) || cfg!(shuttle) { 1 << 7 } else { 1 << 16 };
     let sq = sync::Arc::new(Queue::new(NUM_ELEMENTS));
     let mut handles = Vec::new();
 
```

[...]

PR #77, opened 2022-08-18
Implement `thread::park` and `Thread::unpark`

jamesbornholt:
These were missing, as pointed out in #75. They're not too hard to implement---we just add an explicit `park_state` to each task.

There's a subtlety here in that `park` won't work well with futures due to our one-future-per-thread model, but the behavior of `park` within a future is probably unreliable on any realistic async runtime anyway. This is one of a class of behaviors around async that maybe we should have some way of warning about.

---

### DIFF
--- a/tests/basic/thread.rs
+++ b/tests/basic/thread.rs
@@ -1,5 +1,6 @@
-use shuttle::sync::Mutex;
+use shuttle::sync::{Condvar, Mutex};
 use shuttle::{check_dfs, check_random, thread};
+use std::collections::HashSet;
 use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use std::sync::Arc;
 use test_log::test;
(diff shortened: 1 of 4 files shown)


tests/basic/thread.rs
```diff
@@ -376,3 +377,119 @@ mod thread_local {
...
+        || {
+            let lock = Arc::new(Mutex::new(false));
+            let condvar = Arc::new(Condvar::new());
+
+            let reader = {
+                let lock = Arc::clone(&lock);
+                let condvar = Arc::clone(&condvar);
+                thread::spawn(move || {
+                    let mut guard = lock.lock().unwrap();
+                    while !*guard {
```
jorajeev, line 440: Think a better version of this test would be to replace the `while` with an `if` and then add an `assert!(*guard)` afterwards.

jorajeev (approved):

PR #82, opened 2022-09-26
Inline a simplified varint implementation

jamesbornholt:
The maintainer of the `varmint` crate we're using is moving on and would
like to deprecate the crate [[1]], so let's inline a simplified version
for now. I think we should revisit our serialization format soon anyway,
but this will unblock the `varmint` maintainer.

[1]: https://github.com/mycorrhiza/varmint-rs/issues/8

---

### DIFF
--- a/src/scheduler/serialization.rs
+++ b/src/scheduler/serialization.rs
@@ -4,7 +4,76 @@
 use crate::runtime::task::TaskId;
 use crate::scheduler::{Schedule, ScheduleStep};
 use bitvec::prelude::*;
-use varmint::*;
+
+/// A simplified version of the deprecated [`varmint`](https://github.com/mycorrhiza/varmint-rs)
+/// crate, used under the MIT license.
+mod varint {
+    pub fn space_needed(val: u64) -> usize {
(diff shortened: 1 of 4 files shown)


src/scheduler/serialization.rs
```diff
@@ -4,7 +4,76 @@
 use crate::runtime::task::TaskId;
 use crate::scheduler::{Schedule, ScheduleStep};
 use bitvec::prelude::*;
-use varmint::*;
+
+/// A simplified version of the deprecated [`varmint`](https://github.com/mycorrhiza/varmint-rs)
+/// crate, used under the MIT license.
+mod varint {
+    pub fn space_needed(val: u64) -> usize {
+        let used_bits = u64::min_value().leading_zeros() - val.leading_zeros();
```
jorajeev, line 12: Why `u64::min_value().leading_zeros()` instead of `size_of::<u64>() * 8` ?

jorajeev: I realize you copied this from varmint, just wondering why it's done that way

jorajeev (approved):

PR #84, opened 2022-11-11
Make PCT scheduling not linear in max number of tasks

jamesbornholt:
Currently our PCT implementation does a linear search through the list of priorities. This list's length is determined by the maximum number of tasks a test has ever created. For tests that create many short-lived tasks (like many async tests), this means we'll be doing a very expensive linear scan just to choose from a small set of runnable tasks.

Instead, maintain priorities as a map, and look up the runnable tasks in the map. This does a little bit more work in the case where |priorities| ~= |runnable|, because there's no early exit any more, but we expect the runnable set to generally be fairly small and often significantly smaller than the total set.

---


src/scheduler/pct.rs
```diff
@@ -90,13 +104,19 @@ impl Scheduler for PctScheduler {
...
-            let index = self.rng.gen_range(0, self.priority_queue.len() + 1);
-            self.priority_queue.insert(index, TaskId::from(tid));
+        // If any new tasks were created, assign them priorities by randomly swapping them with an
+        // existing task's priority, so we maintain the invariant that every priority is distinct
+        let max_known_task = self.priorities.len();
+        let max_new_task = usize::from(*runnable.iter().max().unwrap());
+        for new_task_id in max_known_task..1 + max_new_task {
+            let target_task_id = TaskId::from(self.rng.gen_range(0, self.priorities.len()));
+            let target_priority = *self.priorities.get(&target_task_id).expect("priority queue invariant");
+            debug_assert!(self
```
jorajeev, line 114: Also `debug_assert`

src/scheduler/pct.rs
```diff
@@ -90,13 +104,19 @@ impl Scheduler for PctScheduler {
...
-
-        for tid in known_tasks..1 + max_tasks {
-            let index = self.rng.gen_range(0, self.priority_queue.len() + 1);
-            self.priority_queue.insert(index, TaskId::from(tid));
+        // If any new tasks were created, assign them priorities by randomly swapping them with an
+        // existing task's priority, so we maintain the invariant that every priority is distinct
+        let max_known_task = self.priorities.len();
+        let max_new_task = usize::from(*runnable.iter().max().unwrap());
+        for new_task_id in max_known_task..1 + max_new_task {
+            let target_task_id = TaskId::from(self.rng.gen_range(0, self.priorities.len()));
```
jorajeev, line 112: This code ensures that a new task always gets a priority lower than some existing task.  (Note that previously we chose a random number in the range 0..self.priority_queue.len()+1 so a new task could go to the end.)  Maybe generate a target_task_id in range 0..self.priorities.len()+1?

src/scheduler/pct.rs
```diff
@@ -69,8 +72,19 @@ impl Scheduler for PctScheduler {
...
+            debug_assert_eq!(
+                self.priorities.iter().collect::<HashSet<_>>().len(),
+                self.priorities.len()
+            );
+
+            // Reinitialize priorities by shuffling the task IDs
+            let mut priorities = (0..self.priorities.len()).collect::<Vec<_>>();
+            priorities.shuffle(&mut self.rng);
+            for (i, priority) in priorities.into_iter().enumerate() {
+                debug_assert!(self.priorities.insert(TaskId::from(i), priority).is_some());
```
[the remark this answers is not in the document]
jamesbornholt: oh, good grief, fixed

src/scheduler/pct.rs
jamesbornholt: Yeah, nice catch, done.

jorajeev (approved):

PR #89, opened 2022-11-22
Upgrade dependencies

akoshelev:
Projects that use latest version of `rand` crate are largely unable
    to use deterministic random generators from shuttle because `RngCore`
    and friends are incompatible between the latest version (0.6.4) and version
    that Shuttle uses (0.5.1).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,17 +13,17 @@ bitvec = "1.0.1"
 generator = "0.7.1"
 hex = "0.4.2"
 owo-colors = "3.5.0"
-rand_core = "0.5.1"
-rand = "0.7.3"
-rand_pcg = "0.2.1"
+rand_core = "0.6.4"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle"
-version = "0.4.1"
+version = "0.5.0"
```
jamesbornholt, line 3: Could you undo this? I'll prep a release in a separate commit since I need to do the changelog too.

jamesbornholt (approved):

Cargo.toml
jamesbornholt: Never mind, forgot that GitHub lets me do this myself.

akoshelev: ah sorry, next time I'll be more careful and will not touch this

PR #92, opened 2023-01-17
Add descriptive error message when PCT rejects tests without concurrency

bkragl:
It is a sanity check to have some concurrency in every execution of the test closure. However, in PCT the assertion is really there to prevent an underflow. I'm not convinced that we should add similar assertions to other schedulers.

---

### DIFF
--- a/src/scheduler/pct.rs
+++ b/src/scheduler/pct.rs
@@ -68,9 +68,10 @@ impl Scheduler for PctScheduler {
         self.steps = 0;
 
         // On the first iteration, we run a simple oldest-task-first scheduler to determine a
-        // bound on the maximum number of steps. Once we have that, we can initialize PCT.
+        // lower bound on the maximum number of steps. Once we have that, we can initialize PCT.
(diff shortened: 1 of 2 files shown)


src/scheduler/pct.rs
```diff
@@ -68,9 +68,13 @@ impl Scheduler for PctScheduler {
...
 
         // On the first iteration, we run a simple oldest-task-first scheduler to determine a
-        // bound on the maximum number of steps. Once we have that, we can initialize PCT.
+        // lower bound on the maximum number of steps. Once we have that, we can initialize PCT.
+        // Note that we dynamically update the bound if we discover an execution with more steps.
         if self.iterations > 0 {
-            assert!(self.max_steps > 0);
+            assert!(
+                self.max_steps > 0,
+                "test closure admits an execution without any concurrency"
```
jamesbornholt, line 76:
```suggestion
                "test closure did not exercise any concurrency"
```

jamesbornholt (approved):

PR #93, opened 2023-01-20
Implement `lazy_static` support

jamesbornholt:
This change adds support for the `lazy_static` crate. It's mostly built around a `Once` cell for each static plus some global storage.

---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -214,9 +214,12 @@ pub struct Config {
     /// not abort a currently running test iteration; the limit is only checked between iterations.
     pub max_time: Option<std::time::Duration>,
 
-    /// Whether to enable warnings about [Shuttle's unsound implementation of
-    /// `atomic`](crate::sync::atomic#warning-about-relaxed-behaviors).
-    pub silence_atomic_ordering_warning: bool,
+    /// Whether to silence warnings about Shuttle behaviors that may miss bugs or introduce false
+    /// positives:
+    /// 1. [Unsound implementation of `atomic`](crate::sync::atomic#warning-about-relaxed-behaviors)
+    ///    may miss bugs
+    /// 2. [`lazy_static` values are dropped](mod@crate::lazy_static) at the end of an execution
+    pub silence_warnings: bool,
 }
 
 impl Config {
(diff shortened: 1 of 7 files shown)


tests/basic/lazy_static.rs
```diff
@@ -0,0 +1,226 @@
...
+            assert_eq!(**THING, 0);
+
+            thd.join().unwrap();
+        },
+        None,
+    );
+}
+
+#[test]
+fn mutex_dfs() {
```
jorajeev, line 81: Could you improve the test name, and add a comment -- this test checks that DFS will explore the N! interleavings during initialization?

tests/basic/lazy_static.rs
```diff
@@ -0,0 +1,226 @@
...
+                .collect::<Vec<_>>();
+
+            for thd in thds {
+                thd.join().unwrap();
+            }
+
+            assert_eq!(THREADS.lock().unwrap().len(), NUM_THREADS);
+            initializers.lock().unwrap().insert(THREADS.lock().unwrap().clone());
+        },
+        10000,
```
jorajeev, line 148: nit: `10_000`

tests/basic/lazy_static.rs
```diff
@@ -0,0 +1,226 @@
...
+            thd.join().unwrap();
+        },
+        None,
+    );
+}
+
+// Ensure that concurrent Shuttle tests see an isolated version of a lazy_static. This test is best
+// effort, as it spawns OS threads and hopes they race.
+#[test]
+fn shared_static() {
```
jorajeev, line 181: That's a nice test

jorajeev (commented):
Looks good.  One issue is that `lazy_static` [does not run Drop methods on the type](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=00104ebdb977d2b24b30c0e656b70d4e).  However, this version will run the destructors.

We can either (1) alert users to this difference, or (2) do `std::mem::forget` on the value when we finish an execution?

src/lib.rs
```diff
@@ -414,3 +415,64 @@ macro_rules! __thread_local_inner {
             };
     }
 }
+
+/// Declare a new [lazy static value](crate::lazy_static::Lazy), like the `lazy_static` crate.
```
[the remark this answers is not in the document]
jamesbornholt: I actually used this the other day, so added :-)

tests/basic/lazy_static.rs
jamesbornholt: Done, and ended up refactoring a little to share the two tests.

jorajeev (approved):

PR #96, opened 2023-02-22
Add Determinism Check Scheduler

sarsko:
This PR revives @jdeans289s [PR #68](https://github.com/awslabs/shuttle/pull/68).

It takes the commits by @jdeans289, rebases them on the current master, and then addresses the comments left by @jamesbornholt. The only exception is that `ScheduleRecord` still contains a `runnable_tasks: HashSet<TaskId>` rather than a `SmallVec<[TaskId; DEFAULT_INLINE_TASKS]`, as I didn't bother spending time on verifying that the runnable list we pass to schedulers is in a fixed order. 

---


src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl ScheduleRecord {
+    /// Create a record step from the chosen task and options
+    fn new(step: ScheduleStep, runnable_tasks: &[TaskId]) -> Self {
+        let runnable_tasks = runnable_tasks.iter().cloned().collect::<HashSet<_>>();
+        Self { step, runnable_tasks }
+    }
+}
+
+/// /// A `DeterminismCheckScheduler` checks whether a given program is deterministic
```
jamesbornholt, line 23:
```suggestion
/// A `DeterminismCheckScheduler` checks whether a given program is deterministic
```

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    fn next_task(
+        &mut self,
+        runnable_tasks: &[TaskId],
+        current_task: Option<TaskId>,
+        is_yielding: bool,
+    ) -> Option<TaskId> {
+        if self.recording {
+            // Recording a schedule
+            let choice = self.inner.next_task(runnable_tasks, current_task, is_yielding).unwrap();
```
jamesbornholt, line 84: Not sure unwrap is the right thing here? The inner scheduler is allowed to return `None` to stop early, and we should respect that here. (This might also be worth a unit test -- use an inner scheduler that stops after 5 steps or something).

src/scheduler/determinism_check.rs
```diff
@@ -0,0 +1,141 @@
...
+                    if expected.runnable_tasks != actual_options {
+                        panic!("possible nondeterminism: set of runnable tasks is different than expected (expected {:?} but got {:?})", expected.runnable_tasks, actual_options);
+                    }
+
+                    self.current_step += 1;
+
+                    Some(id)
+                }
+                ScheduleStep::Random => {
+                    panic!("Found context switch, but recording expected random number generation")
```
jamesbornholt, line 120:
```suggestion
                    panic!("possible nondeterminism: next step was context switch, but recording expected random number generation")
```

src/scheduler/round_robin.rs
```diff
@@ -16,14 +17,25 @@ impl RoundRobinScheduler {
...
         Self {
             iterations: 0,
+            max_iterations: 1,
+            data_source: RandomDataSource::initialize(0),
+        }
+    }
+
+    /// Construct a new `RoundRobinScheduler` that will execute the test up to max_iteration times, scheduling its
+    /// tasks in a round-robin fashion.
+    pub fn new_multi_run(max_iterations: usize) -> Self {
```
jamesbornholt, line 27: Maybe just collapse this into the existing constructor (which will be API breaking but that's fine)

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[test]
+#[should_panic]
+fn shuttle_determinism_test_spawn_random_amount_of_threads() {
+    shuttle::check_determinism(
+        || {
+            // Should fail
+            let num_threads: u64 = thread_rng().gen::<u64>() % 100;
+            println!("Number of threads: {}", num_threads);
```
jamesbornholt, line 49: delete

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,98 @@
+// use crate::check_replay_roundtrip;
+use shuttle::rand::{thread_rng, Rng};
+// use shuttle::scheduler::RandomScheduler;
+use shuttle::sync::{Arc, Mutex};
+use shuttle::{check_determinism, thread};
+use test_log::test;
+
+#[test]
+#[should_panic]
+fn shuttle_determinism_test_randomly_acquire_lock() {
```
jamesbornholt, line 10: The tests in this file don't need to start with `shuttle_determinism_test` (that's implied by the module we're in).

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,98 @@
...
+                            *num += 1;
+                        }
+                    })
+                })
+                .collect();
+
+            threads.into_iter().for_each(|t| t.join().expect("Failed"));
+
+            let num = lock.lock().unwrap();
+            println!("Lock accessed {} times", num);
```
jamesbornholt, line 35: delete

tests/basic/determinism_check.rs
```diff
@@ -0,0 +1,98 @@
...
+
+            threads.into_iter().for_each(|t| t.join().expect("Failed"));
+
+            let num = lock.lock().unwrap();
+            assert!(*num == num_threads);
+        },
+        100,
+        10,
+    );
+}
```
jamesbornholt, line 98: A couple of other tests that might be good:
- versions that use a DFS scheduler as the inner scheduler
- the early stopping thing I mentioned above

jamesbornholt (commented):
I'm not sure this is quite doing what we want. Our goal is to check that all the nondeterminism of the program being tested is under Shuttle's control. We're not trying to test if the _underlying scheduler_ is non-deterministic. So when replaying a schedule, we don't want to consult the underlying scheduler at all—we should just replay our recorded schedule and see whether the state of the program is the same. If it's not, we know there's some nondeterminism in the program.

For example, this program is fine, because although it's nondeterministic, all the nondeterminism is under Shuttle's control:
```rust
let num_threads = shuttle::rand::thread_rng().gen_range(1..10);
let lock = Arc::new(Mutex::new(0u64));
let threads: Vec<_> = (0..num_threads)
    .map(|_| {
        let my_lock = lock.clone();
        thread::spawn(move || {
            *my_lock.lock().unwrap() += 1;
        })
    })
    .collect();
threads.into_iter().for_each(|t| t.join().expect("Failed"));
let num = lock.lock().unwrap();
assert!(*num == num_threads);
```
But if we change `shuttle::rand::thread_rng()` to just the regular `::rand::thread_rng()`, this is no longer fine, because there's nondeterminism that isn't under our control.

sarsko:
@jamesbornholt : Thanks! I should have resolved all of your comments (with the exception of early stopping, as I couldn't find a scheduler which returns a `None`). Ended up renaming from `DeterminismCheckScheduler` to `UncontrolledRandomnessCheckScheduler` (or rather: I initially had both, then ended up deleting `DeterminismCheckScheduler` as I don't really see the need for it.).

tests/basic/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,216 @@
...
+
+            let lock = Arc::new(Mutex::new(0u64));
+            let threads: Vec<_> = (0..NUM_THREADS)
+                .map(|_| {
+                    let my_lock = lock.clone();
+
+                    thread::spawn(move || {
+                        let x = thread_rng().gen::<u64>();
+
+                        // Fail every x threads
```
jorajeev, line 37: This comment is confusing -- there's no "failure" here, we're just picking one of every 10 threads to increment the shared counter.

In any case, why not just make every thread lock the counter, and add `x % 10` to it?  Why bother with the `if` ?

tests/basic/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,216 @@
...
+
+            threads.into_iter().for_each(|t| t.join().expect("Failed"));
+        },
+        1000,
+    );
+}
+
+#[test]
+#[should_panic = "possible nondeterminism"]
+fn randomly_acquire_lock_regular_rand() {
```
jorajeev, line 54: Could we refactor (e.g., pass a closure for the call to `thread_rng`) to reuse code between this test and the one above?

tests/basic/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,216 @@
...
+
+            let num = lock.lock().unwrap();
+            assert!(*num == num_threads);
+        },
+        100,
+    );
+}
+
+#[test]
+#[should_panic = "possible nondeterminism"]
```
jorajeev, line 109: I'm not seeing what this test and the one above cover beyond what's already achieved by the `randomly_acquire_lock_*_rand` tests above?

src/scheduler/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,133 @@
...
+
+            match &self.previous_schedule[self.current_step] {
+                ScheduleRecord::Task(maybe_id, runnables, was_yielding) => {
+                    if let Some(id) = maybe_id {
+                        if !runnable_tasks.contains(&id) {
+                            panic!("possible nondeterminism: expected next task is not runnable (expected to run {id:?} but runnable tasks were {runnable_tasks:?}");
+                        }
+                    }
+
+                    if runnables != runnable_tasks {
```
jorajeev, line 90: Isn't this check stronger than the check on line 85 above?

jorajeev (commented):
Thanks for doing this CR.  One high-level comment about the tests you have: there are 6 different ways for the nondeterminism check to fail (with line numbers from `uncontrolled_randomness.rs`):

- schedule ends early (line 45)
- schedule doesn't end as expected (lines 76, 118)
- task is not runnable (line 86)
- set of runnables is different (line 91)
- `is_yielding` is different (line 95)
- next-step type mismatch (lines 103, 125)

It would be great if we could have tests that exercise all of these cases.

src/scheduler/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum ScheduleRecord {
+    Task(Option<TaskId>, Vec<TaskId>, bool),
+    Random(u64),
+}
+
+/// An `UncontrolledRandomnessCheckScheduler` checks whether a given program is deterministic
+/// by wrapping an inner `Scheduler` and, for each schedule generated by that scheduler,
+/// replaying the schedule a chosen number of times. On each replay, we check that the
```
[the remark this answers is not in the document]
sarsko: Yes, leftover comment.

src/scheduler/round_robin.rs
```diff
@@ -7,23 +7,25 @@ use crate::scheduler::{Schedule, Scheduler};
 #[derive(Debug)]
 pub struct RoundRobinScheduler {
     iterations: usize,
+    max_iterations: usize,
```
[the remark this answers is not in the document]
sarsko: @jdeans289 added a new constructor for the `RoundRobinScheduler`, so that it would be possible to use it with the determinism checker (otherwise it would only be possible to test it once). @jamesbornholt then commented that we could probably just combine the two. I dunno, there is no particular reason to use the `RoundRobinScheduler` when checking for uncontrolled randomness, so no big need to change it.

src/scheduler/mod.rs
```diff
@@ -27,7 +29,7 @@ pub struct Schedule {
     steps: Vec<ScheduleStep>,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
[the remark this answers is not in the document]
sarsko: It did before when it was used in a HashSet. No need anymore.

tests/basic/uncontrolled_randomness.rs
sarsko: If we have every thread lock the counter, then we will not catch that there is uncontrolled randomness. We are catching discrepancies in the locking of the mutex, not in the value of `num`.

src/scheduler/round_robin.rs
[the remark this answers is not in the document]
sarsko: Left it in. I can unchange it if you want.

src/lib.rs
```diff
@@ -339,6 +339,22 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a scheduler that checks whether the function
+/// contains randomness which is not controlled by Shuttle.
+/// Each iteration will check a different random schedule and replay that schedule once.
+pub fn check_uncontrolled_randomness<F>(f: F, max_iterations: usize)
```
jamesbornholt, line 345: naming nit: I'd probably call them `UncontrolledNondeterminism`—thread scheduling isn't random.

src/scheduler/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,124 @@
...
+            current_step: 0,
+        }
+    }
+}
+
+impl<S: Scheduler> Scheduler for UncontrolledRandomnessCheckScheduler<S> {
+    fn new_execution(&mut self) -> Option<Schedule> {
+        // Dummy schedule. We do this instead of doing `self.scheduler.new_execution`
+        // as that would cause the schedule to be ran half as many times as intended.
+        let mut out = Some(Schedule::new(12345));
```
jamesbornholt, line 40: Just to be explicit that there's nothing special about this number
```suggestion
        let mut out = Some(Schedule::new(0));
```

src/scheduler/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,124 @@
+use crate::runtime::task::TaskId;
+use crate::scheduler::{Schedule, Scheduler};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum ScheduleRecord {
+    Task(Option<TaskId>, Vec<TaskId>, bool),
```
jamesbornholt, line 6: Probably use `SmallVec<[TaskId; DEFAULT_INLINE_TASKS]>` instead of the Vec, to save an allocation on every step.

tests/basic/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    let config = Config::default();
+
+    let scheduler = UncontrolledRandomnessCheckScheduler::new(scheduler);
+
+    let runner = shuttle::Runner::new(scheduler, config);
+    runner.run(f);
+}
+
+fn randomly_acquire_lock<F: (Fn() -> Box<dyn rand::RngCore>) + Send + Sync>(thread_rng: &'static F) {
```
jamesbornholt, line 24:
```suggestion
fn randomly_acquire_lock<R: rand::RngCore, F: (Fn() -> R) + Send + Sync>(thread_rng: &'static F) {
```
and then remove all the `Box`es at the callsites

tests/basic/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,231 @@
...
+                let mut num = my_lock.lock().unwrap();
+                *num += 1;
+            })
+        })
+        .collect();
+
+    threads.into_iter().for_each(|t| t.join().expect("Failed"));
+}
+
+fn spawn_random_amount_of_threads<F: (Fn() -> Box<dyn rand::RngCore>) + Send + Sync>(
```
jamesbornholt, line 62: same thing, don't need Box

jamesbornholt (commented):
Nice! Just some nitpicks.

tests/basic/uncontrolled_randomness.rs
```diff
@@ -0,0 +1,231 @@
...
+    let num_threads: u64 = thread_rng().gen::<u64>() % max_threads;
+    have_n_threads_acquire_mutex(num_threads);
+}
+
+fn spawn_random_amount_of_threads_mutex_rng(rng: &Mutex<StdRng>, max_threads: u64) {
+    let num_threads = rng.lock().unwrap().gen::<u64>() % max_threads;
+    have_n_threads_acquire_mutex(num_threads);
+}
+
+fn have_n_threads_yield(num_threads: u64) {
```
jorajeev: Could you move these helper functions down to just before the tests that use them?  It's much easier to read code when definitions are only introduced before their first use.

jorajeev (commented):
Some minor nits.  But really great that you could exercise all the failure cases.

jorajeev (approved):

PR #98, opened 2023-03-29
Add tag

sarsko:
### The general idea: 
Have each task store a tag which it inherits from its parent. This tag can then be used for debugging to track parenthood and to turn on/off printing (ie set tag once done with setup, set tag when entering/leaving a critical section, set tags for different kinds of tasks to track them differently, etc).

In addition to general debugging, there are a fair amount of things which can be built upon tags. 

One example is a "tag-aware" scheduler (or extending the current to be "tag-aware") which could enable encoding fairness or precedence properties (ie: if there exists a task with [such-and-such tag] then it will be scheduled [at least every N steps]) or custom deadlock properties (if there are no tasks with [such-and-such tag] which are scheduleable, then that is to be considered a deadlock). 

Another example is to be able to model partial failures, by crashing/aborting/not scheduling tasks with certain tags.

### About the current solution:

Currently adds what I think is roughly the minimal needed to be useful. Decided on `u64` because 1: fixed number of bits is nice for parenthood encoding schemes and 2: @jorajeev convinced me that signed integers are actually Not Good 3: something bigger/arbitrary boxed data etc. seems like overkill.

It also exposes `step_cnt` and `current_task_as_string`, which I have found super useful to have access to when debugging. For a bit more context, this started as a hack where I exposed `ExecutionState` and did something ala:
```rust
pub fn task_info() -> String {
        format!(
            "T: {} S: {} I: {:?} ",
            ExecutionState::get_tag(),
            ExecutionState::step_cnt(),
            ExecutionState::me()
        )
}

#[macro_export]
macro_rules! task_info {
    ($($tts:tt)*) => {{
        use $crate::runtime::execution::ExecutionState;
        print!(ExecutionState::task_info());
        print!($($tts)*);
    }};
}

#[macro_export]
macro_rules! prnt {
    ($test:expr, $($tts:tt)*) => {{
        use $crate::task_info;
        if $test {
            task_info!($($tts)*);
        }
    }};
}

#[macro_export]
macro_rules! prntln {
    ($test:expr, $($tts:tt)*) => {{
        use $crate::prnt;
        if $test {
            prnt!($test,$($tts)*);
            println!();
        }
    }};
}

#[macro_export]
macro_rules! t {
    () => {
        ExecutionState::get_colour()
    };
}
```

And then something ala `prntln!(t!() > 0, "GETTING LOCK")` at interesting points in the code.

As for the included test: fairly simple test which tests that tags are inherited and that tags don't "bleed", as well as a few different ways of setting the tags. Tests tags both for futures and threads.

Finally, one thing which was considered was to store a closure of type `u64 -> u64` which specifies how to generate the tag of the child, and then having it default to the identity function. Decided against it for now, as I have not needed it yet. There might be patterns where this could be useful, but I think the scheme of setting the tag immediately after a spawn is easier to comprehend, less error prone, more versatile, and not so much work that it is an issue.

---

### DIFF
--- a/src/current.rs
+++ b/src/current.rs
@@ -34,8 +34,11 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
 }
 
 /// Sets the `tag` field of the current task.
-pub fn set_tag_for_current_task(tag: Tag) {
+/// Returns the `tag` which was there previously.
+pub fn set_tag_for_current_task(tag: Tag) -> Tag {
+    let old_tag = get_tag_for_current_task();
     ExecutionState::set_tag_for_current_task(tag);
+    old_tag
 }
 
 /// Gets the `tag` field of the current task.
--- a/src/current.rs
+++ b/src/current.rs
@@ -22,7 +22,7 @@ pub fn context_switches() -> usize {
 
 /// Get the current thread's vector clock
 pub fn clock() -> VectorClock {
-    crate::runtime::execution::ExecutionState::with(|state| {
+    ExecutionState::with(|state| {
         let me = state.current();
         state.get_clock(me.id()).clone()
     })
@@ -36,9 +36,7 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
 /// Sets the `tag` field of the current task.
 /// Returns the `tag` which was there previously.
 pub fn set_tag_for_current_task(tag: Tag) -> Tag {
-    let old_tag = get_tag_for_current_task();
-    ExecutionState::set_tag_for_current_task(tag);
-    old_tag
+    ExecutionState::set_tag_for_current_task(tag)
 }
 
 /// Gets the `tag` field of the current task.
@@ -46,8 +44,7 @@ pub fn get_tag_for_current_task() -> Tag {
     ExecutionState::get_tag_for_current_task()
 }
 
-/// Gets the `TaskId` of the current task
-/// NOTE: Will panic if there is no current task.
-pub fn get_current_task() -> TaskId {
-    ExecutionState::me()
+/// Gets the `TaskId` of the current task, or `None` if there is no current task.
+pub fn get_current_task() -> Option<TaskId> {
+    ExecutionState::with(|s| Some(s.try_current()?.id()))
 }
--- a/src/current.rs
+++ b/src/current.rs
@@ -7,7 +7,7 @@
 
 use crate::runtime::execution::ExecutionState;
 use crate::runtime::task::clock::VectorClock;
-use crate::runtime::task::TaskId;
+pub use crate::runtime::task::{Tag, TaskId};
 
 /// The number of context switches that happened so far in the current Shuttle execution.
 ///
@@ -31,4 +31,19 @@ pub fn clock() -> VectorClock {
 /// Gets the clock for the thread with the given task ID
 pub fn clock_for(task_id: TaskId) -> VectorClock {
     ExecutionState::with(|state| state.get_clock(task_id).clone())
+}
+
+/// Sets the `tag` field of the current task.
+pub fn set_tag_for_current_task(tag: Tag) {
+    ExecutionState::set_tag_for_current_task(tag);
+}
+
+/// Gets the `tag` field of the current task.
+pub fn get_tag_for_current_task() -> Tag {
+    ExecutionState::get_tag_for_current_task()
+}
+
+/// Gets the `TaskId` of the current task
+pub fn get_current_task() -> TaskId {
+    ExecutionState::me()
 }
--- a/src/current.rs
+++ b/src/current.rs
@@ -44,6 +44,7 @@ pub fn get_tag_for_current_task() -> Tag {
 }
 
 /// Gets the `TaskId` of the current task
+/// NOTE: Will panic if there is no current task.
 pub fn get_current_task() -> TaskId {
     ExecutionState::me()
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -395,13 +395,13 @@ where
 }
 
 /// Sets the `tag` field of the current task.
-pub fn set_tag(tag: i64) {
+pub fn set_tag(tag: u64) {
     use runtime::execution::ExecutionState;
     ExecutionState::set_tag(tag);
 }
 
 /// Gets the `tag` field of the current task.
-pub fn get_tag() -> i64 {
+pub fn get_tag() -> u64 {
     use runtime::execution::ExecutionState;
     ExecutionState::get_tag()
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -394,30 +394,6 @@ where
     runner.run(f);
 }
 
-/// Sets the `tag` field of the current task.
-pub fn set_tag(tag: u64) {
-    use runtime::execution::ExecutionState;
-    ExecutionState::set_tag(tag);
-}
-
-/// Gets the `tag` field of the current task.
-pub fn get_tag() -> u64 {
-    use runtime::execution::ExecutionState;
-    ExecutionState::get_tag()
-}
-
-/// Gets the current step count.
-pub fn step_cnt() -> usize {
-    use runtime::execution::ExecutionState;
-    ExecutionState::step_cnt()
-}
-
-/// Gets the current step count.
-pub fn current_task_as_string() -> String {
-    use runtime::execution::ExecutionState;
-    ExecutionState::current_task_as_string()
-}
-
 /// Declare a new thread local storage key of type [`LocalKey`](crate::thread::LocalKey).
 #[macro_export]
 macro_rules! thread_local {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -394,6 +394,30 @@ where
     runner.run(f);
 }
 
+/// Sets the `tag` field of the current task.
+pub fn set_tag(tag: i64) {
+    use runtime::execution::ExecutionState;
+    ExecutionState::set_tag(tag);
+}
+
+/// Gets the `tag` field of the current task.
+pub fn get_tag() -> i64 {
+    use runtime::execution::ExecutionState;
+    ExecutionState::get_tag()
+}
+
+/// Gets the current step count.
+pub fn step_cnt() -> usize {
+    use runtime::execution::ExecutionState;
+    ExecutionState::step_cnt()
+}
+
+/// Gets the current step count.
+pub fn current_task_as_string() -> String {
+    use runtime::execution::ExecutionState;
+    ExecutionState::current_task_as_string()
+}
+
 /// Declare a new thread local storage key of type [`LocalKey`](crate::thread::LocalKey).
 #[macro_export]
 macro_rules! thread_local {
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -1,6 +1,7 @@
 use crate::runtime::failure::{init_panic_hook, persist_failure, persist_task_failure};
 use crate::runtime::storage::{StorageKey, StorageMap};
 use crate::runtime::task::clock::VectorClock;
+use crate::runtime::task::Tag;
 use crate::runtime::task::{Task, TaskId, DEFAULT_INLINE_TASKS};
 use crate::runtime::thread::continuation::PooledContinuation;
 use crate::scheduler::{Schedule, Scheduler};
@@ -283,7 +284,7 @@ impl ExecutionState {
                 .map(|t| t.span.clone())
                 .unwrap_or_else(tracing::Span::current);
             let task_id = TaskId(state.tasks.len());
-            let tag = state.get_tag_internal();
+            let tag = state.get_tag_or_default_for_current_task();
             let clock = state.increment_clock_mut(); // Increment the parent's clock
             clock.extend(task_id); // and extend it with an entry for the new task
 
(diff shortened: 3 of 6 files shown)


src/runtime/execution.rs
```diff
@@ -568,6 +570,35 @@ impl ExecutionState {
...
+
+    fn set_tag_internal(&mut self, tag: u64) {
+        self.current_mut().set_tag(tag);
+    }
+
+    pub(crate) fn set_tag(tag: u64) {
+        ExecutionState::with(|s| {
+            s.set_tag_internal(tag);
+        });
+    }
```
bkragl, line 582: If somebody calls this in an improper context (i.e., where there is no current task), they'll get a very unhelpful error message. There are some options. One would be to do `try_current().expect("something helpful")` here. Another would be to make `current`, `get`, and friends in `ExecutionState` expect instead of unwrap, now that they are used by public functions.

src/runtime/execution.rs
```diff
@@ -568,6 +570,35 @@ impl ExecutionState {
...
+    fn get_tag_internal(&self) -> u64 {
+        match self.try_current() {
+            Some(current) => current.get_tag(),
+            None => 0,
+        }
+    }
+
+    pub(crate) fn get_tag() -> u64 {
+        ExecutionState::with(|s| s.get_tag_internal())
+    }
```
bkragl, line 593: I think that `get_tag` shouldn't call `get_tag_internal`. The default tag `0` should only be used to initialize the "root task", but not returned if `get_tag` is called in a wrong context.

src/runtime/task/mod.rs
```diff
@@ -58,11 +58,14 @@ pub(crate) struct Task {
     name: Option<String>,
 
     local_storage: StorageMap,
+
+    // Arbitrarily settable tag which is inherited from the parent.
+    tag: u64,
```
bkragl, line 63: Make a `struct Tag(u64)`?

src/lib.rs
```diff
@@ -394,6 +394,30 @@ where
...
+    use runtime::execution::ExecutionState;
+    ExecutionState::step_cnt()
+}
+
+/// Gets the current step count.
+pub fn current_task_as_string() -> String {
+    use runtime::execution::ExecutionState;
+    ExecutionState::current_task_as_string()
+}
+
```
bkragl, line 420: Move to the `current` module? Is `step_cnt` sufficiently different from `context_switches`?

bkragl (commented):
Nice feature! Some comments below.

> tags don't "bleed"

What does that mean?

src/lib.rs
sarsko: Agreed. I was not aware of `context_switches`. I guess it is more fine-grained and thus probably better.

src/runtime/execution.rs
sarsko: I dunno. Again, wrong context is really not possible in user land. This is as it is so that setup (line 288: `let tag = state.get_tag_internal();`) can be done in an easy/readable way.

sarsko:
> > tags don't "bleed"
> 
> What does that mean?

Oh yeah, should probably have written this differently. With no "bleed" I mean the property that one task setting a tag does not affect other tasks, and that future tasks that happen to get mapped onto the same TaskId do not keep the previous tag.

src/runtime/execution.rs
bkragl: I was not completely sure about that. If it is guaranteed that client code can only run when `current_task` is `Some(_)` then it is fine.

src/runtime/execution.rs
bkragl: To make it really readable the function name could be `get_tag_or_default` or something similar. (and maybe make `0` a constant with a descriptive name)

src/runtime/task/mod.rs
sarsko: I think the intended use is for people to do arithmetic with it. I think part of the point is that we cannot change the representation -- we are giving people 64 bits to encode whatever they want and are committing to that.

bkragl: In my mind, the common use case for tags is to represent something discrete (like, we initially called them "colors") that you only want to compare. I understand that it makes sense to commit to a representation and give more flexibility. With the struct clients can still grab the u64 and do with it whatever they want, they just have to be explicit about it. So I don't see any harm in having some extra protection in the common case.

src/runtime/execution.rs
```diff
@@ -568,6 +570,35 @@ impl ExecutionState {
             self.current_span_entered = Some(unsafe { extend_span_entered_lt(self.current_span.enter()) });
         }
     }
+
+    fn set_tag_internal(&mut self, tag: u64) {
```
jorajeev, line 574: This name is too vague.  Something like `set_tag_for_current_task` would be better.

src/lib.rs
```diff
@@ -394,6 +394,30 @@ where
     runner.run(f);
 }
 
+/// Sets the `tag` field of the current task.
+pub fn set_tag(tag: u64) {
```
jorajeev, line 398: These need better names, maybe  `set_tag_for_current_task`?

src/lib.rs
```diff
@@ -394,6 +394,30 @@ where
...
+    use runtime::execution::ExecutionState;
```
jorajeev, line 399: Could we avoid introducing all these `use` statements that take up screen real estate? Either move these `use` up to the top, or just write `runtime::execution::ExecutionState::set_tag`

src/lib.rs
```diff
@@ -394,6 +394,30 @@ where
...
+    ExecutionState::get_tag()
+}
+
+/// Gets the current step count.
+pub fn step_cnt() -> usize {
+    use runtime::execution::ExecutionState;
+    ExecutionState::step_cnt()
+}
+
+/// Gets the current step count.
```
jorajeev, line 415: Copy-n-paste error in comment.

But in any case, I don't see why we are providing a public function to expose the current TaskId as a String.  If needed, we should just expose the `TaskId` itself.

src/lib.rs
```diff
@@ -394,6 +394,30 @@ where
...
+    ExecutionState::set_tag(tag);
+}
+
+/// Gets the `tag` field of the current task.
+pub fn get_tag() -> u64 {
+    use runtime::execution::ExecutionState;
+    ExecutionState::get_tag()
+}
+
+/// Gets the current step count.
```
jorajeev, line 409: Why are we exposing this?

tests/basic/tag.rs
```diff
@@ -0,0 +1,70 @@
...
+fn spawn_some_futures_and_set_tag<F: (Fn(u64, u64) -> u64) + Send + Sync>(
+    tag_on_entry: u64,
+    f: &'static F,
+    num_threads: u64,
+) {
+    let threads: Vec<_> = (0..num_threads)
+        .map(|i| {
+            shuttle::future::spawn(async move {
+                assert!(get_tag() == tag_on_entry);
+                set_tag(f(tag_on_entry, i));
```
jorajeev, line 14: Rewrite as
```
let new_tag = f(tag_on_entry, i);
set_tag(new_tag);
assert_eq!(get_tag(), new_tag);
```
which is easier to read.

src/runtime/task/mod.rs
jorajeev: I'm inclined to agree with Bernhard.  Defining an opaque `Tag` type that has from/into conversions to `u64` might be better, as it gives us a bit more freedom to change the internal representation later.

src/lib.rs
sarsko: Not exposed anymore -- I was not aware of `context_switches()`

tests/basic/tag.rs
```diff
@@ -0,0 +1,70 @@
...
+            })
+        })
+        .collect();
+
+    threads.into_iter().for_each(|t| t.join().expect("Failed"));
+
+    assert!(get_tag() == tag_on_entry);
+}
+
+fn spawn_threads_which_spawn_more_threads(tag_on_entry: u64, num_threads: u64) {
```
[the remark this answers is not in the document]
sarsko: This is not properly addressed in the most recent commit. Made the test into a somewhat ugly test which does kind of what is described, but not what I think you had in mind. I'll revisit once I'm back.

tests/basic/tag.rs
```diff
@@ -1,70 +1,111 @@
...
-                assert!(get_tag() == i);
+                assert_eq!(get_tag_for_current_task(), tag_on_entry);
+                let new_tag = i.into();
+                set_tag_for_current_task(new_tag);
+                assert_eq!(get_tag_for_current_task(), new_tag);
+                spawn_some_threads_and_set_tag(new_tag, &|_, _| 123.into(), 13);
+                assert_eq!(get_tag_for_current_task(), new_tag);
+                spawn_some_threads_and_set_tag(new_tag, &|_, x| (x * 13).into(), 7);
+                assert_eq!(get_tag_for_current_task(), new_tag);
+                spawn_some_threads_and_set_tag(new_tag, &|p, x| ((u64::from(p) << 4) + x).into(), 19);
```
jorajeev, line 68: Feels like three calls to `spawn_some_threads_and_set_tag` are a bit overkill since they

src/current.rs
```diff
@@ -32,3 +32,18 @@ pub fn clock() -> VectorClock {
...
+pub fn set_tag_for_current_task(tag: Tag) {
+    ExecutionState::set_tag_for_current_task(tag);
+}
+
+/// Gets the `tag` field of the current task.
+pub fn get_tag_for_current_task() -> Tag {
+    ExecutionState::get_tag_for_current_task()
+}
+
+/// Gets the `TaskId` of the current task
```
jorajeev, line 46: Add a note that this will panic if there is no current task.

tests/basic/tag.rs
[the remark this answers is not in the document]
jorajeev: I was thinking of something like the following:
```
    fn spawn_thread_and_set_tag(tag_on_entry: u64, new_tag: u64) -> JoinHandle<u64> {
        thread::spawn(move || {
            assert_eq!(get_tag(), tag_on_entry);
            set_tag(new_tag);
            assert_eq!(get_tag(), new_tag);
            new_tag.into()
        })
    }
```

And then the test body is just
```
    set_tag(42);
    let h1 = spawn_thread_and_set_tag(42, 84);
    set_tag(50);
    let h2 = spawn_thread_and_set_tag(50, 100);
    let results = join_all([h1, h2]);
    assert_eq!(results, [84, 100]);
```

jamesbornholt:
High-level comment: can we just implement this as a thread-local exposed in `shuttle::current`? We'd get to reuse the infrastructure for thread locals that we already have.

src/current.rs
```diff
@@ -32,3 +32,19 @@ pub fn clock() -> VectorClock {
...
+/// NOTE: Will panic if there is no current task.
+pub fn get_current_task() -> TaskId {
```
jorajeev, line 48: Could we change this to return an `Option<TaskId>` and use `ExecutionState::try_current`, so that this can be called from a context where there may be no current task?

src/current.rs
```diff
@@ -32,3 +32,19 @@ pub fn clock() -> VectorClock {
 pub fn clock_for(task_id: TaskId) -> VectorClock {
     ExecutionState::with(|state| state.get_clock(task_id).clone())
 }
+
+/// Sets the `tag` field of the current task.
+pub fn set_tag_for_current_task(tag: Tag) {
```
jorajeev, line 37: Could we change this function to also return the old tag?

tests/basic/tag.rs
```diff
@@ -0,0 +1,135 @@
...
+    );
+
+    threads.into_iter().for_each(|t| t.join().expect("Failed"));
+
+    assert_eq!(get_tag_for_current_task(), new_tag_main_thread);
+}
+
+#[test]
+fn threads_which_spawn_threads_which_spawn_threads() {
+    check_random(|| spawn_threads_which_spawn_more_threads(Tag::default(), 15, 10), 10)
```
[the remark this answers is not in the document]
sarsko: Sure

src/current.rs
```diff
@@ -34,8 +34,11 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
 }
 
 /// Sets the `tag` field of the current task.
-pub fn set_tag_for_current_task(tag: Tag) {
+/// Returns the `tag` which was there previously.
+pub fn set_tag_for_current_task(tag: Tag) -> Tag {
+    let old_tag = get_tag_for_current_task();
     ExecutionState::set_tag_for_current_task(tag);
```
jorajeev, line 40: Would be nicer to just let `ExecutionState::set_tag_for_current_task` return the old tag, instead of introducing a local variable and calling `get_tag...`

tests/basic/tag.rs
```diff
@@ -109,3 +110,26 @@ fn spawn_threads_which_spawn_more_threads(
...
+    let h1 = spawn_thread_and_set_tag(42.into(), 84.into());
+    set_tag_for_current_task(50.into());
+    let h2 = spawn_thread_and_set_tag(50.into(), 100.into());
+    let results = [h1.join().unwrap(), h2.join().unwrap()];
+    assert_eq!(results, [84, 100]);
+}
+
+#[test]
+fn test_spawn_and_join() {
+    check_random(spawn_and_join, 20)
```
jorajeev, line 134: This could be `check_dfs`

src/current.rs
jorajeev: I was thinking we might want to use it in a custom scheduler (which does not have access to internal types, but is called from within Shuttle).

src/current.rs
sarsko: Lol, true, my bad

jorajeev (approved):

PR #101, opened 2023-04-07
Support spurious wakeups for `thread::park`

lukenels:
The documentation for `std::thread::park`
(https://doc.rust-lang.org/stable/std/thread/fn.park.html) has the following sentences regarding spurious wakeups:

> The `thread::park` function blocks the current thread unless or
> until the token is available for its thread handle, at which point
> it atomically consumes the token. It may also return spuriously,
> without consuming the token.

The documentation goes on to state:

> Notice that being unblocked does not imply any synchronization with
> someone that unparked this thread, it could also be spurious. For
> example, it would be a valid, but inefficient, implementation to
> make both park and unpark return immediately without doing anything.

Shuttle does not currently implement spurious wakeups for `thread::park`, making it possible for Shuttle to miss bugs in programs that (erroneously) rely on spurious wakeups not happening.

As an example, take the following code from
`tests/basic/thread.rs:thread_park`.  The test assumes that when `park` returns, the `flag` will have been set to true by the main thread. This is not correct since `park` can spuriously wakeup and should not be used for synchronization.  Shuttle does not find this case because it doesn't support spurious wakeups.

```rust
let flag = Arc::new(AtomicBool::new(false));
let thd = {
    let flag = Arc::clone(&flag);
    thread::spawn(move || {
        thread::park();
        assert!(flag.load(Ordering::SeqCst));
    })
};

flag.store(true, Ordering::SeqCst);
thd.thread().unpark();
thd.join().unwrap();
```

This commit adds support to Shuttle for spuriously waking up blocked threads. Not all causes of blocking can have spurious wakeups---for now, only `thread::park` is allowed to do so.

Spurious wakeup support is implemented by treating a subset of blocked tasks as being runnable from the perspective of the scheduler. However, unlike regular runnable threads, blocked threads that could be spuriously woken up do not count as being live for the purposes of deadlock detection. This is because spurious wakeups are not _guaranteed_ to happen, and correct programs should not rely on spurious wakeups to get out of deadlocks.

This commit updates the tests in `tests/basic/thread.rs` and adds a few new ones related to the spurious wakeup case. The erroneous test described above has been changed to `#[must_panic]`, an there is now a separate "fixed" version of the test that checks the flag in a loop.

Follow-up work:

- We should also change `Condvar::wait` to allow for spurious wakeups.

- We could change `thread::park_timeout` so that Shuttle won't treat it as a deadlock if there are threads blocked in `park_timeout` under the assumption the timeout will eventually fire. (This commit does not change the behavior compared to the previous logic, which treats `park_timeout` exactly like `park`.)

---

### DIFF
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -517,7 +517,7 @@ impl ExecutionState {
         }
 
         let mut unfinished_attached = false;
-        let runnable = self
+        let mut runnable = self
             .tasks
             .iter()
             .inspect(|t| unfinished_attached = unfinished_attached || (!t.finished() && !t.detached))
@@ -535,6 +535,13 @@ impl ExecutionState {
             return Ok(());
         }
 
+        // Some blocked tasks can be woken up spuriously, even though the condition the task is
+        // blocked on hasn't happened yet. We'll add such tasks to the list of runnable tasks, but
+        // they won't contribute to the check on `runnable.is_empty()` if the only runnable tasks
+        // are ones that are waiting for a potential spurious wakeup, it should still be treated as
+        // a deadlock since there's no guarantee that spurious wakeups will ever occur.
+        runnable.extend(self.tasks.iter().filter(|t| t.can_spuriously_wakeup()).map(|t| t.id));
+
         let is_yielding = std::mem::replace(&mut self.has_yielded, false);
 
         self.next_task = self
@@ -543,6 +550,17 @@ impl ExecutionState {
             .next_task(&runnable, self.current_task.id(), is_yielding)
             .map(ScheduledTask::Some)
             .unwrap_or(ScheduledTask::Stopped);
+
+        // If the task chosen by the scheduler is blocked, then it should be one that can be
+        // spuriously woken up, and we need to unblock it here so that it can execute.
+        if let Some(tid) = self.next_task.id() {
+            let task = self.get_mut(tid);
+            assert!(task.runnable() || task.blocked());
+            if task.blocked() {
+                assert!(task.can_spuriously_wakeup());
+                task.unblock();
+            }
+        }
 
         trace!(?runnable, next_task=?self.next_task);
 
--- a/src/runtime/task/mod.rs
+++ b/src/runtime/task/mod.rs
@@ -80,7 +80,7 @@ impl Task {
             waker,
             woken: false,
             detached: false,
-            park_state: ParkState::Unavailable,
+            park_state: ParkState::default(),
             name,
             local_storage: StorageMap::new(),
         }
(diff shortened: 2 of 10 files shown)


src/runtime/execution.rs
```diff
@@ -537,13 +537,31 @@ impl ExecutionState {
 
         let is_yielding = std::mem::replace(&mut self.has_yielded, false);
 
+        // Some blocked tasks can be woken up spuriously, even though the condition the task is
+        // blocked on hasn't happened yet. We'll add such tasks to the list of runnable tasks, but
+        // they won't contribute to the check on `runnable.is_empty()` if the only runnable tasks
+        // are ones that are waiting for a potential spurious wakeup, it should still be treated as
+        // a deadlock since there's no guarantee that spurious wakeups will ever occur.
+        runnable.extend(self.tasks.iter().filter(|t| t.can_spuriously_wakeup()).map(|t| t.id));
```
jamesbornholt, line 545: nit: I'd move this above the `let is_yielding = ...` line

src/runtime/task/mod.rs
```diff
@@ -278,22 +312,27 @@ pub(crate) enum TaskState {
...
+#[derive(PartialEq, Eq, Clone, Copy, Debug, Default)]
+pub(crate) struct ParkState {
+    /// Whether the task's park token is curently available. If it's available, then the next time
+    /// the task calls `park`, the token will be atomically consumed and the task will continue
+    /// executing. If it's not available, then the task will block until either another task makes
+    /// it available with `unpark`, or a spurious wakeup occurs.
+    token_available: bool,
+
+    /// Whether the task is currently blocked in a call to `park`.
+    /// Invariant: `!(token_available && parked)`. If the token is available, then the task
```
jamesbornholt, line 331: I think?
```suggestion
    /// Invariant: `!(token_available && blocked_in_park)`. If the token is available, then the task
```

tests/basic/thread.rs
```diff
@@ -478,6 +514,22 @@ fn thread_unpark_park() {
     )
 }
 
+// Test that `unpark` is not cumulative: multiple calls to `unpark` will still only unblock the very
+// next call to `park.
```
jamesbornholt, line 518:
```suggestion
// next call to `park`.
```

src/runtime/task/mod.rs
```diff
@@ -248,27 +264,45 @@ impl Task {
...
+            "task cannot park while already parked"
+        );
+        assert!(!self.blocked(), "task cannot park while blocked by something else");
+
+        if self.park_state.token_available {
+            self.park_state.token_available = false;
+            false
+        } else {
+            self.park_state.blocked_in_park = true;
+            self.block(true);
```
jamesbornholt, line 284: Should this also set `self.has_yielded`? Otherwise a PCT scheduler will test spurious wakeups (almost) every time, because PCT (almost) always schedules the same thread if it's still runnable and not yielding.

jamesbornholt: Er, `has_yielded` is in `ExecutionState`, but you get the idea.

lukenels: Yep, I added a call to `ExecutionState::request_yield()` in the `park()` handler where the `ExecutionState` is available.

jamesbornholt (approved):

PR #102, opened 2023-04-10
Fix an issue with `sync::Barrier` leader selection with reuse

lukenels:
The Rust standard library documentation for [`sync::Barrier`](https://doc.rust-lang.org/std/sync/struct.Barrier.html) states the following about `Barrier::wait`:

> Blocks the current thread until all threads have rendezvoused here.
> Barriers are re-usable after all threads have rendezvoused once,
> and can be used continuously.  A single (arbitrary) thread will
> receive a BarrierWaitResult that returns true from
> BarrierWaitResult::is_leader() when returning from this function,
> and all other threads will receive a result that will return false
> from BarrierWaitResult::is_leader().

This documentation is a bit ambiguous. It says that "a single thread" becomes the leader after returning from `wait`, but it isn't clear if that means a single thread for the existence of the `Barrier`, or a single thread for each "batch" of threads released by the barrier. The latter interpretation seems to make the most sense given the context of being able to reuse barriers, and it seems more useful to be able to assume that a leader is set for each batch. That is in fact the behavior implemented by the standard library as evidenced by [the implementation](https://doc.rust-lang.org/src/std/sync/barrier.rs.html#126-143) and this unit test, which reuses a barrier on a new thread:

```rust
let barrier = Barrier::new(1);
assert!(barrier.wait().is_leader());

let thd = thread::spawn(move || {
    assert!(barrier.wait().is_leader());
});

thd.join().unwrap();
```

Let's assume this behavior is correct, and that ambiguity in the Rust documentation would be resolved in favor of the existing implementation. Then the Shuttle behavior for `Barrier` is incorrect and causes this test to fail, because there is only a single, global `leader` field on the barrier which is set once the first time it's used.

This commit fixes the problem by adding the notion of an "epoch" to the `Barrier` implementation, which allows there to be different leaders for each batch / group of threads released by the barrier. (This is similar to the standard library implementation's notion of generation).

The commit also adds some new unit tests to capture the issue with the old behavior and prevent regressions. It also tweaks `Barrier::wait` to become an explicit thread yield point, in order to vary which thread becomes the leader rather than just choosing the first one that happens to call `wait.

---

### DIFF
--- a/src/sync/barrier.rs
+++ b/src/sync/barrier.rs
@@ -20,11 +20,38 @@ impl BarrierWaitResult {
     }
 }
 
+/// We implement [Barrier] by keeping track of a list of [TaskId]s of threads that have called
+/// `wait`. When the numbers of waiters gets to the barrier's `bound`, they all become unblocked.
+/// Whichever task wakes up first after becoming unblocked will be designated as the "leader" via
+/// the return value of `wait`.
+///
+/// Because barriers can be reused, it does not suffice to designate a single, permanent thread as
+/// the leader in the [BarrierState], since if a different batch of threads waits on the same
+/// barrier, one of those new threads should be the leader of that batch.
+///
+/// For example, if there's a barrier where the bound is 2 and there are 6 threads all concurrently
+/// calling `wait`, then all threads will become unblocked in 3 batches of 2 threads each, with
+/// each batch having its own leader, resulting in 3 (unique) leaders.
+///
+/// We implement this by tracking an `epoch` counter that gets incremented whenever a batch of
+/// threads is released by `wait`. When that happens, we make a "leader token_" available for that
+/// batch. The first thread of a batch to get scheduled after becoming unblocked takes the leader
+/// token (without affecting any threads that are part of a different batch).
 #[derive(Debug)]
 struct BarrierState {
+    /// The number of tasks that must call `wait` before they all get unblocked (and one gets
+    /// chosen as the leader).
     bound: usize,
-    leader: Option<TaskId>,
+    /// A counter of the number of "batches" of threads that have been released by the barrier,
+    /// needed in order to keep track of the leaders of each batch separately.
+    epoch: u64,
+    /// The set of waiting tasks for the current epoch. Then the size of this set becomes equal to
+    /// the bound, all waiters are unblocked and one of them becomes the leader.
     waiters: HashSet<TaskId>,
+    /// The set of epochs of this [Barrier] that have reached the number of waiters to be
+    /// unblocked, but haven't had a leader selected yet. The first thread to wake up after wait
+    /// will take the token (by removing the epoch from the set) and become the leader.
+    leader_tokens: HashSet<u64>,
     clock: VectorClock,
 }
 
@@ -41,8 +68,9 @@ impl Barrier {
     pub fn new(n: usize) -> Self {
         let state = BarrierState {
             bound: n,
-            leader: None,
+            epoch: 0,
             waiters: HashSet::new(),
+            leader_tokens: HashSet::new(),
             clock: VectorClock::new(),
         };
 
@@ -53,19 +81,10 @@ impl Barrier {
 
     /// Blocks the current thread until all threads have rendezvoused here.
     pub fn wait(&self) -> BarrierWaitResult {
-        let me = ExecutionState::me();
-
         let mut state = self.state.borrow_mut();
-        trace!(leader=?state.leader, waiters=?state.waiters, "waiting on barrier {:p}", self);
+        let my_epoch = state.epoch;
 
-        // TODO The documentation for `Barrier` states that
-        // TODO    A single (arbitrary) thread will receive a `BarrierWaitResult` that returns true
-        // TODO    from `BarrierWaitResult::is_leader()` when returning from this function
-        // TODO Currently, the first waiter becomes the leader.  We should use Shuttle's nondeterministic
-        // TODO choice to generalize this so that _any_ participant could become the leader.
-        if state.leader.is_none() {
-            state.leader = Some(me);
-        }
+        trace!(waiters=?state.waiters, epoch=my_epoch, "waiting on barrier {:p}", self);
 
         // Update the barrier's clock with the clock of this thread
         ExecutionState::with(|s| {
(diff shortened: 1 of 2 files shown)


src/sync/barrier.rs
```diff
@@ -56,57 +67,77 @@ impl Barrier {
...
-            is_leader: state.leader.unwrap() == me,
         };
 
         drop(state);
 
+        // Mark the task as requesting an explicit yield and schedule a thread switch, otherwise
+        // some schedulers would nearly always pick to continue executing the current task.
+        // Requesting a yield allows other threads to be the first one to wake up and grab the
+        // leader token.
+        ExecutionState::request_yield();
```
jamesbornholt, line 131: Unlike #101 I think we don't want to yield in this case. We needed it there for fairness because `park` gets used in spinloops, which are doomed by unfair schedules. But barriers aren't used like that—threads generally do useful work once unblocked by a barrier. (I do wish we had better rules for thinking about this, but it's all random in the end anyway, so it gets a bit handwave-y).

src/sync/barrier.rs
```diff
@@ -56,57 +67,77 @@ impl Barrier {
...
                     t.clock.increment(tid);
                     t.clock.update(&clock);
-                    t.unblock();
+
+                    // A thread blocked on the barrier will be blocked iff it's not the current
+                    // one, since this is the last thread arriving at the barrier for this epoch.
+                    assert_eq!(t.blocked(), tid != me);
+                    if t.blocked() {
+                        t.unblock();
+                    }
```
jamesbornholt, line 120: Don't think this needs the guard (`unblock` doesn't check).
```suggestion
                    t.unblock();
```

src/sync/barrier.rs
```diff
@@ -56,57 +67,77 @@ impl Barrier {
...
+
+            debug_assert!(state.waiters.len() == state.bound || state.bound == 0);
+
+            // Make the leader token available for this epoch. The first task to wake up will
+            // take it and become the leader. The token shouldn't already be available.
+            assert!(state.leader_tokens.insert(my_epoch));
+
+            // Drain the set of waiters and increment the barrier's epoch, so any other task that
+            // calls `wait` from now on becomes part of a separate group with its own leader.
+            let waiters: Vec<TaskId> = state.waiters.drain().collect();
```
jamesbornholt, line 97:
```suggestion
            let waiters: Vec<_> = state.waiters.drain().collect();
```

src/sync/barrier.rs
```diff
@@ -56,57 +67,77 @@ impl Barrier {
...
         ExecutionState::with(|s| {
             let clock = s.increment_clock();
             state.clock.update(clock);
         });
 
-        if state.waiters.len() + 1 < state.bound {
-            // Block current thread
-            assert!(state.waiters.insert(me)); // current thread shouldn't already be in the set
+        // The current thread shouldn't already be in `waiters`.
+        assert!(state.waiters.insert(me));
```
jamesbornholt, line 81: I think I marginally preferred the existing code that didn't needlessly add the current thread to the waiters set, so that the invariant was that all waiters are blocked. But we should still assert here that `me` isn't already a waiter.

src/sync/barrier.rs
```diff
@@ -22,9 +22,19 @@ impl BarrierWaitResult {
```
jamesbornholt, line 22: Could you write a comment like the one we have for Condvar explaining how/why the implementation works?

https://github.com/awslabs/shuttle/blob/94b9d8bc87b0cb11c213cccdb45240e32a16e6eb/src/sync/condvar.rs#L40

It should probably be a doc comment on `BarrierState`? (the Condvar one should probably have been a doc comment on `CondvarState`).

jamesbornholt (commented):
I ended up writing this test because I wasn't sure about the case where `1 < barrier_size < num_threads`; I think it subsumes some of yours:

```rust
// Test that any combination of threads can be chosen as leader of an epoch.
fn barrier_test_nondeterministic_leader(batch_size: usize, num_threads: usize) {
    // Don't test cases that might deadlock; every thread needs to be released eventually
    assert!(num_threads % batch_size == 0);

    let seen_leaders = Arc::new(std::sync::Mutex::new(HashSet::new()));

    {
        let seen_leaders = Arc::clone(&seen_leaders);
        check_dfs(
            move || {
                let barrier = Arc::new(Barrier::new(batch_size));
                let leaders = Arc::new(std::sync::Mutex::new(vec![]));
                let handles = (0..num_threads).map(|i| {
                    let barrier = Arc::clone(&barrier);
                    let leaders = Arc::clone(&leaders);

                    thread::spawn(move || {
                        let result = barrier.wait();
                        if result.is_leader() {
                            leaders.lock().unwrap().push(i);
                        }
                    })
                }).collect::<Vec<_>>();

                for thd in handles {
                    thd.join().unwrap();
                }

                let leaders = Arc::try_unwrap(leaders).unwrap().into_inner().unwrap();
                seen_leaders.lock().unwrap().insert(leaders);
            },
            None,
        );
    }

    let seen_leaders = Arc::try_unwrap(seen_leaders).unwrap().into_inner().unwrap();
    let num_batches = num_threads / batch_size;
    let fact = |i: usize| (2..i+1).reduce(|acc, i| acc * i).unwrap_or(1);

    // Every execution should have one leader per batch
    assert!(seen_leaders.iter().all(|l| l.len() == num_batches));
    // There should be `num_threads permute num_batches` different leader vectors
    assert_eq!(seen_leaders.len(), fact(num_threads) / fact(num_threads - num_batches));
}

#[test]
fn barrier_test_nondeterministic_leader_1() {
    barrier_test_nondeterministic_leader(1, 4);
}

#[test]
fn barrier_test_nondeterministic_leader_2() {
    barrier_test_nondeterministic_leader(2, 4);
}

#[test]
fn barrier_test_nondeterministic_leader_4() {
    barrier_test_nondeterministic_leader(4, 4);
}
```

src/sync/barrier.rs
lukenels: >  It ends up having roughly the same number of special cases for me, just different ones.

What special cases for `me` do you mean? Assuming I kill `assert_eq!(t.blocked(), tid != me)` (which is really just for debugging), then the only use of `me` in the function is in the line you commented on here which I could just inline as:

```rust
assert!(state.waiters.insert(ExecutionState::me()));
```

---

> [...] the invariant was that all waiters are blocked.

Arguably, that still is an invariant. The barrier's `state` is never release while this isn't true (and every thread in `waiters` is always one that called `wait`).

src/sync/barrier.rs
lukenels: My intention here was to prevent an unfair schedule from making it so that it's (almost) always the last thread that gets to the barrier that becomes the leader, by giving every other thread a chance to wake up first and become the leader instead. In most situations I can think of though, it's probably already random enough which thread happens to call wait last, so the "extra" fairness here won't help much. I'll remove the `request_yield` here.

jamesbornholt (approved):

Issue #100, opened 2023-04-07
Failing schedule missing in output


bkragl:
I've seen the following failure output in CI runs. Without the failing schedule I'm unable to reproduce the failure.

```
thread 'my_shuttle_test' panicked at 'exceeded max_steps bound 1000000. this might be caused by an unfair schedule (e.g., a spin loop)?
pass that string to `shuttle::replay` to replay the failure', /path/to/shuttle/src/runtime/execution.rs:143:17
stack backtrace:
   ...
```

I'm quite baffled, since the error message must come from
https://github.com/awslabs/shuttle/blob/0310685fb9e4dcdacbc9b58f6f8a196b1e4dec94/src/runtime/failure.rs#L74-L77
and I don't understand how the failing schedule part can be lost.

bkragl:
OK, turns out this is an issue with the CI swallowing log lines, not a Shuttle issue.

PR #103, opened 2023-04-10
Split schedule output over multiple lines

bkragl:
Failing schedules can be long, and printing them on a single long line can be a problem (e.g., #100).

---

### DIFF
--- a/src/runtime/failure.rs
+++ b/src/runtime/failure.rs
@@ -72,7 +72,7 @@ fn persist_failure_inner(schedule: &Schedule, message: String, config: &Config) 
         }
     }
     format!(
-        "{}\nfailing schedule: \"{}\"\npass that string to `shuttle::replay` to replay the failure",
+        "{}\nfailing schedule:\n\"\n{}\n\"\npass that string to `shuttle::replay` to replay the failure",
         message, serialized_schedule
     )
 }
--- a/src/scheduler/serialization.rs
+++ b/src/scheduler/serialization.rs
@@ -86,6 +86,8 @@ mod varint {
 
 const SCHEDULE_MAGIC_V2: u8 = 0x91;
 
+const LINE_WIDTH: usize = 76;
+
 pub(crate) fn serialize_schedule(schedule: &Schedule) -> String {
     use self::varint::{space_needed, WriteVarInt};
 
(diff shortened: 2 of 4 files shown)


src/scheduler/serialization.rs
```diff
@@ -128,12 +131,23 @@ pub(crate) fn serialize_schedule(schedule: &Schedule) -> String {
...
+    let num_line_feeds = (buf.len() + BYTES_PER_LINE - 1) / BYTES_PER_LINE;
+    let mut hex_buf = vec![0; buf.len() * 2 + num_line_feeds];
+    let mut offset = 0;
+    for chunk in buf.chunks(BYTES_PER_LINE) {
+        hex::encode_to_slice(chunk, &mut hex_buf[offset..offset + chunk.len() * 2]).unwrap();
+        offset += chunk.len() * 2;
+        hex_buf[offset] = LINE_FEED;
+        offset += 1;
+    }
+    String::from_utf8(hex_buf).unwrap()
```
jamesbornholt, line 144: This is a little worse allocation-wise, but probably clearer, and doesn't add a new line at the end:
```suggestion
    let serialized = hex::encode(buf);
    let lines = serialized.as_bytes().chunks(BYTES_PER_LINE).collect::<Vec<_>>();
    let wrapped = lines.join(&['\n' as u8][..]);
    String::from_utf8(wrapped).unwrap()
```

src/runtime/failure.rs
```diff
@@ -72,7 +72,7 @@ fn persist_failure_inner(schedule: &Schedule, message: String, config: &Config)
         }
     }
     format!(
-        "{}\nfailing schedule: \"{}\"\npass that string to `shuttle::replay` to replay the failure",
+        "{}\nfailing schedule:\n{}pass that string to `shuttle::replay` to replay the failure",
```
jamesbornholt, line 75: Maybe something like this to really highlight where the string starts/ends:
```suggestion
        "{}\nfailing schedule:\n\"\n{}\n\"\npass that string to `shuttle::replay` to replay the failure",
```

src/scheduler/serialization.rs
bkragl: Haha, I contemplated descending to unsafe to skip zeroing the buffer and using `from_utf8_unchecked`. But fair enough, this is not performance critical, so accepting the extra allocations should be fine. Same as the extra allocation we now have in `deserialize_schedule` to remove newlines.

jamesbornholt (approved):

PR #105, opened 2023-05-02
Fix FuturesUnordered

sarsko:
Currently, code using FuturesUnordered has a chance to cause Shuttle to erroneously report a deadlock.

For instance, if we take the following code:
```rust
  shuttle::future::block_on(async {
        let mut tasks = (0..100).map(|_| spawn(async move {})).collect::<FuturesUnordered<_>>();

        while let Some(result) = tasks.next().await {
            result.unwrap();
        }
    });
```
When we call `next()` on a task, that task gets dequeued from the `ready_to_run_queue` inside `FuturesUnordered`. If that task is `Poll::Ready`, then all is fine. If it is `Poll::Pending` (in the case where we call `next()` before the `spawn` has had a chance to execute) however, then the task is never polled again, as we don't wake the associated waker.

This PR seeks to solve this issue by waking the waker, which is stored together with the result in the `Wrapper`.

Currently the test suite is a bit lacking, and only contains three tests of common patterns with futures that don't do anything.  I have some more tests locally, but they are really just reskins of existing tests, so I am not entirely sure they check any interesting behaviour. I want to add some more interesting tests, so going to look at that next. If anyone has any suggestions, then I am all ears. Also putting this out here to get feedback on the solution in general.

---

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -11,9 +11,8 @@ use crate::runtime::thread;
 use std::future::Future;
 use std::pin::Pin;
 use std::result::Result;
-use std::task::{Context, Poll};
-
-type ResultAndWaker<T> = std::sync::Arc<std::sync::Mutex<(Option<Result<T, JoinError>>, Option<std::task::Waker>)>>;
+use std::sync::Arc;
+use std::task::{Context, Poll, Waker};
 
 /// Spawn a new async task that the executor will run to completion.
 pub fn spawn<T, F>(fut: F) -> JoinHandle<T>
@@ -21,27 +20,35 @@ where
     F: Future<Output = T> + Send + 'static,
     T: Send + 'static,
 {
-    let result_and_waker = std::sync::Arc::new(std::sync::Mutex::new((None, None)));
     let stack_size = ExecutionState::with(|s| s.config.stack_size);
-    let task_id = ExecutionState::spawn_future(
-        Wrapper::new(fut, std::sync::Arc::clone(&result_and_waker)),
-        stack_size,
-        None,
-    );
+    let inner = Arc::new(std::sync::Mutex::new(JoinHandleInner::default()));
+    let task_id = ExecutionState::spawn_future(Wrapper::new(fut, inner.clone()), stack_size, None);
 
     thread::switch();
 
-    JoinHandle {
-        task_id,
-        result_and_waker,
-    }
+    JoinHandle { task_id, inner }
 }
 
 /// An owned permission to join on an async task (await its termination).
 #[derive(Debug)]
 pub struct JoinHandle<T> {
     task_id: TaskId,
-    result_and_waker: ResultAndWaker<T>,
+    inner: std::sync::Arc<std::sync::Mutex<JoinHandleInner<T>>>,
+}
+
+#[derive(Debug)]
+struct JoinHandleInner<T> {
+    result: Option<Result<T, JoinError>>,
+    waker: Option<Waker>,
+}
+
+impl<T> Default for JoinHandleInner<T> {
+    fn default() -> Self {
+        JoinHandleInner {
+            result: None,
+            waker: None,
+        }
+    }
 }
 
 impl<T> JoinHandle<T> {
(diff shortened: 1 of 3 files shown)


src/future/mod.rs
```diff
@@ -128,6 +139,10 @@ where
                     }
                 });
```
jamesbornholt, line 140: I think this "Unblock our waiter if we have one and it's still alive" code can just be deleted, along with the `set_waiter` call in this module, right? That code is just wrong -- we should be invoking the waker for whoever's waiting for us, not this weird special-case thing. I think we just copied and pasted it from the `thread` implementation, which doesn't have access to a waker and so has to build its own waker-esque mechanism with `set_waiter`/`take_waiter`.

jamesbornholt (commented):
In terms of unit tests, I think the thing we're missing is polling the JoinHandle with a custom waker implementation -- things accidentally work today if the JoinHandle is polled with a waker provided by Shuttle, because it defers to the internal `set_waiter` mechanism. So we want a unit test that somehow validates the waker is invoked when it's supposed to be.

tests/future/streams.rs
```diff
@@ -0,0 +1,50 @@
...
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, _>>()
+            .unwrap();
+    });
+}
+
+#[test]
+fn join_all_empty_tasks() {
+    check_random(futures_join_all, 100);
```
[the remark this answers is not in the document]
jorajeev: +1 You should reduce the #tasks to the smallest number (think 1 or 2 should work) -- and ensure that the tests fail without the fix.

tests/future/streams.rs
```diff
@@ -0,0 +1,50 @@
...
+#[test]
+fn collect_empty_tasks() {
+    check_random(futures_unordered_collect, 100);
+}
+
+fn futures_unordered_next() {
+    block_on(async {
+        let mut tasks = (0..100).map(|_| spawn(async move {})).collect::<FuturesUnordered<_>>();
+
+        while let Some(result) = tasks.next().await {
```
jorajeev, line 24: Not sure this test really adds anything new, since the above collect essentially does the same thing, right?

sarsko: Hmm, kinda, the internals of `Next` and `Collect` are a bit different. I think it is fine to have both.

src/future/mod.rs
```diff
@@ -74,37 +81,33 @@ impl<T> Future for JoinHandle<T> {
...
+        if let Some(result) = lock.result.take() {
             Poll::Ready(result)
         } else {
-            ExecutionState::with(|state| {
-                let me = state.current().id();
-                let r = state.get_mut(self.task_id).set_waiter(me);
-                assert!(r, "task shouldn't be finished if no result is present");
-            });
-            *lock = (None, Some(cx.waker().clone()));
+            lock.waker = Some(cx.waker().clone());
```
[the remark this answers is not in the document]
sarsko: Removed the entire setting of the waiter mechanism. Can readd the assertion.

src/future/mod.rs
```diff
@@ -11,28 +11,44 @@ use crate::runtime::thread;
...
+    waker: Option<Waker>,
+}
+
+impl<T> Default for JoinHandleInner<T> {
+    fn default() -> Self {
+        JoinHandleInner {
+            result: None,
+            waker: None,
+        }
+    }
```
jamesbornholt, line 51: nit: this can just be derived

src/future/mod.rs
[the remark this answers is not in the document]
jamesbornholt: The assert doesn't make sense any more (we don't expect the waiter to get set in this case).

src/future/mod.rs
```diff
@@ -117,16 +131,11 @@ where
...
-                ExecutionState::with(|state| {
-                    if let Some(waiter) = state.current_mut().take_waiter() {
-                        if !state.get_mut(waiter).finished() {
-                            state.get_mut(waiter).unblock();
-                        }
-                    }
-                });
+                if let Some(waker) = self.inner.lock().unwrap().waker.take() {
+                    waker.wake();
+                }
```
jamesbornholt, line 138: don't take the lock twice here?

jorajeev (approved):

sarsko:
Just a heads up that a fix with the double locking fixed and a test for the custom waker is coming.

src/future/mod.rs
jamesbornholt: Oh right, this is the stupid thing where the derive macro always emits a `T: Default` bound even if it's not required.

sarsko:
Hmm, it can be reapproved and merged now. The behaviour the test is intended to cover is currently exercised by FuturesUnordered through the existing tests. Can add the test later as we don't want to rely on FuturesUnordered internals.

PR #121, opened 2023-08-07
Add `as_any` to Tag trait

progwriter:
This enables downcasting dynamic trait to concrete types to get detailed tag info from a task

---

### DIFF
--- a/src/runtime/task/mod.rs
+++ b/src/runtime/task/mod.rs
@@ -44,7 +44,19 @@ pub(crate) const DEFAULT_INLINE_TASKS: usize = 16;
 /// identify tasks in failing Shuttle tests. A task's [Tag] can be set with the
 /// [set_tag_for_current_task](crate::current::set_tag_for_current_task) function. Newly spawned
 /// threads and futures inherit the tag of their parent at spawn time.
-pub trait Tag: Debug {}
+pub trait Tag: Debug {
+    /// Return the tag as `Any`, typically so that it can be downcast to a known concrete type
+    fn as_any(&self) -> &dyn Any;
+}
+
+impl<T> Tag for T
+where
+    T: Debug + Any,
+{
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
 
 /// A `Task` represents a user-level unit of concurrency. Each task has an `id` that is unique within
 /// the execution, and a `state` reflecting whether the task is runnable (enabled) or not.
--- a/tests/basic/tag.rs
+++ b/tests/basic/tag.rs
@@ -19,8 +19,6 @@ use tracing::{Event, Id, Metadata, Subscriber};
 #[derive(PartialEq, Eq, Clone, Copy, Debug, Default, Hash, PartialOrd, Ord)]
 pub struct Tag(u64);
 
-impl shuttle::current::Tag for Tag {}
-
 impl From<u64> for Tag {
     fn from(tag: u64) -> Self {
         Tag(tag)
@@ -76,12 +74,15 @@ fn spawn_some_threads_and_set_tag<F: (Fn(Tag, u64) -> Tag) + Send + Sync>(
 }
 
 fn convert_to_tag(tag: Arc<dyn shuttle::current::Tag>) -> Tag {
-    let ptr = Arc::into_raw(tag).cast::<Tag>();
-    *unsafe { Arc::from_raw(ptr) }
+    *tag.as_any().downcast_ref::<Tag>().unwrap()
 }
 
 fn curr_tag() -> Tag {
-    convert_to_tag(get_tag_for_current_task().unwrap())
+    *get_tag_for_current_task()
+        .unwrap()
+        .as_any()
+        .downcast_ref::<Tag>()
+        .unwrap()
 }
 
 fn spawn_threads_which_spawn_more_threads(num_threads_first_block: u64, num_threads_second_block: u64) {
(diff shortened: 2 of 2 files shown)


sarsko:
Thanks. 

I'm not sure I see why this is needed / what the gain is ?

Isn't the functionality of regaining the concrete type already possible with ie (from the tag test):
```rust
fn convert_to_tag(tag: Arc<dyn shuttle::current::Tag>) -> Tag {
    let ptr = Arc::into_raw(tag).cast::<Tag>();
    *unsafe { Arc::from_raw(ptr) }
}
```
?

progwriter:
Ah, I missed the conversion you have in place. Is there a way to do it without resorting to using an `unsafe` block? If yes, happy to adopt that.

My understanding with `Any` was that it enabled a safer option via [`downcast_ref`](https://doc.rust-lang.org/std/any/trait.Any.html#method.downcast_ref), so elsewhere in my code I'd do 
`t.as_any().downcast_ref::<MyTag>()` which returns an `Option<MyTag>` if the cast is successful.

sarsko:
The way it is done currently requires unsafe as far as I am aware.

I'm not entirely opposed to adding the `as_any` to `Tag`, it should be more convenient and it is also better to fail on an `unwrap` than on some memory violation issue.

That said, I'm not seeing when the cast would fail — do you currently have a use case for non-homogeneous tags?

Another option is to add the `+ Any` trait bound to `Tag`, which has the advantage that creating a `Tag` requires slightly less code, but the disadvantage that you don't get the `.as_any()`-nicety. It would enable checking type equality and graceful failure.

A note is that the `from_raw` unsafety is fully checkable. The preconds are that the raw pointer has to have been gotten from `into_raw`, and the type has to have the correct alignment and size (this can be checked with `align_of_val` and `size_of_val`).

sarsko:
We could also add a blanket implementation:
```rust
pub trait Tag: Debug {
    /// Return the tag as `Any`, typically so that it can be downcast to a known concrete type
    fn as_any(&self) -> &dyn std::any::Any;
}

impl<T> Tag for T where T: Debug + Any {
    fn as_any(&self) -> &dyn std::any::Any {
        self
    }
}
```
as there should not be any other way to define `as_any`. Of course, that makes all types which are `Debug` + `Any` (which most types are) valid to use as tags, which in a sense loses the "typedness" of tags, but I guess that can be reestablished by wrapping `set_tag_for_current_task` anyways.

Dunno, I have no strong preferences either way.

progwriter:
I quite like your idea of blanket implementation. I'll adopt that.

> that makes all types which are Debug + Any (which most types are) valid to use as tags, which in a sense loses the "typedness" of tags

I'm not quite sure I follow, is that not the case already? Any struct can be a tag if it implements `Tag` for which only requirement is `Debug`

sarsko:
> I quite like your idea of blanket implementation. I'll adopt that.
> 
> > that makes all types which are Debug + Any (which most types are) valid to use as tags, which in a sense loses the "typedness" of tags
> 
> I'm not quite sure I follow, is that not the case already? Any struct can be a tag if it implements `Tag` for which only requirement is `Debug`

Ah yeah, sorry, poorly explained. Without the blanket implementation we will get a compilation error if something which has not implemented `Tag` manually is used as a tag, whereas adding the blanket implementation opens up for accidentally passing something which was not intended to be used as a tag as a tag. Since most types are `Debug` + `Any`, we go from default "nothing is allowed to be used as a tag except those which have manually implemented `Tag`" to default "basically everything is allowed to be used as a tag, and we won't do any checks".

src/runtime/task/mod.rs
```diff
@@ -44,7 +44,16 @@ pub(crate) const DEFAULT_INLINE_TASKS: usize = 16;
...
 /// [set_tag_for_current_task](crate::current::set_tag_for_current_task) function. Newly spawned
 /// threads and futures inherit the tag of their parent at spawn time.
-pub trait Tag: Debug {}
+pub trait Tag: Debug {
+    /// Return the tag as `Any`, typically so that it can be downcast to a known concrete type
+    fn as_any(&self) -> &dyn Any;
+}
+
+impl<T> Tag for T where T: Debug + Any {
+    fn as_any(&self) -> &dyn std::any::Any {
```
jorajeev, line 53: Why the full import `std::any::Any` when you already did a `use std::any::Any` above?

jorajeev (commented):
As part of this change, can you remove the `convert_to_tag` function entirely?  Just use `downcast_ref` instead?

progwriter:
Addressed comments.

* The imports: that was just my IDE being silly, standardized on the already existing pattern, using the import not fully qualifying. 
* I tried removing `convert_to_tag` but that made a number of call sites really verbose, making code harder to read. So I kept the function but removed the `unsafe` manipulations and replaced them with `downcast_ref`. If you object I can do another revision and inline the `convert_to_tag`

sarsko:
Didn't think of this until now, but one way to undo the 'everything is now `Tag`' change is to introduce a new marker trait and have `Tag` extend that. I added a "suggestion PR" here: https://github.com/awslabs/shuttle/pull/124.

I don't feel strongly about this, but think I prefer to introduce the marker trait. It is slightly convenient (and more pretty) to not have it, but I think I (and/or someone else) is going to accidentally tag something with the wrong type and then figure that out at runtime instead of at compile time and then have to go through all of the places tags are set in order to find the source, and I think that inconvenience outweighs the slight convenience of not having to do `impl shuttle::current::Taggable for X {}`.

progwriter:
That's great, thank you! Let's just merge your PR instead

jorajeev:
Closing this, as we merged https://github.com/awslabs/shuttle/pull/124

PR #128, opened 2023-11-27
Tracing fix

sarsko:
I want to add a bit more testing. At least something fairly minimal like connecting the tests to a `tracing_subscriber::fmt` which outputs to something else than stdout and then check on that, and a simple "X threads enter and exit Y spans and assert that the current span is what they expect" test.

Opening a PR in order to get feedback and suggestions. Feel free to suggest alternative test ideas as well. Would be nice to do a programmatic check against Tokio / no runtime as well.

# Fixing tracing

I think I have done it right this time. One known divergence from Tokio (listed below), might be more, but I think the general scheme should work. Also adds a fix for spans stacking on test panic inside a `catch_unwind` (read: proptest minimization).

## Overview of the changes

- New config option, `record_steps_in_span`. 
     - The `step{i=X, task=Y}` span is now  `step{task=Y}` by default, and `step{task=Y, i=X}` if `record_steps_in_span` is `true`
- Updating tracing span is no longer done in `advance_to_next_task`. It is instead done in `step`, by entering a `Task`s span before continuing its continuation, and leaving the Span when we return from the continuation.
- Added `ResetSpanOnDrop` which is constructed in `run`. This is to solve the issue of span stacking on panics inside a `catch_unwind` (read: proptest minimization). I found this way of solving it much cleaner than finding all the places we could exit a Shuttle test and making sure to clean up spans there.
- A couple of traces which were previously emitted in whatever span was present (ie the span of a `Task`) are now emitted in (an extension of) Shuttle's span.

## Some notes

- I mentioned that I considered doing a "default to no span" approach, meaning that Shuttle would exit whatever span was present on test entry, store that, and then have an explicit enter/leave policy for all tracing, thus avoiding all accidental "span-bleed". I decided against this because ensuring that the correct span would be entered, for instance in `thread_fn` was cumbersome and thus verbose, and thus more error-prone than using whatever trace was present. This means there may be traces emitted where we would prefer to use a different span than the one we are using currently (as I have not looked into every single `Event` emission), these can be handled as we find them.
- The `in_scope` in `spawn_thread` and `Task::new()` are matters of taste — I prefer for these to be emitted under Shuttle's span, but, they are reactions to actions taken by the function under test, so there is an argument for emitting them under whatever span is currently entered.

## Known issues / discrepancies

Below are some tests which do the same thing using either Shuttle `sync`/`future` threads, `std::sync` threads and `tokio::spawn` threads. Shuttle `sync`/`future` and `std::sync` does the same thing, `tokio::spawn` does something different. I think the thing does currently is sane enough (arguably better than the mess you may get when running under Tokio — this is not how to combine futures and spans).

### Tokio
```rust
async fn awaiting_is_relevant() {
    let lock = Arc::new(Mutex::new(0));
    let threads: Vec<_> = (0..3).map(|i| {
        let lock = lock.clone();
        tokio::spawn(async move {
            let outer = warn_span!("outer", tid=i);
            let _outer = outer.enter();

            let lock_clone = lock.clone();
            let t = tokio::spawn(async move 
            {
                let mut locked = lock_clone.lock().unwrap();
                let inner = warn_span!("inner1", tid=i);
                let _inner = inner.enter();
                error!("incrementing from {}", *locked);
                *locked += 1;
            });

            t.await.unwrap();

            let _t = tokio::spawn(async move 
            {
                let mut locked = lock.lock().unwrap();
                let inner = warn_span!("inner2", tid=i);
                let _inner = inner.enter();
                error!("incrementing from {}", *locked);
                *locked += 1;
            });

        })
    }).collect();

    for thread in threads {
        thread.await.unwrap();
    }
}

#[tokio::test]
async fn tokio_awaiting_is_relevant() {
    let _ = tracing_subscriber::fmt::try_init();
    awaiting_is_relevant().await;
}
```
Doing `RUST_LOG=warn cargo test tokio_awaiting` gives:
```
running 1 test
2023-11-27T16:33:30.579493Z ERROR outer{tid=0}:outer{tid=1}:outer{tid=2}:inner1{tid=0}: play: incrementing from 0
2023-11-27T16:33:30.579761Z ERROR outer{tid=0}:outer{tid=1}:outer{tid=2}:inner1{tid=1}: play: incrementing from 1
2023-11-27T16:33:30.579794Z ERROR outer{tid=0}:outer{tid=1}:outer{tid=2}:inner1{tid=2}: play: incrementing from 2
2023-11-27T16:33:30.579850Z ERROR inner2{tid=0}: play: incrementing from 3
2023-11-27T16:33:30.579869Z ERROR inner2{tid=1}: play: incrementing from 4
2023-11-27T16:33:30.579885Z ERROR inner2{tid=2}: play: incrementing from 5
test tokio_awaiting_is_relevant ... ok
```
As we can see, doing `t.await.unwrap();` for the first `t` makes it execute under the parent spans, and not doing it for the second `t` makes those have no parent.

### Shuttle `future::spawn`
```rust
async fn shuttle_awaiting_is_irrelevant() {
    let lock = Arc::new(Mutex::new(0));
    let threads: Vec<_> = (0..3)
        .map(|i| {
            let lock = lock.clone();
            shuttle::future::spawn(async move {
                let outer = warn_span!("outer", tid = i);
                let _outer = outer.enter();

                let lock_clone = lock.clone();
                let t = shuttle::future::spawn(async move {
                    let mut locked = lock_clone.lock().unwrap();
                    let inner = warn_span!("inner1", tid = i);
                    let _inner = inner.enter();
                    error!("incrementing from {}", *locked);
                    *locked += 1;
                });

                t.await.unwrap();

                let _t = shuttle::future::spawn(async move {
                    let mut locked = lock.lock().unwrap();
                    let inner = warn_span!("inner2", tid = i);
                    let _inner = inner.enter();
                    error!("incrementing from {}", *locked);
                    *locked += 1;
                });
            })
        })
        .collect();

    for thread in threads {
        thread.await.unwrap();
    }
}

#[test]
fn shuttle_test_awaiting_is_irrelevant() {
    let _ = tracing_subscriber::fmt::try_init();
    shuttle::check_random(
        move || {
            shuttle::future::block_on(async {
                shuttle_awaiting_is_irrelevant().await;
            })
        },
        1,
    );
}
```
Doing `RUST_LOG=warn cargo test shuttle_test_awaiting` gives:
```
running 1 test
2023-11-27T16:45:55.192800Z ERROR inner1{tid=0}: r#mod::basic::tracing: incrementing from 0
2023-11-27T16:45:55.194048Z ERROR inner1{tid=1}: r#mod::basic::tracing: incrementing from 1
2023-11-27T16:45:55.199122Z ERROR inner2{tid=0}: r#mod::basic::tracing: incrementing from 2
2023-11-27T16:45:55.200569Z ERROR inner2{tid=1}: r#mod::basic::tracing: incrementing from 3
2023-11-27T16:45:55.200632Z ERROR inner1{tid=2}: r#mod::basic::tracing: incrementing from 4
2023-11-27T16:45:55.200699Z ERROR inner2{tid=2}: r#mod::basic::tracing: incrementing from 5
test basic::tracing::shuttle_test_awaiting_is_irrelevant ... ok
```
As we can see, this differs from what we saw when running under Tokio.

### No runtime `std::thread::spawn`
```rust
#[test]
fn std_thread_joining_is_irrelevant() {
    let _ = tracing_subscriber::fmt::try_init();
    let lock = Arc::new(std::sync::Mutex::new(0));
    let threads: Vec<_> = (0..3)
        .map(|i| {
            let lock = lock.clone();
            std::thread::spawn(move || {
                let outer = warn_span!("outer", tid = i);
                let _outer = outer.enter();

                let lock_clone = lock.clone();
                let t = std::thread::spawn(move || {
                    let mut locked = lock_clone.lock().unwrap();
                    let inner = warn_span!("inner1", tid = i);
                    let _inner = inner.enter();
                    error!("incrementing from {}", *locked);
                    *locked += 1;
                });

                t.join().unwrap();

                let _t = std::thread::spawn(move || {
                    let mut locked = lock.lock().unwrap();
                    let inner = warn_span!("inner2", tid = i);
                    let _inner = inner.enter();
                    error!("incrementing from {}", *locked);
                    *locked += 1;
                });
            })
        })
        .collect();

    for thread in threads {
        thread.join().unwrap();
    }
}
```
Doing `RUST_LOG=warn cargo test std_thread -- --nocapture` gives:
```
running 1 test
2023-11-27T17:38:49.695078Z ERROR inner1{tid=0}: r#mod::basic::tracing: incrementing from 0
2023-11-27T17:38:49.695650Z ERROR inner1{tid=1}: r#mod::basic::tracing: incrementing from 1
2023-11-27T17:38:49.696859Z ERROR inner1{tid=2}: r#mod::basic::tracing: incrementing from 2
2023-11-27T17:38:49.697743Z ERROR inner2{tid=0}: r#mod::basic::tracing: incrementing from 3
test basic::tracing::std_thread_joining_is_irrelevant ... ok
2023-11-27T17:38:49.697917Z ERROR inner2{tid=1}: r#mod::basic::tracing: incrementing from 4

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 
```
As we can see, the same as above.

### Shuttle `thread::spawn`
```rust
fn shuttle_joining_is_irrelevant() {
    let lock = Arc::new(Mutex::new(0));
    let threads: Vec<_> = (0..3)
        .map(|i| {
            let lock = lock.clone();
            thread::spawn(move || {
                let outer = warn_span!("outer", tid = i);
                let _outer = outer.enter();

                let lock_clone = lock.clone();
                let t = thread::spawn(move || {
                    let mut locked = lock_clone.lock().unwrap();
                    let inner = warn_span!("inner1", tid = i);
                    let _inner = inner.enter();
                    error!("incrementing from {}", *locked);
                    *locked += 1;
                });

                t.join().unwrap();

                let _t = thread::spawn(move || {
                    let mut locked = lock.lock().unwrap();
                    let inner = warn_span!("inner2", tid = i);
                    let _inner = inner.enter();
                    error!("incrementing from {}", *locked);
                    *locked += 1;
                });
            })
        })
        .collect();

    for thread in threads {
        thread.join().unwrap();
    }
}

#[test]
fn shuttle_test_joining_is_irrelevant() {
    let _ = tracing_subscriber::fmt::try_init();
    shuttle::check_random(shuttle_joining_is_irrelevant, 1);
}
```
Running `RUST_LOG=warn cargo test shuttle_test_join -- --nocapture` gives:
```
running 1 test
2023-11-27T17:40:09.863872Z ERROR inner1{tid=0}: r#mod::basic::tracing: incrementing from 0
2023-11-27T17:40:09.876532Z ERROR inner2{tid=0}: r#mod::basic::tracing: incrementing from 1
2023-11-27T17:40:09.878446Z ERROR inner1{tid=2}: r#mod::basic::tracing: incrementing from 2
2023-11-27T17:40:09.878521Z ERROR inner1{tid=1}: r#mod::basic::tracing: incrementing from 3
2023-11-27T17:40:09.878573Z ERROR inner2{tid=2}: r#mod::basic::tracing: incrementing from 4
2023-11-27T17:40:09.878639Z ERROR inner2{tid=1}: r#mod::basic::tracing: incrementing from 5
test basic::tracing::shuttle_test_joining_is_irrelevant ... ok
```
As we can see, the same as the two above.

---


sarsko:
Issue when running with `RUST_LOG=info`:
```
tried to clone Id(195906583790617176), but no span exists with that ID
This may be caused by consuming a parent span (`parent: span`) rather than borrowing it (`parent: &span`).
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: <tracing_subscriber::registry::sharded::Registry as tracing_core::subscriber::Subscriber>::clone_span
   3: <tracing_subscriber::fmt::Subscriber<N,E,F,W> as tracing_core::subscriber::Subscriber>::enter
   4: amzn_shuttle_fork::runtime::execution::Execution::step::{{closure}}::{{closure}}
   5: amzn_shuttle_fork::runtime::execution::Execution::step
   6: scoped_tls::ScopedKey<T>::set
   7: amzn_shuttle_fork::runtime::execution::Execution::run
   8: tracing::span::Span::in_scope
   9: scoped_tls::ScopedKey<T>::set
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

sarsko:
There is a question of whether the step and execution spans should be ERROR so that they are always included. Having that information is pretty much always useful.

src/runtime/execution.rs
```diff
@@ -144,7 +144,42 @@ impl Execution {
...
+                let result = panic::catch_unwind(panic::AssertUnwindSafe(|| continuation.borrow_mut().resume()));
+
+                // Leave the Task's span and store which span it exited in order to restore it the next time the Task is run
+                ExecutionState::with(|state| {
+                    tracing::dispatcher::get_default(|subscriber| {
+                        let current_span = tracing::Span::current();
+                        if let Some(span_id) = current_span.id().as_ref() {
+                            subscriber.exit(span_id);
+                        }
+                        state.current_mut().span = current_span;
```
jorajeev, line 174: This makes me wonder if the field should be renamed from `span` to `last_span` ?

jorajeev (commented):
Thanks for the CR - this will debugging a lot easier, esp when using Shuttle with proptesting.  And the `*_awaiting_is_irrelevant` examples are really helpful!

src/runtime/task/mod.rs
```diff
@@ -170,9 +188,10 @@ impl Task {
         id: TaskId,
         name: Option<String>,
         clock: VectorClock,
-        parent_span: tracing::Span,
+        parent_id: Option<tracing::span::Id>,
```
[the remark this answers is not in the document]
sarsko: Agreed on all of these.

tests/basic/tracing.rs
```diff
@@ -0,0 +1,82 @@
...
+            })
+        })
+        .collect();
+
+    for thread in threads {
+        thread.join().unwrap();
+    }
+}
+
+#[test]
```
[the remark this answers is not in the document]
sarsko: Yes. I'd like to add a Subscriber which checks them. Can ignore for now, then add that later (I'll open an issue).

src/runtime/execution.rs
```diff
@@ -287,11 +327,8 @@ impl ExecutionState {
     {
         Self::with(|state| {
             let schedule_len = state.current_schedule.len();
+            let parent_id = state.span.id();
```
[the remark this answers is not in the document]
sarsko: Why won't this play nicely with `tracing::instrument`? It creates a mapping between polling and dropping a future and entering its span. We can still use the span as a parent for another span, or enter it elsewhere.

src/runtime/execution.rs
```diff
@@ -240,6 +279,7 @@ impl ExecutionState {
             current_schedule: initial_schedule,
             #[cfg(debug_assertions)]
             has_cleaned_up: false,
+            span: tracing::Span::current(),
```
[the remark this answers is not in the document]
sarsko: Hmm, I agree on the rename. There is some pedantry which can be done on what is the "top level"; This span is created inside the "execution" span, which is created inside whichever span the `Scheduler` has created, which is created inside whichever span it is created under (ie to get a root for an entire run), and so on (ie there might be another span for the entire test, to get all tracing output for a full proptest run or similar). From the view of the `ExecutionState` it is the top-level, from the view of the Shuttle test it is not.

sarsko:
Nevermind, I did not look into the non ERROR/WARN-traces sufficiently. There is Span-stacking on `instrument`ed futures.

src/runtime/execution.rs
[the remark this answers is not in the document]
sarsko: This turned out to be correct. There is span stacking on `instrument`ed futures. I'll have a look.

sarsko:
Interestingly the issues with instrumented futures don't show up when running with Tracing 0.2.0 (the current pre-release version).

src/runtime/execution.rs
```diff
@@ -593,31 +642,20 @@ impl ExecutionState {
             }
         }
 
-        trace!(?runnable, next_task=?self.next_task);
+        self.top_level_span
```
jorajeev, line 645: Add a comment explaining why this is needed?

src/runtime/runner.rs
```diff
@@ -15,6 +15,37 @@ use std::thread;
 use std::time::Instant;
 use tracing::{span, Level};
 
+// A helper struct which on `drop` exits all current spans, then enters the span which was entered when it was constructed.
+// The reason this exist is to solve the "span-stacking" issue which occurs when there is a panic inside `run` which is
```
jorajeev, line 19: typo: "exist" -> "exists"

src/runtime/runner.rs
```diff
@@ -15,6 +15,37 @@ use std::thread;
...
+// then caught by `panic::catch_unwind()` (such as when Shuttle is run inside proptest).
+// In other words: it enables correct spans when doing proptest minimization.
+struct ResetSpanOnDrop {
+    span: tracing::Span,
+}
+
+impl ResetSpanOnDrop {
+    #[must_use]
```
jorajeev, line 27: Nit: just put the `[must_use]` on the struct definition?

src/runtime/runner.rs
```diff
@@ -66,7 +98,12 @@ impl<S: Scheduler + 'static> Runner<S> {
                 let execution = Execution::new(self.scheduler.clone(), schedule);
                 let f = Arc::clone(&f);
 
-                span!(Level::INFO, "execution", i).in_scope(|| execution.run(&self.config, move || f()));
+                // This is a slightly lazy way to ensure that everything outside of the "execution" span gets
+                // established correctly between executions. Fully `exit`ing and fully `enter`ing (explicitly
+                // `enter`/`exit` all `Span`s) would most likely obviate the need for this.
+                let _rsod = ResetSpanOnDrop::new();
```
jorajeev, line 104: Make the naming more consistent: outer one is `_span_drop_guard` and this one is `_rsod`

src/runtime/task/mod.rs
```diff
@@ -89,7 +90,16 @@ pub(crate) struct Task {
     name: Option<String>,
 
     local_storage: StorageMap,
-    pub(super) span: tracing::Span,
+
+    // The `Span` containing the `Task`s `id` and the current step count (if step count recording is enabled)
```
jorajeev, line 94: I don't understand what it means to say "the Span containing the Task's id" -- could you clarify?

src/runtime/task/mod.rs
```diff
@@ -89,7 +90,16 @@ pub(crate) struct Task {
...
+    pub(super) step_span: Span,
+    // The current `Span` "stack" of the `Task`. There are two things to note:
+    // 1: We have to own the `Span`s (versus storing `Id`s) for the `Span` to not get dropped while the task is switched out.
+    // 2: We have to store the stack of `Span`s in order to return to the correct `Span` once the `Entered<'_>` from an
+    //    `instrument`ed future is dropped.
+    // `span_stack` is (as the name suggests) a stack. We `pop` it empty when resuming a `Task`, and `push` + `exit`
```
jorajeev, line 100: Also add a comment here stating the order in which spans are stored in the `span_stack` (the outermost span is at the top of the stack)

src/runtime/task/mod.rs
```diff
@@ -89,7 +90,16 @@ pub(crate) struct Task {
...
+    // `tracing::Span::current()` until there is no entered `Span` when we switch out of the `Task`.
```
jorajeev, line 101: Add a newline before this comment

src/runtime/task/mod.rs
```diff
@@ -117,11 +128,11 @@ impl Task {
         let waker = make_waker(id);
         let continuation = Rc::new(RefCell::new(continuation));
 
-        let span = if name == Some("main-thread".to_string()) {
-            parent_span
-        } else {
-            tracing::info_span!(parent: parent_span.id(), "step", i = schedule_len, task = id.0)
-        };
+        let step_span = error_span!(parent: parent_span_id.clone(), "step", task = id.0, i = field::Empty);
+        // Note that this is slightly lazy — we are starting storing at the step_span, but could have gotten the
```
jorajeev, line 132: Hm, if we know that storing the full stack would be better, why are we cutting corners here?  Why not just do the right thing?

tests/basic/tracing.rs
```diff
@@ -0,0 +1,127 @@
...
+use shuttle::{check_random, thread};
+use std::time::Duration;
+use test_log::test;
+use tracing::{instrument::Instrument, warn, warn_span};
+
+// NOTE:
+// All of the tests testing tracing will either have to have weak assertions,
+// or be #[ignore]d. The reason for this is that they are not thread safe
+// (since everything tracing is managed globally), and there is no way to
+// ensure that the tests are run single-threaded.
```
jorajeev, line 14: There is this crate: https://docs.rs/serial_test/latest/serial_test/index.html

See if it's worth using here?

tests/basic/tracing.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+async fn spawn_instrumented_futures() {
+    let jhs = (0..2)
+        .map(|_| {
+            spawn(async {
+                let span_id = tracing::Span::current().id();
+                async {
+                    // NOTE: Just a way to get a `thread::switch` call.
+                    // Consider `pub`ing `thread::switch` ?
```
jorajeev, line 99: Why not just call `yield_now` ?

src/runtime/execution.rs
```diff
@@ -144,7 +144,46 @@ impl Execution {
         // Run a single step of the chosen task.
         let ret = match next_step {
             NextStep::Task(continuation) => {
-                panic::catch_unwind(panic::AssertUnwindSafe(|| continuation.borrow_mut().resume()))
+                // Enter the Task's span
+                // Note that if any issues arises with spans and tracing, then calling `exit` until `None`,
```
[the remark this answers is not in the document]
sarsko: Neither. I am assuming that the current scheme works, but, if it for some reason does not work, then the noted change should be tried.

src/runtime/execution.rs
```diff
@@ -144,7 +144,46 @@ impl Execution {
...
+                // storing the entirety of the `span_stack` when creating the `Task` and storing
+                // `top_level_span` as a stack should be tried.
+                ExecutionState::with(|state| {
+                    tracing::dispatcher::get_default(|subscriber| {
+                        if let Some(span_id) = tracing::Span::current().id().as_ref() {
+                            subscriber.exit(span_id);
+                        }
+
+                        while let Some(span) = state.current_mut().span_stack.pop() {
```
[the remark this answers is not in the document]
sarsko: Can add. There already is one where `span_stack` is defined (in `Task`), do we want a description here as well?

src/runtime/task/mod.rs
sarsko: ?
Sure, but I'm not sure how it can be described differently (in a way which is understandable). This is the span which contains the word "step" (thus the name `step_span`) and which contains the task = id.0 field and optionally the current step count (denoted i = _)

src/runtime/task/mod.rs
sarsko: Because I don't think it is better / I don't believe this is cutting corners. This comment and the other one of a similar nature are meant as the following: "I really believe the current scheme is correct. However, spans are both weird and tricky. If there ever are any issues, then start by doing the changes in the comments, in order to definitely rule them out.". In other words: I think the current way of doing it is "just right", and that the suggestions are "too much". I'd rather do the "just right" approach, both for overall code cleanness, and for the opportunity to be proven wrong (and thus learn more about Spans), but leave the comments in case I am wrong.

tests/basic/tracing.rs
sarsko: Will do, thanks!

tests/basic/tracing.rs
sarsko: Hmm true.

tests/basic/tracing.rs
sarsko: It seems to me that it won't solve the issue, unless all other tests are marked with `parallel`, which isn't worth it. This is because marking a test with `serial` only guarantees that it won't be ran at the same time as other `serial` or `parallel` tests, but it can still run at the same time as tests which are not marked.

jorajeev (approved):

PR #136, opened 2024-02-27
Implement RefUnwindSafe for Atomic

sarsko:
Atomics should be `RefUnwindSafe`, and currently they are not. This PR makes them `RefUnwindSafe`.

---

### DIFF
--- a/src/sync/atomic/mod.rs
+++ b/src/sync/atomic/mod.rs
@@ -63,6 +63,7 @@ use crate::runtime::execution::ExecutionState;
 use crate::runtime::task::clock::VectorClock;
 use crate::runtime::thread;
 use std::cell::RefCell;
+use std::panic::RefUnwindSafe;
 
 static PRINTED_ORDERING_WARNING: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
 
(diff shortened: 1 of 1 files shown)


bkragl (approved):
Two things that can be done separately.

First, it seems like `Mutex` and `RwLock` explicitly implement `UnwindSafe`, although that shouldn't be necessary.

Second, I think it would be nice to add tests that assert the auto traits that we expect our sync primitives to implement. Something like https://github.com/indexmap-rs/indexmap/blob/184fe4bfcba20e21242c924220170b98be157d45/src/map/core.rs#L636-L642.

sarsko:
> First, it seems like Mutex and RwLock explicitly implement UnwindSafe, although that shouldn't be necessary.

True, though doing it explicitly is a low-cost way to avoid an accidental footgun (if their internals are ever changed). It also makes their documentation match that of their `std` counterpart (through moving `UnwindSafe` up from the "Auto Trait Implementations" section.

> Second, I think it would be nice to add tests that assert the auto traits that we expect our sync primitives to implement. Something like [indexmap-rs/indexmap@184fe4b/src/map/core.rs#L636-L642](https://github.com/indexmap-rs/indexmap/blob/184fe4bfcba20e21242c924220170b98be157d45/src/map/core.rs#L636-L642).

Yes, agreed.

PR #137, opened 2024-03-02
Update tag modification test + make it not rely on Shuttle internals

sarsko:
As noted by @jorajeev, this test relies on the main thread being `TaskId(0)`. Updated the test to remove this reliance on Shuttle internals, and added an assertion that the spawned task is unaffected by the tag setting while at it.

---


sarsko:
Closed as tags are deprecated.

PR #138, opened 2024-03-04
Add support for typed task Labels.

jorajeev:
[PR98](https://github.com/awslabs/shuttle/pull/98) added support for associated a single untyped `Tag` with each task and thread.  As we've gained experience with Tags, we've increasingly felt a need to have a mechanism that is both better typed, and allows more than one tag to be associated with tasks. This commit introduces `Labels`, which are inspired by `Extensions` in [the http crate](https://docs.rs/http/latest/http/struct.Extensions.html). Users can attach any set of Labels to a task or thread, with the only caveat being that there can be at most one Label for a given type T.  This is not too onerous a restriction, since one can use the common [newtype idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) to easily work around this.

For tracing, we also provide a newtype `TaskName` that can be converted to and from a `String`. If the `TaskName` label is set for a task, tracing output will show the `TaskName` (in addition to the `TaskId`) to make logs easier to read.

Since the current functionality provided by `Tag` is superseded by `Labels`, we also mark `Tag` as deprecated.  They will be removed in a future release.

---

### DIFF
--- a/src/current.rs
+++ b/src/current.rs
@@ -4,10 +4,22 @@
 //! building tools that need to exploit Shuttle's total ordering of concurrent operations; for
 //! example, a tool that wants to check linearizability might want access to a global timestamp for
 //! events, which the [`context_switches`] function provides.
+//!
+//! This module also provides functions to manage the assocation of `labels` to threads and async tasks.
+//! Labels are typed values that can be associated with a task.  They are useful for debugging: for
+//! instance, the `TaskName` label can be set to assign names to tasks to make debug output easier to read.
+//! Labels can also be used to build customized schedulers: for instance, they can be used to assign
+//! numeric weights to tasks, which can be used to implement a priority-preemptive scheduler.
 
-use crate::runtime::execution::{ExecutionState, TASK_ID_TO_TAGS};
+#[allow(deprecated)]
+use crate::runtime::execution::TASK_ID_TO_TAGS;
+use crate::runtime::execution::{ExecutionState, LABELS};
 use crate::runtime::task::clock::VectorClock;
-pub use crate::runtime::task::{Tag, Taggable, TaskId};
+pub use crate::runtime::task::labels::Labels;
+pub use crate::runtime::task::{ChildLabelFn, TaskId, TaskName};
+#[allow(deprecated)]
+pub use crate::runtime::task::{Tag, Taggable};
+use std::fmt::Debug;
 use std::sync::Arc;
 
 /// The number of context switches that happened so far in the current Shuttle execution.
@@ -34,15 +46,41 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
     ExecutionState::with(|state| state.get_clock(task_id).clone())
 }
 
-/// Sets the `tag` field of the current task.
-/// Returns the `tag` which was there previously.
-pub fn set_tag_for_current_task(tag: Arc<dyn Tag>) -> Option<Arc<dyn Tag>> {
-    ExecutionState::set_tag_for_current_task(tag)
+/// Apply the given function to the Labels for the specified task
+pub fn with_labels_for_task<F, T>(task_id: TaskId, f: F) -> T
+where
+    F: FnOnce(&mut Labels) -> T,
+{
+    LABELS.with(|cell| {
+        let mut map = cell.borrow_mut();
+        let m = map.entry(task_id).or_default();
+        f(m)
+    })
 }
 
-/// Gets the `tag` field of the current task.
-pub fn get_tag_for_current_task() -> Option<Arc<dyn Tag>> {
-    ExecutionState::get_tag_for_current_task()
+/// Get a label of the given type for the specified task, if any
+pub fn get_label_for_task<T: Clone + Debug + 'static>(task_id: TaskId) -> Option<T> {
+    with_labels_for_task(task_id, |labels| labels.get().cloned())
+}
+
+/// Add the given label to the specified task, returning the old label for the type, if any
+pub fn set_label_for_task<T: Clone + Debug + 'static>(task_id: TaskId, value: T) -> Option<T> {
+    with_labels_for_task(task_id, |labels| labels.insert(value))
+}
+
+/// Remove a label of the given type for the specified task, returning the old label for the type, if any
+pub fn remove_label_for_task<T: Clone + Debug + 'static>(task_id: TaskId) -> Option<T> {
+    with_labels_for_task(task_id, |labels| labels.remove())
+}
+
+/// Get the debug name for a task
+pub fn get_name_for_task(task_id: TaskId) -> Option<TaskName> {
+    get_label_for_task::<TaskName>(task_id)
+}
+
+/// Set the debug name for a task, returning the old name, if any
+pub fn set_name_for_task(task_id: TaskId, task_name: impl Into<TaskName>) -> Option<TaskName> {
+    set_label_for_task::<TaskName>(task_id, task_name.into())
 }
 
 /// Gets the `TaskId` of the current task, or `None` if there is no current task.
@@ -50,7 +88,29 @@ pub fn get_current_task() -> Option<TaskId> {
     ExecutionState::with(|s| Some(s.try_current()?.id()))
 }
 
+/// Get the `TaskId` of the current task.  Panics if there is no current task.
+pub fn me() -> TaskId {
+    get_current_task().unwrap()
+}
+
+/// Sets the `tag` field of the current task.
+/// Returns the `tag` which was there previously.
+#[deprecated]
+#[allow(deprecated)]
+pub fn set_tag_for_current_task(tag: Arc<dyn Tag>) -> Option<Arc<dyn Tag>> {
+    ExecutionState::set_tag_for_current_task(tag)
+}
+
+/// Gets the `tag` field of the current task.
+#[deprecated]
+#[allow(deprecated)]
+pub fn get_tag_for_current_task() -> Option<Arc<dyn Tag>> {
+    ExecutionState::get_tag_for_current_task()
+}
+
 /// Gets the `tag` field of the specified task.
+#[deprecated]
+#[allow(deprecated)]
 pub fn get_tag_for_task(task_id: TaskId) -> Option<Arc<dyn Tag>> {
     TASK_ID_TO_TAGS.with(|cell| {
         let map = cell.borrow();
@@ -59,6 +119,8 @@ pub fn get_tag_for_task(task_id: TaskId) -> Option<Arc<dyn Tag>> {
 }
 
 /// Sets the `tag` field of the specified task.
+#[deprecated]
+#[allow(deprecated)]
 pub fn set_tag_for_task(task: TaskId, tag: Arc<dyn Tag>) -> Option<Arc<dyn Tag>> {
     ExecutionState::set_tag_for_task(task, tag)
 }
--- a/src/runtime/task/labels.rs
+++ b/src/runtime/task/labels.rs
@@ -1,0 +1,339 @@
+//! Labels are a way to attach an arbitrary set of values with a task, with the only
+//! constraint being that at most one value of a given type `T` can be attached at any time,
+//! using `get::<T>()` to retrieve, and `set::<T>(..)` to set the value.
+//! Labels behave almost identically to [Extensions](https://docs.rs/http/latest/http/struct.Extensions.html)
+//! in the `http` crate.
+
+/*
+** The code below is directly copied from https://github.com/hyperium/http/blob/master/src/extensions.rs
+** but renaming 'Extensions' to 'Labels'
+**
+** The key idea is to keep a HashMap (named `AnyMap`) that maps the `TypeId` for a type
+** to its associated value, so a `get::<T>()` is translated to `get(TypeId::of::<T>())`.
+*/
+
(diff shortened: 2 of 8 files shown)


src/current.rs
```diff
@@ -34,23 +39,74 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
...
+
+/// Add the given label to the specified task, returning the old label for the type, if any
+#[inline]
+pub fn set_label_for_task<T: Clone + std::fmt::Debug + Send + Sync + 'static>(task_id: TaskId, value: T) -> Option<T> {
+    with_labels_for_task(task_id, |labels| labels.insert(value))
+}
+
+/// Remove a label of the given type for the specified task, returning the old label for the type, if any
+#[inline]
+pub fn remove_label_for_task<T: Clone + std::fmt::Debug + Send + Sync + 'static>(task_id: TaskId) -> Option<T> {
```
sarsko, line 68: Nit: I think these should have the `std::fmt::` part removed.

src/runtime/task/labels.rs
```diff
@@ -0,0 +1,330 @@
+/*
+** Code directly copied from https://github.com/hyperium/http/blob/master/src/extensions.rs
+** but renaming 'Extensions' to 'Labels'
+*/
```
sarsko, line 4: Would be nice with a doc comment explaining what the code does.

tests/basic/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+        Id::from_u64(1)
+    }
+
+    fn record(&self, _span: &Id, _values: &Record<'_>) {}
+    fn record_follows_from(&self, _span: &Id, _follows: &Id) {}
+
+    fn event(&self, event: &Event<'_>) {
+        let metadata = event.metadata();
+        let target = metadata.target();
+        if target.contains("shuttle") && target.ends_with("::runtime::execution") {
```
sarsko, line 310: Why is this no longer `target() ==` and why is it split up (why is it not `target.ends_with("shuttle::runtime::execution")`?)

tests/basic/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+// The root waits for all tasks to complete, and collects the returned usize pairs. It then checks
+// - that the final AtomicUsize value is 4
+// - that each (parent, child) pair in the tree above is reported exactly once
+//
+// This test checks the following properties
+// - tasks inherit the Labels of their parent
+// - changing child Labels doesn't affect parent Labels
+// - child Labels are independent of each other
+async fn spawn_tasks(counter: Arc<AtomicUsize>) -> HashSet<(usize, usize)> {
+    set_label_for_task(me(), Ident(1));
```
sarsko, line 42: Not super important, but I find the `set_label_for_task(me()` slightly annoying (the annoying part is needing two imports/calls to do the most common operation for `Labels`). Not super fan of `set_label_for_current_task` either (though I think I prefer that over the `set me`. Maybe `set_label(label: Label)` ?

src/runtime/task/mod.rs
```diff
@@ -38,6 +41,73 @@ use waker::make_waker;
...
+/// shuttle::check_dfs(|| {
+///   set_label_for_task(me(), LabelFn(Arc::new(|labels| { labels.insert(TaskName::from("ChildTask")); })));
+///   shuttle::thread::spawn(|| {
+///     assert_eq!(get_name_for_task(me()).unwrap(), TaskName::from("ChildTask")); // child task already has the name
+///     // ... rest of child
+///   }).join().unwrap();
+/// }, None);
+/// ```
+#[derive(Clone)]
+pub struct LabelFn(pub Arc<dyn Fn(&mut Labels) + Send + Sync + 'static>);
```
jamesbornholt, line 103: Instead of this, maybe we could add a new method to `std::thread::Builder` that sets the labels for the new task atomically with its creation? Oh, but there's no builders for async tasks. That's too bad.

src/runtime/task/mod.rs
```diff
@@ -38,6 +41,73 @@ use waker::make_waker;
...
+    }
+}
+
+/// By default, when a task or thread T is spawned, it inherits all labels from its parent.
+/// It's often useful to modify or add new Labels to T.  One approach is to put label changes
+/// at the beginning of the closure that is passed to `spawn`, but this approach has the drawback
+/// that the changes are applied only when T is first selected for execution, and the closure
+/// is invoked.  To overcome this drawback, we introduce the `LabelFn` label, which allows
+/// users to define a function that modifies the labels inherited from the parent when T is
+/// initially created.
```
jamesbornholt, line 85: Maybe a little more detailed: `LabelFn` should be used as a label on the parent task. When a task with a `LabelFn` label spawns a new task, the wrapped function will be applied to the child's label set at spawn time.

Also, `LabelFn` is a bit too generic as a name; maybe `ChildLabelFn` or something.

src/runtime/task/mod.rs
```diff
@@ -464,14 +539,13 @@ pub(crate) struct ParkState {
 pub struct TaskId(pub(super) usize);
 
 impl Debug for TaskId {
+    // If the `TaskName` label is set, use that when generating the Debug string
```
jamesbornholt, line 542: Might be nice to make `std::thread::Builder` set the `TaskName` label on a new thread if it has a `.name()` set?

src/runtime/task/labels.rs
jorajeev: Added.

src/runtime/task/mod.rs
jorajeev: Yeah, exactly.  It's also easier to add this to existing code that typically just calls `thread::spawn(move || { .. })` or `tokio::spawn(async move { .. })`.

src/runtime/task/mod.rs
jorajeev: Renamed to `ChildLabelFn` and updated the comment.

src/runtime/task/mod.rs
jorajeev: Think I'm already doing that: `Builder::spawn` calls `thread::spawn_named` which calls `ExecutionState::spawn_thread` which calls `ExecutionState::set_labels_for_new_task` which sets the name if one was provided.

tests/basic/labels.rs
jorajeev: It provides a bit more flexibility, and doesn't make the check much weaker.

src/current.rs
jorajeev: Agreed, and fixed.

tests/basic/labels.rs
jorajeev: Yeah, I didn't want to duplicate the function set with `{get,set}_label_for_task` and `{get,set}_label_for_current_task`.  One alternative I considered was to allow the argument to be `Option<TaskId>`, and making `None` always mean the current task.  I wasn't sure how readable that would be, and it's just as many characters as `me()`.

src/runtime/task/mod.rs
jorajeev: I added a check to `tests::basic::thread::thread_builder_name` to check that the name provided to the builder is assigned when you call `get_name_for_task()`.

tests/basic/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+    #[allow(clippy::type_complexity)]
+    let handles: Vec<future::JoinHandle<(usize, usize, Vec<future::JoinHandle<(usize, usize)>>)>> = (0..2)
```
sarsko, line 44: This resolves with `Vec<_>`

tests/basic/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+        },
+        10_000,
+    );
+}
+
+// Same test as above, except with threads
+fn spawn_threads(counter: Arc<AtomicUsize>) -> HashSet<(usize, usize)> {
+    set_label_for_task(me(), Ident(1));
+    #[allow(clippy::type_complexity)]
+    let handles: Vec<thread::JoinHandle<(usize, usize, Vec<thread::JoinHandle<(usize, usize)>>)>> = (0..2)
```
sarsko, line 120: This resolves with `Vec<_>`

tests/basic/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+        .map(|i| {
+            let counter2 = counter.clone();
+            thread::spawn(move || {
+                // Set this task's Ident to be (2b + i) where (b) is the Ident of the parent
+                let Ident(parent) = get_label_for_task(me()).unwrap();
+                set_label_for_task(me(), Parent(parent));
+                set_label_for_task(me(), Ident(2 * parent + i));
+
+                let handles: Vec<thread::JoinHandle<(usize, usize)>> = (0..2)
```
sarsko, line 129: This resolves with `Vec<_>`

sarsko:
(discussed with Rajeev, posting here to not forget it / have an entry where it may be discussed)

One thing which is lost from `Label`s vs `Tag`s is the ability to gain more information than a simple string in the tracing output from Shuttle. There is a question of whether there should be the possibility to have more information logged, for instance by exposing a way to provide a closure with a type signature ala `&[Label] -> Debug`.

tests/basic/labels.rs
sarsko: https://github.com/awslabs/shuttle/commit/496b60e92566a8ae5ba265ca94a1b0e5d2b7f4a2

sarsko:
`Send + Sync` bound can be removed: https://github.com/awslabs/shuttle/commit/982213842338aedf2ca93901eb01fb5e07c908ac

tests/basic/labels.rs
jorajeev: Nice catch; thanks

jorajeev:
Summary of changes since the first reviewed version:
* renamed `LabelFn` -> `ChildLabelFn`, and changed the function signature to `Fn(TaskId, &mut Labels)`, because there are situations where the `TaskId` is useful
* removed `Send + Sync` requirement for types usable as Labels

src/current.rs
```diff
@@ -34,23 +40,74 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
...
+{
+    LABELS.with(|cell| {
+        let mut map = cell.borrow_mut();
+        let m = map.entry(task_id).or_default();
+        f(m)
+    })
+}
+
+/// Get a label of the given type for the specified task, if any
+#[inline]
```
jamesbornholt, line 56: Do these inline annotations end up having any effect?

src/runtime/task/labels.rs
jamesbornholt: nit: `//!`

src/current.rs
```diff
@@ -34,23 +40,74 @@ pub fn clock_for(task_id: TaskId) -> VectorClock {
...
+
+/// Get the `TaskId` of the current task.  Panics if there is no current task.
+#[inline]
+pub fn me() -> TaskId {
+    get_current_task().unwrap()
+}
+
 /// Sets the `tag` field of the current task.
 /// Returns the `tag` which was there previously.
+#[allow(deprecated)]
```
jamesbornholt, line 98: Don't we also want these to be marked as `#[deprecated]`, since they're the public API we're about to remove?

src/runtime/task/mod.rs
```diff
@@ -38,6 +41,74 @@ use waker::make_waker;
...
+    }
+}
+
+impl<'a> From<&'a TaskName> for &'a String {
+    fn from(task_name: &'a TaskName) -> Self {
+        &task_name.0
+    }
+}
+
+/// By default, when a task or thread T is spawned, it inherits all labels from its parent.
```
jamesbornholt, line 79: nit: The docs style is to write a one-sentence summary (which shows up on index pages) and then the rest of the text as a second+ paragraph.

src/current.rs
```diff
@@ -5,9 +5,15 @@
 //! example, a tool that wants to check linearizability might want access to a global timestamp for
 //! events, which the [`context_switches`] function provides.
```
jamesbornholt, line 6: Could we add a paragraph to this doc comment with a quick overview of labels and what they're useful for?

src/runtime/task/mod.rs
```diff
@@ -38,6 +41,74 @@ use waker::make_waker;
 
 pub(crate) const DEFAULT_INLINE_TASKS: usize = 16;
 
+/// To make debugging easier, if a task is assigned a `TaskName(s)` Label,
+/// Shuttle will display the String `s` in addition to the `TaskId` in debug output.
```
jamesbornholt, line 45: nit: same thing about docs style being one-line summary of "what the thing is", then more paragraphs

jamesbornholt (commented):
just docs nits

src/current.rs
jorajeev: Yeah, I wondered about that, since the compiler can figure this out on its own anyway.  Removing.

src/current.rs
jorajeev: Agreed; done.

src/current.rs
jorajeev: Added.

src/runtime/task/mod.rs
jorajeev: Done

jamesbornholt (approved):

PR #139, opened 2024-03-07
Fix dropping TLS for detached futures

jamesbornholt:
When a detached future is being cleaned up after the end of an
execution, it won't be able to access the current task's storage any
more. We need to bail out early from the cleanup to avoid a panic.

I also fixed two small issues:
1. We accidentally took a dependency on tracing 0.1.36 by using a value type in a `Span::record` call, despite only specifying an earlier version.
2. The docs weren't building on older smallvec versions because of some const weirdness.

---

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -145,6 +145,13 @@ where
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.future.as_mut().poll(cx) {
             Poll::Ready(result) => {
+                // If we've finished execution already (this task was detached), don't clean up. We
+                // can't access the state any more to destroy thread locals, and don't want to run
+                // any more wakers (which will be no-ops anyway).
+                if ExecutionState::try_with(|state| state.is_finished()).unwrap_or(true) {
+                    return Poll::Ready(());
+                }
+
                 // Run thread-local destructors before publishing the result.
                 // See `pop_local` for details on why this loop looks this slightly funky way.
                 // TODO: thread locals and futures don't mix well right now. each task gets its own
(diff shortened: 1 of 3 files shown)


src/future/mod.rs
```diff
@@ -145,6 +145,13 @@ where
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.future.as_mut().poll(cx) {
             Poll::Ready(result) => {
+                // If we've finished execution already (this task was detached), don't clean up. We
+                // can't access the state any more to destroy thread locals, and don't want to run
+                // any more wakers (which will be no-ops anyway).
+                if ExecutionState::try_with(|state| state.is_finished()).unwrap_or(true) {
+                    return Poll::Ready(());
+                }
```
bkragl, line 153: Why are we still polling the future if execution already finished?

Was the task's local storage already dropped in `ExecutionState::cleanup` when we drained tasks?

jamesbornholt: We're not re-polling the future after execution finishes. But at `ExecutionState::cleanup` time, we drop the continuation, which involves _resuming_ the task so that it can unwind its stack. That's how we end up in this code. We do still clean up the task's local storage in`ExecutionState::cleanup`, so this change isn't actually leaking anything, just changing where the cleanup happens.

tests/future/basic.rs
```diff
@@ -385,3 +386,28 @@ fn is_finished_on_join_handle() {
         None,
     );
 }
+
+struct ShuttleSpawn;
+
+impl Spawn for ShuttleSpawn {
```
jorajeev, line 392: Nice - didn't know about the `Spawn` trait.

PR #147, opened 2024-05-14
#[allow(clippy::new_without_default)] for Once

bkragl:
fixes clippy error https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default

---


sarsko (commented):
Not the correct fix; [`Once` is not `Default`](https://doc.rust-lang.org/std/sync/struct.Once.html). Correct fix would be to `#[allow(clippy::new_without_default)]`.

jorajeev (approved):

PR #150, opened 2024-06-26
Add support for 'abort'ing shuttle::future::JoinHandle's

sarsko:
Adds support for `JoinHandle::abort`.

---


src/runtime/task/mod.rs
```diff
@@ -343,6 +343,12 @@ impl Task {
         self.detached = true;
     }
 
+    pub(crate) fn abort(&mut self) {
+        self.finish();
+        let mut continuation = self.continuation.borrow_mut();
+        continuation.wipe();
```
jamesbornholt, line 349: Maybe worth a comment about why `finish()` isn't good enough — we want to run the task's destructors now.

tests/future/basic.rs
```diff
@@ -221,6 +221,78 @@ fn async_counter() {
...
+            let jh = future::spawn(acquire_and_loop(mutex.clone()));
+
+            std::mem::forget(jh);
+
+            let _g = mutex.lock();
+        },
+        1000,
+    );
+}
+
```
jamesbornholt, line 295: Something like this doesn't work, because it's possible for the abort to happen while `t2` is blocked acquiring the lock:
```rust
#[test]
fn abort_while_blocked() {
    check_dfs(
        || {
            let lock = Arc::new(Mutex::new(0i32));
            let lock2 = lock.clone();
            let t1 = thread::spawn(move || {
                let _l = lock2.lock().unwrap();
            });
            let t2 = future::spawn(async move {
                let _l = lock.lock().unwrap();
            });
            t2.abort();
            t1.join().unwrap();
        },
        None,
    );
}
```
The way it fails right now is a bit funny — when `t1` drops the lock, it [asserts that all the waiters are still blocked](https://github.com/awslabs/shuttle/blob/9076a444c7f611eeb94fa0b562263bf29a178fbd/src/sync/mutex.rs#L277-L278), but when the future aborts it's no longer blocked.

Not sure what the right thing to do here is. Maybe we need to check lazily whether a task has been aborted, only once [a poll of the `Wrapper` task returns Pending](https://github.com/awslabs/shuttle/blob/9076a444c7f611eeb94fa0b562263bf29a178fbd/src/future/mod.rs#L173)? But that might be more trouble than it's worth for what you want to do.

src/future/mod.rs
```diff
@@ -54,12 +54,22 @@ impl<T> Default for JoinHandleInner<T> {
...
     /// Abort the task associated with the handle.
     pub fn abort(&self) {
         ExecutionState::try_with(|state| {
             if !state.is_finished() {
                 let task = state.get_mut(self.task_id);
-                task.detach();
+                task.abort();
             }
         });
     }
```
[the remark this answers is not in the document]
sarsko: Ah yes indeed, I'll have a look.

src/runtime/task/mod.rs
sarsko: Yes

tests/future/basic.rs
sarsko: Hmm, what about removing the `assert!` ([which is repeated in `unblock` anyways](https://github.com/awslabs/shuttle/blob/9076a444c7f611eeb94fa0b562263bf29a178fbd/src/runtime/task/mod.rs#L363)), and then updating `unblock` to allow state finished as well ? I need to do a pass and check that we don't ever do "silent passing" of synchro primitives if so. If we do "silent passing" then the "passer" has to check that the task it passes to isn't finished.

---

Actually, wait, is our implementation of `MutexGuard::drop` correct? Aren't `std::sync::Mutex`es fair?

jamesbornholt: Their fairness is platform-dependent IIRC.

Removing the assert might be OK, though we should add an explicit aborted flag as a sanity check that the only way a task gets unblocked is if it was aborted. But tokio's JoinHandle semantics are that the task cancels at the next `await` point. I worry about aborting a task mid-synchronization operation because it leaves data structures in an intermediate state that isn't reachable in real code. Like this:
```rust
let t1 = future::spawn(async move {
    let l1 = lock1.lock().unwrap();
    *l1 += 1;
    let l2 = lock2.lock().unwrap();
    *l2 += 1;
});
t1.abort();
assert_eq!(*l1.lock().unwrap(), *l2.lock().unwrap());
```
In real Rust this assertion can't fail because the task body is atomic from the perspective of `abort`, but in our implementation it can.

sarsko: Hmm yeah, the immediate `drop` of the `Continuation` doesn't seem to be the correct solution. `drop`ping immediately is correct if it has never been scheduled, else we need to do it at the next `await` point. I guess lazily checking in `Wrapper` is the correct solution.

PR #151, opened 2024-07-10
Batch semaphore, causality tracking

Aurel300:
This adds batch semaphores into Shuttle (not originally implemented by me) as the core primitive to be used for other synchronisation primitives. Vector clocks were added to the batch semaphore to track causality across `acquire`, `release`, and `try_acquire` calls. 

Implementing other primitives using the batch semaphore will thus also cause causality to be tracked. The eventual goal is for primitives to only rely on the API of batch semaphore, without calls to e.g. `ExecutionState`. As a result, batch semaphore and the runtime internals can be pulled out of Shuttle into a separate `shuttle-core` crate, while both the stdlib and tokio primitives can be placed into new sibling crates, `shuttle-stdlib` and `shuttle-tokio`, respectively.

An open question is how to proceed with the crate restructuring. One option is for the crate called `shuttle` now to "become" `shuttle-stdlib`, as it will contain the same primitives, but this would require various re-exports. We could also publish a wrapper crate called `shuttle` (intended to be deprecated?) which will re-export items from both `shuttle-core` and `shuttle-stdlib`.

Planned for this PR still:

- [x] Test cases.
  - [x] Test case to show where causality tracking is *not* precise.
- [x] Option for the semaphore to not be strictly fair, by waking all feasible waiters at once, then letting the scheduler resolve the race.

### DIFF
--- a/src/future/batch_semaphore.rs
+++ b/src/future/batch_semaphore.rs
@@ -1,0 +1,361 @@
+//! A counting semaphore supporting both async and sync operations.
+use crate::runtime::execution::ExecutionState;
+use crate::runtime::task::TaskId;
+use std::cell::RefCell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::sync::Mutex;
+use std::task::{Context, Poll, Waker};
+use tracing::trace;
+
+#[derive(Debug)]
+struct Waiter {
+    task_id: TaskId,
+    num_permits: usize,
+    is_queued: AtomicBool,
+    has_permits: AtomicBool,
+    waker: Mutex<Option<Waker>>,
+}
+
+impl Waiter {
+    fn new(num_permits: usize) -> Self {
+        Self {
+            task_id: ExecutionState::me(),
+            num_permits,
+            is_queued: AtomicBool::new(false),
+            has_permits: AtomicBool::new(false),
+            waker: Mutex::new(None),
+        }
+    }
+}
+
+/// A counting semaphore which permits waiting on multiple permits at once,
+/// and supports both asychronous and synchronous blocking operations.
+///
+/// The semaphore is strictly fair, so earlier requesters always get priority
+/// over later ones.
+///
+/// TODO: Provide an option to support weaker models for fairness.
+#[derive(Debug)]
+struct BatchSemaphoreState {
+    // Key invariants:
+    //
+    // (1) if `waiters` is nonempty and the head waiter is `H`,
+    // then `H.num_permits > permits_available`.  (In other words, we are
+    // never in a state where there are enough permits available for the
+    // first waiter.  This invariant is ensured by the `drop` handler below.)
+    //
+    // (2) W is in waiters iff W.is_queued
+    //
+    // (3) W.is_queued ==> !W.has_permits
+    // Note: the converse is not true.  We can have !W.has_permits && !W.is_queued
+    // when the Acquire is created but not yet polled.
+    //
+    // (4) closed ==> waiters.is_empty()
+    waiters: VecDeque<Arc<Waiter>>,
+    permits_available: usize,
+    closed: bool,
+}
+
+impl BatchSemaphoreState {
+    fn acquire_permits(&mut self, num_permits: usize) -> Result<(), TryAcquireError> {
+        if self.closed {
+            Err(TryAcquireError::Closed)
+        } else if self.waiters.is_empty() && num_permits <= self.permits_available {
+            // No one is waiting and there are enough permits available
+            self.permits_available -= num_permits;
+            Ok(())
+        } else {
+            Err(TryAcquireError::NoPermits)
+        }
+    }
+}
+
+/// Counting semaphore
+#[derive(Debug)]
+pub struct BatchSemaphore {
+    state: RefCell<BatchSemaphoreState>,
+}
+
+/// Error returned from the [`BatchSemaphore::try_acquire`] function.
+#[derive(Debug, PartialEq, Eq)]
+pub enum TryAcquireError {
+    /// The semaphore has been closed and cannot issue new permits.
+    Closed,
+
+    /// The semaphore has no available permits.
+    NoPermits,
+}
+
+/// Error returned from the [`BatchSemaphore::acquire`] function.
+///
+/// An `acquire*` operation can only fail if the semaphore has been
+/// closed.
+#[derive(Debug)]
+pub struct AcquireError(());
+
+impl AcquireError {
+    fn closed() -> AcquireError {
+        AcquireError(())
+    }
+}
+
+impl fmt::Display for AcquireError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(fmt, "semaphore closed")
+    }
+}
+
+impl std::error::Error for AcquireError {}
+
+impl BatchSemaphore {
+    /// Creates a new semaphore with the initial number of permits.
+    pub fn new(num_permits: usize) -> Self {
+        let state = RefCell::new(BatchSemaphoreState {
+            waiters: VecDeque::new(),
+            permits_available: num_permits,
+            closed: false,
+        });
+        Self { state }
+    }
+
+    /// Creates a new semaphore with the initial number of permits.
+    pub const fn const_new(num_permits: usize) -> Self {
+        let state = RefCell::new(BatchSemaphoreState {
+            waiters: VecDeque::new(),
+            permits_available: num_permits,
+            closed: false,
+        });
+        Self { state }
+    }
+
+    /// Returns the current number of available permits.
+    pub fn available_permits(&self) -> usize {
+        let state = self.state.borrow();
+        state.permits_available
+    }
+
+    /// Closes the semaphore. This prevents the semaphore from issuing new
+    /// permits and notifies all pending waiters.
+    pub fn close(&self) {
+        let mut state = self.state.borrow_mut();
+        state.closed = true;
+        // Wake up all the waiters.  Since we've marked the state as closed, they
+        // will all return `AcquireError::closed` from their acquire calls.
+        let ptr = &*state as *const BatchSemaphoreState;
+        for waiter in state.waiters.drain(..) {
+            trace!(
+                "semaphore {:p} removing and waking up waiter {:?} on close",
+                ptr,
+                waiter,
+            );
+            assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+            assert!(!waiter.has_permits.load(Ordering::SeqCst)); // sanity check
+            ExecutionState::with(|exec_state| {
+                if !exec_state.in_cleanup() {
+                    exec_state.get_mut(waiter.task_id).unblock();
+                }
+            });
+            let mut maybe_waker = waiter.waker.lock().unwrap();
+            if let Some(waker) = maybe_waker.take() {
+                waker.wake();
+            }
+        }
+    }
+
+    /// Returns true iff the semaphore is closed.
+    pub fn is_closed(&self) -> bool {
+        let state = self.state.borrow();
+        state.closed
+    }
+
+    /// Try to acquire the specified number of permits from the Semaphore.
+    /// If the permits are available, returns Ok(())
+    /// If the semaphore is closed, returns `Err(TryAcquireError::Closed)`
+    /// If there aren't enough permits, returns `Err(TryAcquireError::NoPermits)`
+    pub fn try_acquire(&self, num_permits: usize) -> Result<(), TryAcquireError> {
+        let mut state = self.state.borrow_mut();
+        state.acquire_permits(num_permits)
+    }
+
+    fn enqueue_waiter(&self, waiter: &Arc<Waiter>) {
+        let mut state = self.state.borrow_mut();
+
+        trace!("enqueuing waiter {:?} for semaphore {:p}", waiter, &self.state);
+        state.waiters.push_back(waiter.clone());
+
+        assert!(!waiter.has_permits.load(Ordering::SeqCst));
+        assert!(!waiter.is_queued.swap(true, Ordering::SeqCst));
+    }
+
+    fn remove_waiter(&self, waiter: &Arc<Waiter>) {
+        let mut state = self.state.borrow_mut();
+
+        trace!(waiters = ?state.waiters, "removing waiter {:?} from semaphore {:p}", waiter, &self.state);
+
+        // sanity checks
+        assert!(!state.closed);
+        assert!(!waiter.has_permits.load(Ordering::SeqCst));
+
+        let index = state
+            .waiters
+            .iter()
+            .position(|x| Arc::ptr_eq(x, waiter))
+            .expect("did not find waiter");
+
+        state.waiters.remove(index).unwrap();
+        assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+    }
+
+    /// Acquire the specified number of permits (async API)
+    pub fn acquire(&self, num_permits: usize) -> Acquire<'_> {
+        Acquire::new(self, num_permits)
+    }
+
+    /// Acquire the specified number of permits (blocking API)
+    pub fn acquire_blocking(&self, num_permits: usize) -> Result<(), AcquireError> {
+        crate::future::block_on(self.acquire(num_permits))
+    }
+
+    /// Release `num_permits` back to the Semaphore
+    pub fn release(&self, num_permits: usize) {
+        if num_permits == 0 {
+            return;
+        }
+
+        if ExecutionState::should_stop() {
+            return;
+        }
+
+        let mut state = self.state.borrow_mut();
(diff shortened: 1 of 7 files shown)


sarsko:
Sweet.

> [...] while both the stdlib and tokio primitives can be placed into new sibling crates, shuttle-stdlib and shuttle-tokio, respectively.

> An open question is how to proceed with the crate restructuring. One option is for the crate called shuttle now to "become" shuttle-stdlib, as it will contain the same primitives, but this would require various re-exports. We could also publish a wrapper crate called shuttle (intended to be deprecated?) which will re-export items from both shuttle-core and shuttle-stdlib.

There's more.

Shuttle currently:
1. "Core": all the functionality to run the program under test, ie starting up, pausing and resuming tasks, shutting down, etc.
2. The schedulers (`scheduler`).
3. std-lib mirrors (`sync`,`thread.rs`, `hint.rs`, `thread_local!` in `lib.rs`).
4. `future`, ie a mix of `futures::executor` and `tokio::task::JoinHandle`.
5. `rand` mirror.
6. `lazy_static` mirror.
7. The "external" interface: `lib.rs` (minus `thread_local!` and `lazy_static!`) and `current.rs`

My opinion is that all of these should be its own crate.

The question then becomes whether something which is called Shuttle should stick around. I think the thing called Shuttle will be the crate corresponding to bullet number 7. There is a question of whether it should also provide reexports of some of the other crates. I have not thought about that enough to have a definitive opinion, but I am generally of the opinion that the correct approach is to use the wrapper crates over what is provided by Shuttle directly. I am also of the opinion that `future` should be deprecated (ie it should not be provided by Shuttle, but may be provided by some crate).

src/future/batch_semaphore.rs
```diff
@@ -57,17 +226,26 @@ struct BatchSemaphoreState {
     //
     // (4) closed ==> waiters.is_empty()
     waiters: VecDeque<Arc<Waiter>>,
-    permits_available: usize,
+    permits_available: PermitsAvailable,
+    // TODO: should there be a clock for the close event?
```
jorajeev, line 230: Yeah, think we need to add a causal dependency from the close event to subsequent `Closed` error returns.  But we can do that in a followup.

src/future/batch_semaphore.rs
```diff
@@ -179,7 +358,18 @@ impl BatchSemaphore {
     /// If there aren't enough permits, returns `Err(TryAcquireError::NoPermits)`
     pub fn try_acquire(&self, num_permits: usize) -> Result<(), TryAcquireError> {
         let mut state = self.state.borrow_mut();
-        state.acquire_permits(num_permits)
+        state.acquire_permits(num_permits).map_err(|err| {
+            // Conservatively, the requester causally depends on the
+            // last successful acquire.
+            // TODO: This is not precise, but `try_acquire` causal
```
jorajeev, line 364: Think you should add the commentary on what are the options, and why this choice is not as precise as the alternatives (of tracking clocks for all current permit holders).

src/future/batch_semaphore.rs
```diff
@@ -28,11 +30,178 @@ impl Waiter {
...
             waker: Mutex::new(None),
         }
     }
 }
 
+#[derive(Debug)]
+enum PermitsAvailable {
+    /// Arbitrarily many permits are available to be acquired. Acquiring one
+    /// is causally dependent on an event with the given clock.
+    Infinity(VectorClock),
```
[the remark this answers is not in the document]
Aurel300: I think "causally dependent on an event with the given clock" explains that?

src/future/batch_semaphore.rs
```diff
@@ -28,11 +30,178 @@ impl Waiter {
...
+                    // resulting clock should be the join of all the batches
+                    // used (fully or partially), since the acquiry causally
+                    // depends on the releases that created those batches.
+                    let mut clock = VectorClock::new();
+                    while let Some((batch_size, batch_clock)) = permit_clocks.front_mut() {
+                        clock.update(batch_clock);
+
+                        // An infinite number of permits was released at some
+                        // point, switch variant to reduce future operations.
+                        if *batch_size == usize::MAX {
```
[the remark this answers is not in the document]
Aurel300: We cannot do all of this on `release` because we don't want to completely forget the clocks of the permits that are already in the `permit_clocks`, which (in global time) are earlier events than the release of the "infinity" batch.

src/future/batch_semaphore.rs
Aurel300: Yes. I tried to implement this, but ran into issues with getting a clock in the `close` method, because it may happen as part of the cleanup process/drop handlers. We talked with @sarsko a bit about re-organising how drops happen, which would unblock this (adding a clock for closing).

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,614 @@
...
+                    let semaphore = semaphore.clone();
+                    let order1 = order1.clone();
+                    let order2 = order2.clone();
+                    thread::spawn(move || {
+                        // once the ID is pushed to the vector here and
+                        // observed in the busy loop below, the thread is
+                        // assumed to be blocked, because there is no yield
+                        // point between the push and the acquire
+                        order1.lock().unwrap().push(tid); // stdlib mutex
+                        let val = [2, 1, 1][tid];
```
jorajeev, line 61: nice

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,614 @@
...
+            (vec![2, 0, 1], vec![(2, 1), (1, 1)]),
+        ])
+    );
+}
+
+#[test]
+fn batch_semaphore_clock_1() {
+    check_dfs(
+        || {
+            let s = Arc::new(BatchSemaphore::new(0, Fairness::StrictlyFair));
```
jorajeev, line 139: This test would work for both forms of fairness, right? Maybe run it twice?

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,614 @@
...
+        },
+        None,
+    );
+}
+
+#[test]
+fn batch_semaphore_clock_2() {
+    check_dfs(
+        || {
+            let s = Arc::new(BatchSemaphore::new(0, Fairness::StrictlyFair));
```
jorajeev, line 162: Also works for both types of fairness

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,614 @@
...
+        },
+        None,
+    );
+}
+
+#[test]
+fn batch_semaphore_clock_3() {
+    check_dfs(
+        || {
+            let s = Arc::new(BatchSemaphore::new(0, Fairness::StrictlyFair));
```
jorajeev, line 189: Works for both types of fairness

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,606 @@
...
+                            .permits_available
+                            .acquire(waiter.num_permits, waiter.clock.clone())
+                            .unwrap();
+
+                        // Update waiter state as it is no longer in the queue
+                        assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+                        assert!(!waiter.has_permits.swap(true, Ordering::SeqCst));
+                        ExecutionState::with(|s| {
+                            let task = s.get_mut(waiter.task_id);
+                            if !task.finished() {
```
sarsko, line 437: I think this is wrong — we are giving the permits (line 427) to a task which is finished, and which will thus never release the permits. If the task is finished then we can just remove it and give the permits to the next eligible waiter. Also: how does a task become finished while queued for a semaphore?

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,606 @@
...
+            trace!("Acquire::poll for waiter {:?} already queued", self.waiter);
+
+            // Sanity check: there should be a waker, because the waiter is in
+            // the queue. Also true for unfair semaphores, which wake by ref.
+            assert!(self.waiter.waker.lock().unwrap().is_some());
+
+            // If the semaphore is fair, simply return. This corresponds to the
+            // future being polled after already having enqueued. No progress
+            // is made here, the semaphore will notify this waiter once ready.
+            if matches!(self.semaphore.state.borrow().fairness, Fairness::StrictlyFair) {
```
sarsko, line 540: Hmm. IMO this hides control flow and creates a footgun in the case we ever add another `Fairness`. I would much prefer a regular `match`.

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,606 @@
...
+                return Poll::Pending;
+            }
+
+            // Otherwise, try to acquire permits again. The result depends on
+            // which waiter(s) won the race after the `release` call.
+            let mut state = self.semaphore.state.borrow_mut();
```
sarsko, line 546: The code below is, apart from the `self.semaphore.remove_waiter(&self.waiter);` in the `Ok(())` branch, identical with the `else` branch below. Is it possible to condense the two pieces of code into one?

src/future/batch_semaphore.rs
Aurel300: (This is not a change introduced by this PR, but) I agree this seems strange. Changing this to a defensive `assert!(!task.finish())` does not break Shuttle tests at least.

src/future/batch_semaphore.rs
Aurel300: Gotcha, see if you find the current version more readable.

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,604 @@
...
+            // so the waiter will try to acquire some permits. If successful,
+            // the waiter need not be enqueued, and the future is resolved.
+            // Otherwise, the waiter is added to the queue.
+            //
+            // 3. is slightly different: the future was polled, even though the
+            // waiter was already in the queue. This can happen either because
+            // the semaphore just received some permits and woke the waiter up,
+            // or because the future itself was polled manually. Either way,
+            // the semaphore is queried.
+            let try_to_acquire = match (self.semaphore.state.borrow().fairness, is_queued) {
```
sarsko, line 549: ? Why dow we bind to `try_to_acquire` ?

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,604 @@
...
+        assert!(num_permits > 0);
+
+        // Permits here can be acquired in one of two scenarios:
+        // - The waiter queue is empty; nobody else is waiting for permits,
+        //   so if there are enough available, immediately succeed.
+        // - The semaphore is operating in an unfair mode; the current thread
+        //   is either requesting permits for the first time, or it was woken
+        //   and selected by the scheduler. In either case, the thread may
+        //   succeed, as long as there are enough permits.
+        let can_acquire = self.waiters.is_empty() || matches!(self.fairness, Fairness::Unfair);
```
sarsko, line 203: I find this style of binding when there's no need odd. Is it only to have something to attach the comment to or am I missing something?

sarsko: Why isn't the `if self.closed` branch right under the `assert!`?

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,604 @@
...
+        trace!(task = ?me, avail = ?state.permits_available, waiters = ?state.waiters, "released {} permits for semaphore {:p}", num_permits, &self.state);
+
+        match state.fairness {
+            Fairness::StrictlyFair => {
+                // in a strictly fair mode we will always pick the first waiter
+                // in the queue, as long as there are enough permits available
+                while let Some(front) = state.waiters.front() {
+                    if front.num_permits <= state.permits_available.available() {
+                        let waiter = state.waiters.pop_front().unwrap();
+                        trace!("granted {:?} permits to waiter {:?}", waiter.num_permits, waiter);
```
sarsko, line 420: Nit: either `granting` or move `trace` down to below the `acquire`

src/future/batch_semaphore.rs
Aurel300: Yes, I just want it to be clear what the comment is about. Inlining this into the condition might make it unclear: is the comment about the overall code in that condition? I think in this case (and the `try_to_acquire` case) it is easier to read.

src/future/batch_semaphore.rs
Aurel300: (see below)

src/future/batch_semaphore.rs
sarsko: I may have agreed with your reasoning if the branch actually contained any code, and here it becomes clear from the `Permits here can be acquired in one of two scenarios:` line.

Regardless: I think the `if self.closed` branch should be moved up. The current placement adds cognitive overhead when reading the function top to bottom.

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,602 @@
...
+    pub fn release(&self, num_permits: usize) {
+        if num_permits == 0 {
+            return;
+        }
+
+        let mut state = self.state.borrow_mut();
+
+        if ExecutionState::should_stop() {
+            return;
+        }
```
sarsko, line 398: Nit: ordering of this and `let mut state = self.state.borrow_mut();` should be swapped

src/future/batch_semaphore.rs
```diff
@@ -0,0 +1,602 @@
...
+            // Otherwise, the waiter is added to the queue.
+            //
+            // 3. is slightly different: the future was polled, even though the
+            // waiter was already in the queue. This can happen either because
+            // the semaphore just received some permits and woke the waiter up,
+            // or because the future itself was polled manually. Either way,
+            // the semaphore is queried.
+            let try_to_acquire = match (self.semaphore.state.borrow().fairness, is_queued) {
+                // written this way to mirror the cases described above
+                (Fairness::StrictlyFair, false) | (Fairness::Unfair, true) | (Fairness::Unfair, false) => true,
```
sarsko, line 549: Nit: `(Fairness::Unfair, _)`

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,654 @@
...
+}
+
+/// Shows a case in which causality tracking in the batch semaphore is
+/// currently imprecise. The test sets up a semaphore with two permits and two
+/// threads, each of which acquires one permit, then releases one permit, then
+/// repeats. Neither thread can be blocked in this situation, and so neither
+/// thread should causally depend on the other, but currently they do.
+#[test]
+#[should_panic(expected = "doesn't satisfy predicate")]
+fn batch_semaphore_clock_imprecise() {
```
Aurel300, line 257: @jorajeev Here is a test which currently fails but in theory should succeed if our causality tracking was "precise".

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,654 @@
...
+use shuttle::future::{self, batch_semaphore::*};
+use shuttle::{check_dfs, check_random, current, thread};
+use std::collections::HashSet;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::sync::Mutex;
+use test_log::test;
+
+#[test]
+fn batch_semaphore_basic() {
```
sarsko, line 13: nit: Do we need the `batch_semaphore` prefix for all of the tests — they are in the `batch_semaphore.rs` file after all.

tests/future/batch_semaphore.rs
```diff
@@ -0,0 +1,654 @@
...
+    }
+}
+
+/// Shows a case in which causality tracking in the batch semaphore is
+/// currently imprecise. The test sets up a semaphore with two permits and two
+/// threads, each of which acquires one permit, then releases one permit, then
+/// repeats. Neither thread can be blocked in this situation, and so neither
+/// thread should causally depend on the other, but currently they do.
+#[test]
+#[should_panic(expected = "doesn't satisfy predicate")]
```
sarsko, line 256: This shouldn't be `should_panic`, but `ignore` no ?

sarsko: Yes. Why do we want to waste CPU cycles to confirm that we have not improved Shuttle? I like the mechanical-ness of "oh hey btw here's a test which now works", but that could also be gotten by running this test (and optionally other tests) in CI.

Not blocking for me, I'll leave it for others to decide.

jorajeev: Actually, I asked Aurel to add this, so we have a scenario showing that our causality tracking is imprecise.

jorajeev (approved):

tests/future/batch_semaphore.rs
sarsko: > Actually, I asked Aurel to add this, so we have a scenario showing that our causality tracking is imprecise.

?

I don't disagree on the test but whether it should be `ignore` (and not `should_panic`) or `should_panic` (and not `ignore`)

src/future/batch_semaphore.rs
bkragl: Should we track this in an issue?

tests/future/batch_semaphore.rs
bkragl: My two cents: with the explanation in the comment, the risk of someone mistaking this test as desirable behavior is pretty low. I think it's better to run the test than not to run it. If the test starts to break without addressing the issue, we'd like to know about it, instead of hoping that it still does what we want when running it manually.

PR #152, opened 2024-07-16
Add test to check that stdlib mutex is unfair

Aurel300:
The Shuttle version of `Mutex` is indeed unfair, but no test catches this if it isn't. This PR adds such a test. (I want this in preparation for rewriting `Mutex` on top of batch semaphore.)

### DIFF
--- a/tests/basic/mutex.rs
+++ b/tests/basic/mutex.rs
@@ -377,3 +377,69 @@ fn mutex_into_inner() {
         None,
     )
 }
+
+/// Checks that the mutex is unfair, i.e. if there are two threads blocked
+/// on the same mutex, unlocking the mutex may unblock either thread.
+#[test]
+fn mutex_unfair() {
+    let observed_values = Arc::new(std::sync::Mutex::new(HashSet::new()));
+    let observed_values_clone = Arc::clone(&observed_values);
+
+    check_random(
+        move || {
+            let lock = Arc::new(Mutex::new(vec![]));
+            let guard = lock.lock().unwrap();
+
+            // Here we use a stdlib mutex to avoid introducing yield points.
+            // It is used to record in which order the two threads were
+            // enqueued, because `thread::spawn` is a yield point and as such
+            // the enqueing can happen in either order.
+            let order = Arc::new(std::sync::Mutex::new(vec![]));
+            let threads = (0..2)
+                .map(|tid| {
+                    let lock = lock.clone();
+                    let order = order.clone();
+                    thread::spawn(move || {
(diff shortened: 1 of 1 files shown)


tests/basic/mutex.rs
```diff
@@ -377,3 +377,54 @@ fn mutex_into_inner() {
         None,
     )
 }
+
+/// Checks that the stdlib mutex is unfair, i.e. if there are two threads
```
bkragl, line 381: "stdlib" might be confusing here. I'd drop it.

tests/basic/mutex.rs
```diff
@@ -377,3 +377,54 @@ fn mutex_into_inner() {
...
+                .map(|(tid, blocked)| {
+                    let lock = lock.clone();
+                    let blocked = blocked.clone();
+                    thread::spawn(move || {
+                        // once the flag is set here and observed in the busy
+                        // loop below, the thread is assumed to be blocked,
+                        // because there is no yield point between the store
+                        // and the mutex lock
+                        blocked.store(true, std::sync::atomic::Ordering::SeqCst);
+                        lock.lock().unwrap().push(tid);
```
bkragl, line 408: I'm not sure I understand the current test. Fairness means that threads are granted the lock in the order they get blocked, right? So to show unfairness, we need to track the order in which threads get blocked.

So what I'd expect is `blocked` to be a `std::sync::Mutex<Vec<usize>>` in which we also push `tid`, and then we assert that we see all four combinations. Am I missing something?

Aurel300: Good catch, I missed this: I now added the mutex that you are suggesting, *in addition* to the atomics (because there is no good way to block without using another sync. primitive otherwise). I tried to make sure the test would catch an unfair mutex by adding a `break` [here](https://github.com/awslabs/shuttle/blob/9076a444c7f611eeb94fa0b562263bf29a178fbd/src/sync/mutex.rs#L279), but I did not realise that `waiters` in the Shuttle mutex is a `TaskSet`, which iterates in order of increasing task IDs, so I made the semaphore unfair, but biased specifically to lower task IDs.

bkragl: Can't we eliminate the atomics by making the while condition `order.lock().unwrap().len() != 2`?

Aurel300: Ah, right. Fixed. I tried to block on the mutex with a condvar (which does not work because it actually blocks the main thread without dropping back to shuttle), but we can of course busyloop on the mutex itself.

bkragl (approved):

PR #155, opened 2024-07-24
`std::sync::Mutex` using batch semaphore

Aurel300:
Reimplements `std::sync::Mutex` using batch semaphore. The tests already include cases to check causal dependencies and the number of context switches, so these have not changed.

This PR changes `block_on` to not yield before the future is polled for the first time. This caused one replay test to fail. I'm not sure if we should bump versions for this, since it is technically a breaking change for schedules?

We can delay this PR until after we start the crate reorganisation, but this is ready for reviews in any case.

---

While trying to make sure the new semaphore implementation is equivalent to the old one in terms of context switches, I found it useful to write out how the methods work at a high-level, including their yield points. Including this here in case it's useful:

- `Mutex::lock` when blocked
  - check `available_permits` (0 available)
  - (re-entrant deadlock check)
  - `BatchSemaphore::acquire_blocking(1)` blocked
    - check state (-> not enough permits)
    - enqueue waiter
    - **YIELD** (`block_on` poll yield; wait until woken)
    - check and update state (-> permits now available)
    - remove waiter
    - block other failed waiters
    - **YIELD** (let other threads fail a `try_acquire`)
  - (`acquire_blocking` returns)
  - update state
  - acquire inner mutex
- `Mutex::lock` when not blocked
  - check `available_permits` (1)
  - `BatchSemaphore::acquire_blocking(1)` not blocked
    - check state (-> sufficient permits)
    - update state
    - block other failed waiters
    - **YIELD** (let other threads fail a `try_acquire`)
  - (`acquire_blocking` returns)
  - update state
  - acquire inner mutex
- `Mutex::try_lock` failed
  - `BatchSemaphore::try_acquire(1)` failed
    - **YIELD** (so another `try_acquire` in this thread may succeed)
  - (`try_acquire` returns)
- `Mutex::try_lock` successful
  - `BatchSemaphore::try_acquire(1)` successful
    - update state
    - block other failed waiters
    - **YIELD** (let other threads fail a `try_acquire`)
  - (`try_acquire` returns)
  - update state
  - acquire inner mutex
- `Mutex::drop`
  - release inner mutex
  - update state
  - `BatchSemaphore::release(1)`
    - update state
    - unblock all feasible waiters
    - **YIELD** (allow waiters to race and wake)
  - (`release` returns)
- `Mutex::drop` when panicking
  - release inner mutex
  - update state
  - `BatchSemaphore::release(1)`
    - update state
    - forget waiters
    - close semaphore
  - (`release` returns)

---

### DIFF
--- a/src/future/batch_semaphore.rs
+++ b/src/future/batch_semaphore.rs
@@ -185,17 +185,16 @@ struct BatchSemaphoreState {
     // (4) closed ==> waiters.is_empty()
     waiters: VecDeque<Arc<Waiter>>,
     permits_available: PermitsAvailable,
-    fairness: Fairness,
     // TODO: should there be a clock for the close event?
     closed: bool,
 }
 
 impl BatchSemaphoreState {
-    fn acquire_permits(&mut self, num_permits: usize) -> Result<(), TryAcquireError> {
+    fn acquire_permits(&mut self, num_permits: usize, fairness: Fairness) -> Result<(), TryAcquireError> {
         assert!(num_permits > 0);
         if self.closed {
             Err(TryAcquireError::Closed)
-        } else if self.waiters.is_empty() || matches!(self.fairness, Fairness::Unfair) {
+        } else if self.waiters.is_empty() || matches!(fairness, Fairness::Unfair) {
             // Permits here can be acquired in one of two scenarios:
             // - The waiter queue is empty; nobody else is waiting for permits,
             //   so if there are enough available, immediately succeed.
@@ -223,6 +222,7 @@ impl BatchSemaphoreState {
 #[derive(Debug)]
 pub struct BatchSemaphore {
     state: RefCell<BatchSemaphoreState>,
+    fairness: Fairness,
 }
 
 /// Error returned from the [`BatchSemaphore::try_acquire`] function.
@@ -262,10 +262,9 @@ impl BatchSemaphore {
         let state = RefCell::new(BatchSemaphoreState {
             waiters: VecDeque::new(),
             permits_available: PermitsAvailable::new(num_permits),
-            fairness,
             closed: false,
         });
-        Self { state }
+        Self { state, fairness }
     }
 
     /// Creates a new semaphore with the initial number of permits.
@@ -273,10 +272,9 @@ impl BatchSemaphore {
         let state = RefCell::new(BatchSemaphoreState {
             waiters: VecDeque::new(),
             permits_available: PermitsAvailable::const_new(num_permits),
-            fairness,
             closed: false,
         });
-        Self { state }
+        Self { state, fairness }
     }
 
     /// Returns the current number of available permits.
@@ -326,7 +324,7 @@ impl BatchSemaphore {
     /// If there aren't enough permits, returns `Err(TryAcquireError::NoPermits)`
     pub fn try_acquire(&self, num_permits: usize) -> Result<(), TryAcquireError> {
         let mut state = self.state.borrow_mut();
-        let res = state.acquire_permits(num_permits).map_err(|err| {
+        let res = state.acquire_permits(num_permits, self.fairness).map_err(|err| {
             // Conservatively, the requester causally depends on the
             // last successful acquire.
             // TODO: This is not precise, but `try_acquire` causal dependency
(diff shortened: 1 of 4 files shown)


src/sync/mutex.rs
```diff
@@ -24,78 +24,61 @@ pub struct MutexGuard<'a, T: ?Sized> {
...
-        if let Some(holder) = state.holder {
-            if holder == me {
-                panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
-            }
+        // Try to acquire without blocking first. If we succeed, we are done
+        // and can move on to acquiring the inner mutex. This is a yield point.
+        match self.semaphore.try_acquire(1) {
+            Ok(()) => (),
+            Err(TryAcquireError::Closed) => unreachable!(),
+            Err(_) => {
```
sarsko, line 57: Swap to `TryAcquireError::NoPermits`

src/sync/mutex.rs
```diff
@@ -24,78 +24,61 @@ pub struct MutexGuard<'a, T: ?Sized> {
...
+                // If the lock is already held, then we are blocked. This can
+                // happen either because the lock is held by another thread,
+                // or because the lock is held by the current thread. For the
+                // latter case, we check the state to report a more precise
+                // error message.
+                state = self.state.borrow_mut();
+                if let Some(holder) = state.holder {
```
sarsko, line 64: If this is `None`, then we have a bug in the `Mutex` – swap to `state.holder.expect(...)` ?

src/sync/mutex.rs
```diff
@@ -123,71 +100,36 @@ impl<T: ?Sized> Mutex<T> {
...
+        let result = match self.inner.try_lock() {
+            Ok(guard) => Ok(MutexGuard {
+                inner: Some(guard),
+                mutex: self,
+            }),
+            Err(TryLockError::Poisoned(guard)) => Err(TryLockError::Poisoned(PoisonError::new(MutexGuard {
+                inner: Some(guard.into_inner()),
+                mutex: self,
+            }))),
+            Err(TryLockError::WouldBlock) => panic!("mutex state out of sync"),
```
sarsko, line 131: Nit: `unreachable!` ?

src/sync/mutex.rs
Aurel300: I don't think this is the case: `state.holder` may still be `None` because there is a yield between the acquiry of a permit from the semaphore and the actual state update (the two are not performed in a single atomic step).

However, the check is still sufficient here because in the case of a thread attempting to re-acquire a lock it already holds, it must have returned from the first `lock` call by the time the second one is checking the state.

src/sync/mutex.rs
```diff
@@ -24,78 +24,59 @@ pub struct MutexGuard<'a, T: ?Sized> {
...
         let mut state = self.state.borrow_mut();
-        trace!(holder=?state.holder, waiters=?state.waiters, "waiting to acquire mutex {:p}", self);
+        trace!(holder=?state.holder, semaphore=?self.semaphore, "waiting to acquire mutex {:p}", self);
+        drop(state);
 
-        // If the lock is already held, then we are blocked
-        if let Some(holder) = state.holder {
-            if holder == me {
-                panic!("deadlock! task {:?} tried to acquire a Mutex it already holds", me);
+        if !self.semaphore.is_closed() {
```
jorajeev, line 50: Feels a bit strange that you're checking if the semaphore is closed here, but then you call `unwrap` on the acquire call on line 70.  If we're not expecting the semaphore to be closed, maybe just assert that here?

src/future/batch_semaphore.rs
```diff
@@ -391,12 +428,24 @@ impl BatchSemaphore {
...
+        let mut state = self.state.borrow_mut();
+
         if ExecutionState::should_stop() {
+            // In case we are panicking, we release permits, but also clear
+            // the waiters queue: we should not unblock the threads at this
+            // point. However, the permits are released such that future
+            // acquires may succeed, as long as the requesters were not
+            // blocking on the semaphore at the time of the panic. This is
+            // used to correctly model lock poisoning.
+            state.permits_available.release(num_permits, VectorClock::new());
```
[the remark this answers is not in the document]
Aurel300: When this is done as part of `Drop` handlers, `current::clock` may panic. I think this is the same thing as mentioned [here for the `close` clock](https://github.com/awslabs/shuttle/pull/151#discussion_r1676520645): the way execution stops needs to be changed a bit.

src/sync/mutex.rs
Aurel300: This is the other way though: *if* it is not closed, then we can `unwrap` the `acquire_blocking` call. If it is closed, then we will continue and should get a poisoned lock from the inner mutex.

jorajeev (approved):

PR #156, opened 2024-08-09
Causal replay

Aurel300:
An attempt to add causal dependency to the replay scheduler. To use this, `set_target_clock` should be invoked on a `ReplayScheduler` instance with the clock of the failure event (provided as a slice of integers). Some things to resolve still:

- [ ] To make this work, schedulers now receive references to `Task`s rather than just `TaskId`s. I think we discussed that schedulers should generally have access to more information about tasks, so if this is the approach we take, the `Task` interface might deserve some more cleanup. Even more so if schedulers are eventually intended to be defined in a separate crate.
- [ ] (minor) Any RNG steps in a schedule following a skipped task should also be scheduled. Need to add a test for this and implement.

---


src/scheduler/replay.rs
```diff
@@ -45,13 +46,21 @@ impl ReplayScheduler {
...
 
     /// Set flag to allow early termination of a schedule
     pub fn set_allow_incomplete(&mut self) {
         self.allow_incomplete = true;
     }
+
+    /// Set a clock of the failure to be reproduced. Events which are not
+    /// causally related to this clock (i.e., events concurrent to the failure)
+    /// will not be scheduled.
+    pub fn set_target_clock(&mut self, clock: &[u32]) {
```
jorajeev, line 61: Could make the second argument anything that implements `Into<VectorClock>` ?

src/scheduler/replay.rs
```diff
@@ -64,26 +73,46 @@ impl Scheduler for ReplayScheduler {
...
+                                // The target event causally depends on this
+                                // event, so we schedule it.
+                                return Some(next);
+                            } else {
+                                // The target even is concurrent with this
+                                // event, so it is irrelevant to the replay.
+                                // At this point, we also need to skip over
+                                // any random steps made by the thread that
+                                // would have been scheduled.
+                                while let Some(ScheduleStep::Random) = self.schedule.steps.get(self.steps) {
```
jorajeev, line 100: Would be good to generate a `trace!` here recording that a step was skipped (and also how many random steps).  Also add a counter to the scheduler so we can also report how many steps were skipped during a replay run.

src/scheduler/replay.rs
```diff
@@ -64,26 +76,54 @@ impl Scheduler for ReplayScheduler {
...
+                ScheduleStep::Task(next) => {
+                    if let Some(task) = runnable.iter().find(|t| t.id() == next) {
+                        self.steps += 1;
+                        if let Some(target_clock) = &self.target_clock {
+                            if task.clock <= *target_clock {
+                                // The target event causally depends on this
+                                // event, so we schedule it.
+                                return Some(next);
+                            } else {
+                                // The target even is concurrent with this
```
jorajeev, line 98: nit: typo "even" -> "event"

tests/basic/replay.rs
```diff
@@ -178,3 +178,115 @@ fn replay_persist_none() {
...
+        let schedule = Schedule::new_from_task_ids(0, vec![0, 1, 0, 2, 0, 3, 0, 1, 1, 2, 2]);
+        let mut scheduler = ReplayScheduler::new_from_schedule(schedule);
+        scheduler.set_target_clock(&[2, 2, 1]);
+        let mut config = Config::new();
+        config.failure_persistence = FailurePersistence::None;
+        let runner = Runner::new(scheduler, config);
+        runner.run(move || {
+            let lock = Arc::new(Mutex::new(0usize));
+            let lock_clone = Arc::clone(&lock);
+            thread::spawn(move || {
```
jorajeev, line 209: Since we're relying on specific TaskIds here, could you add an `assert_eq!(me(), 1)` etc to each of these threads?

tests/basic/replay.rs
```diff
@@ -178,3 +178,115 @@ fn replay_persist_none() {
...
+                // acquiring a lock" would be indistinguishable. However, we
+                // need the non-Shuttle atomic to smuggle data out of a this
+                // panicking test.
+                flag_clone.store(true, Ordering::SeqCst);
+            });
+        });
+    })
+    .expect_err("test should panic");
+    let output = result.downcast::<&str>().unwrap();
+    assert_eq!(*output, "expected panic");
```
jorajeev, line 232: Could you also add an assert about `steps_skipped` to test that functionality too?

tests/basic/replay.rs
```diff
@@ -178,3 +178,115 @@ fn replay_persist_none() {
...
+                let mut thread_rng = shuttle::rand::thread_rng();
+                thread_rng.gen::<u64>();
+                thread_rng.gen::<u64>();
+                thread_rng.gen::<u64>();
+            });
+        });
+    })
+    .expect_err("test should panic");
+    let output = result.downcast::<&str>().unwrap();
+    assert_eq!(*output, "expected panic");
```
jorajeev, line 291: Similar comments here: add assertions for `me()` and `steps_skipped`

tests/basic/replay.rs
Aurel300: I'm not sure how to do this (easily): `Runner::new` takes the scheduler we construct, so there is no access to it afterwards.

jorajeev: Okay for now. Maybe we can find a way to get additional metrics through the metrics scheduler

jorajeev (dismissed):

jorajeev (approved):

PR #157, opened 2024-08-13
`std::sync::RwLock` using batch semaphore

Aurel300:
Reimplements `std::sync::RwLock` using batch semaphore.

---

### DIFF
--- a/src/sync/rwlock.rs
+++ b/src/sync/rwlock.rs
@@ -1,13 +1,16 @@
+use crate::future::batch_semaphore::{BatchSemaphore, Fairness};
 use crate::runtime::execution::ExecutionState;
-use crate::runtime::task::clock::VectorClock;
 use crate::runtime::task::{TaskId, TaskSet};
-use crate::runtime::thread;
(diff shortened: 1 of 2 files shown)


src/sync/rwlock.rs
```diff
@@ -441,38 +345,23 @@ impl<T: ?Sized> Drop for RwLockReadGuard<'_, T> {
...
-            }
-            _ => panic!("exiting a reader but rwlock is in the wrong state {:?}", state.holder),
-        }
-
-        if ExecutionState::should_stop() {
-            return;
+        let RwLockHolder::Read(readers) = &mut state.holder else {
+            panic!("exiting a reader but rwlock is in the wrong state {:?}", state.holder);
+        };
+        let was_reader = readers.remove(self.me);
```
jorajeev, line 357: Could avoid introducing the `was_reader` variable here?

Aurel300: I think it was there for clarity, but removed now.

jorajeev (approved):

PR #161, opened 2024-09-09
Reporting & replaying failing random seeds

funemy:
- When a failing case is found by `RandomScheduler`, shuttle now reports the random seed being used in addition to the failing schedule.
    - This requires a minor change in the implementation of `RandomScheduler`. Now, on a new execution, `RandomScheduler.rng` will be reinitialized with the new seed generated from `data_source`. 
       This is necessary because the two RNGs used in `RandomScheduler` (i.e., `rng` and `data_source`), though initialized using the same seed, advances at different paces. Therefore, the two RNGs will most likely be out of sync after the first iteration, and we cannot replay a failure case with just one seed from then. This change synchronizes the two RNGs at each iteration.
       This change does not seem to affect the performance of `RnadomScheduler`.
- Support replaying the failure with just random seeds.

---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -322,13 +322,45 @@ where
 
 /// Run the given function under a randomized concurrency scheduler for some number of iterations.
 /// Each iteration will run a (potentially) different randomized schedule.
+/// When the environment variable SHUTTLE_RANDOM_SEED is set to a u64, this number will be used
+/// as the seed to initialize the random scheduler.
 pub fn check_random<F>(f: F, iterations: usize)
 where
     F: Fn() + Send + Sync + 'static,
 {
     use crate::scheduler::RandomScheduler;
+    use tracing::info;
 
-    let scheduler = RandomScheduler::new(iterations);
+    let seed_env = std::env::var("SHUTTLE_RANDOM_SEED");
+    let runner = match seed_env {
+        Ok(s) => match s.as_str().parse::<u64>() {
+            Ok(seed) => {
+                info!(
+                    "Initializing RandomScheduler with the seed provided by SHUTTLE_RANDOM_SEED: {}",
+                    seed
+                );
+                Runner::new(RandomScheduler::new_from_seed(seed, iterations), Default::default())
+            }
+            Err(err) => panic!("The seed provided by SHUTTLE_RANDOM_SEED is not a valid u64: {}", err),
+        },
+        Err(_) => Runner::new(RandomScheduler::new(iterations), Default::default()),
+    };
+
+    runner.run(f);
+}
+
+/// Run function `f` using `RandomScheduler` initialized with the provided `seed` for the given
+/// `iterations`.
+/// This makes generating the random seed for each execution independent from `RandomScheduler`.
+/// Therefore, this can be used with a library (like proptest) that takes care of generating the
+/// random seeds.
+pub fn check_random_with_seed<F>(f: F, seed: u64, iterations: usize)
+where
+    F: Fn() + Send + Sync + 'static,
+{
+    use crate::scheduler::RandomScheduler;
+
+    let scheduler = RandomScheduler::new_from_seed(seed, iterations);
     let runner = Runner::new(scheduler, Default::default());
     runner.run(f);
 }
--- a/src/scheduler/random.rs
+++ b/src/scheduler/random.rs
@@ -18,6 +18,33 @@ pub struct RandomScheduler {
     rng: Pcg64Mcg,
     iterations: usize,
     data_source: RandomDataSource,
+    current_seed: CurrentSeedDropGuard,
+}
+
+#[derive(Debug, Default)]
+struct CurrentSeedDropGuard {
+    inner: Option<u64>,
+}
+
+impl CurrentSeedDropGuard {
+    fn clear(&mut self) {
+        self.inner = None
+    }
+
+    fn update(&mut self, seed: u64) {
+        self.inner = Some(seed)
+    }
+}
+
+impl Drop for CurrentSeedDropGuard {
+    fn drop(&mut self) {
+        if let Some(s) = self.inner {
+            eprintln!(
+                "failing seed:\n\"\n{}\n\"\nTo replay the failure, either:\n    1) pass the seed to `shuttle::check_random_with_seed, or\n    2) set the environment variable SHUTTLE_RANDOM_SEED to the seed and run `shuttle::check_random`.",
+                s
+            )
+        }
+    }
 }
 
 impl RandomScheduler {
@@ -37,6 +64,7 @@ impl RandomScheduler {
             rng,
             iterations: 0,
             data_source: RandomDataSource::initialize(seed),
+            current_seed: CurrentSeedDropGuard::default(),
         }
     }
 }
@@ -44,10 +72,14 @@ impl RandomScheduler {
 impl Scheduler for RandomScheduler {
     fn new_execution(&mut self) -> Option<Schedule> {
         if self.iterations >= self.max_iterations {
+            self.current_seed.clear();
             None
         } else {
             self.iterations += 1;
-            Some(Schedule::new(self.data_source.reinitialize()))
+            let seed = self.data_source.reinitialize();
+            self.rng = Pcg64Mcg::seed_from_u64(seed);
+            self.current_seed.update(seed);
+            Some(Schedule::new(seed))
         }
     }
 
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -6,7 +6,7 @@ mod demo;
 mod future;
 
 use shuttle::scheduler::{ReplayScheduler, Scheduler};
-use shuttle::{replay_from_file, Config, FailurePersistence, Runner};
+use shuttle::{check_random_with_seed, replay_from_file, Config, FailurePersistence, Runner};
 use std::panic::{self, RefUnwindSafe, UnwindSafe};
 use std::sync::Arc;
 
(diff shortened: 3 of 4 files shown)


jorajeev (commented):
Thanks for the PR.  Two major comments:

(1) The PCT scheduler is also initialized with a random seed - could you extend it in the same way you've extended the RandomScheduler

(2) Could you add a `check_random_with_seed` function that initializes the RandomScheduler with a given seed (and similarly add a `check_pct_with_seed`).  This would be useful when using Shuttle with a library like proptest which would generate the initial random seed.

src/lib.rs
```diff
@@ -414,6 +414,49 @@ where
     runner.run(f);
 }
 
+/// Run the given function using random scheduler according to a seed.
+///
+/// This function allows deterministic replay of a failing schedule, as long as `f` contains no
+/// non-determinism other than that introduced by scheduling.
```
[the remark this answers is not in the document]
funemy: I updated the comment to make it clearer.

funemy:
@jorajeev I've updated the PR according to our discussion, ready for another review:

- Added `check_random_with_seed`
- Updated the comments about uncontrolled non-determinism
- Leaving PCT as it is for now.

jamesbornholt:
So the reason we didn’t implement replay this way the first time is that you lose the ability to detect divergence. If you’re debugging a failure from a schedule and insert some new code that happens to perform a synchronization operation, you’ll just silently start executing a different schedule from that point forward. The current version of replay is fairly likely to detect this kind of divergence.

I don’t have the context for why you want this change, but I’d like to make sure we don’t give up that protection in the common case. (And actually I’d love if we could make it even stronger, because it’s not guaranteed at the moment)

jorajeev:
> So the reason we didn’t implement replay this way the first time is that you lose the ability to detect divergence. If you’re debugging a failure from a schedule and insert some new code that happens to perform a synchronization operation, you’ll just silently start executing a different schedule from that point forward. The current version of replay is fairly likely to detect this kind of divergence.
> 
> I don’t have the context for why you want this change, but I’d like to make sure we don’t give up that protection in the common case. (And actually I’d love if we could make it even stronger, because it’s not guaranteed at the moment)

The need for this replay arises because we often run Shuttle in build pipelines where saving a failing schedule to a file is not easy.  (And printing to console is not feasible either, because build logs can get truncated.). For such situations, it's a lot more convenient to just print the single random seed that can be used to reproduce the failure locally.

Note that we're adding a new replay function, so the old one is still available (and the default if you use `check_random` as before).

You raise a good point about strengthening the check to avoid replay confusion when code changes.  I expect we could store additional information (maybe vector clocks, in some compressed form) to achieve this.

src/lib.rs
```diff
@@ -333,6 +333,21 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a random seed for one iteration.
+/// This effectively makes generating the random seed for each execution independent from
+/// `RandomScheduler`. Therefore, this can be used with a library (like proptest) that takes care of
+/// generating the random seeds.
+pub fn check_random_with_seed<F>(f: F, seed: u64)
```
jorajeev, line 340: This needs to take the number of iterations as a parameter, so it should be:

`pub fn check_random_with_seed<F>(f: F, seed: usize, iterations: usize)`

funemy: updated.

src/lib.rs
```diff
@@ -333,6 +333,21 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a random seed for one iteration.
```
sarsko, line 336: It isn't clear what `under a random seed for one iteration` means. Also the `one iteration` part is wrong. We run the function `f` under a `RandomScheduler` initialized with the provided `seed` for `iterations` iterations.

src/lib.rs
```diff
@@ -333,6 +333,21 @@ where
...
+/// This effectively makes generating the random seed for each execution independent from
```
sarsko, line 337: Nit: remove `effectively`

src/lib.rs
```diff
@@ -414,6 +429,52 @@ where
     runner.run(f);
 }
 
+/// Run the given function using random scheduler according to a seed.
```
sarsko, line 432: Nit: `according to a seed` -> `initialized with the provided seed` (or similar)

src/lib.rs
```diff
@@ -414,6 +429,52 @@ where
...
+///
+/// This function allows deterministic replay of a failing schedule, as long as `f` contains no
+/// uncontrolled non-determinism. Uncontrolled non-determinism can be checked using
+/// `check_uncontrolled_nondeterminism`.
+///
+/// This is a convenience function for constructing a [`Runner`] that uses
+/// [`RandomScheduler::new_from_seed`](scheduler::RandomScheduler::new_from_seed) for one
+/// iteration.
+pub fn replay_random_from_seed<F>(f: F, seed: u64)
```
sarsko, line 441: Not sure there is any point in having this function. This is just `check_random_with_seed(f, seed, 1)`.

src/scheduler/random.rs
```diff
@@ -18,6 +18,33 @@ pub struct RandomScheduler {
...
+    fn update(&mut self, seed: u64) {
+        self.inner = Some(seed)
+    }
+}
+
+impl Drop for CurrentSeedDropGuard {
+    fn drop(&mut self) {
+        if let Some(s) = self.inner {
+            eprintln!(
+                "failing seed:\n\"\n{}\n\"\npass the seed to `shuttle::replay_from_seed` to replay the failure.",
```
sarsko, line 43: If we decide to do replay based on environment variables, then the comment here needs to be updated.

tests/mod.rs
```diff
@@ -94,6 +94,25 @@ where
     // `FailurePersistence`s for each test
 }
 
+/// Validates that replay-from-seed works by running a failing seed found by a random scheduler for one
```
sarsko, line 97: Nit: `replay-from-seed` isn't anything. Either name the function (`replay_random_from_seed`) or "abstract over it"; "the replay from seed functionality".

tests/mod.rs
```diff
@@ -94,6 +94,25 @@ where
...
+where
+    F: Fn() + Send + Sync + UnwindSafe + 'static,
+{
+    let result = {
+        panic::catch_unwind(move || {
+            replay_random_from_seed(test_func, seed);
+        })
+        .expect_err("replay should panic")
+    };
+    let new_output = result.downcast::<String>().unwrap();
```
sarsko, line 110: Nit: `new_output`: there is no other `output`.

tests/mod.rs
```diff
@@ -94,6 +94,25 @@ where
...
+    let new_schedule = parse_schedule::from_stdout(&new_output).expect("output should contain a schedule");
```
sarsko, line 111: nit: `new_schedule` is non-descriptive. Maybe `schedule_from_replay` ? Maybe rename `schedule` in the params to `expected_schedule` ?

src/lib.rs
funemy: oops, good catch -- I forgot to update the comment

funemy:
@sarsko all comments should be addressed, the big changes are:

- `reply_random` and `reply_random_from_seed` are removed
- `check_random` now first tries reading from the env var `SHUTTLE_REPLAY_SEED` and if a valid seed is read, initializes the scheduler with it
- error messages printed by `CurrentSeedDropGuard` is updated.

src/lib.rs
```diff
@@ -322,13 +322,46 @@ where
...
+/// When the environment variable SHUTTLE_REPLAY_SEED is set to a u64, this number will be used
+/// as the seed to initialize the random scheduler.
 pub fn check_random<F>(f: F, iterations: usize)
 where
     F: Fn() + Send + Sync + 'static,
 {
     use crate::scheduler::RandomScheduler;
 
-    let scheduler = RandomScheduler::new(iterations);
+    let seed_env = std::env::var("SHUTTLE_REPLAY_SEED");
```
sarsko, line 333: Not sure how much I care, but calling it `SHUTTLE_REPLAY_SEED` is odd, as it strictly speaking is independent of replaying.

sarsko (commented):
Looks good to me, a couple of minor comments. Would like one more look from Rajeev/James/Bernhard.

sarsko (dismissed):

src/lib.rs
funemy: maybe `SHUTTLE_RANDOM_SEED`?

jorajeev: Yeah, given the changes, let's go with `SHUTTLE_RANDOM_SEED`

src/lib.rs
```diff
@@ -322,13 +322,46 @@ where
...
+    let runner;
```
jorajeev, line 334: The idiomatic way would be to write

`let runner = match seed_env { .. };`

src/lib.rs
```diff
@@ -322,13 +322,46 @@ where
...
+
+    match seed_env {
+        Ok(s) => match s.as_str().parse::<u64>() {
+            Ok(seed) => {
+                println!(
+                    "Initializing RandomScheduler with the seed provided by SHUTTLE_REPLAY_SEED: {}",
+                    seed
+                );
```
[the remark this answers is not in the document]
jorajeev: +1

src/scheduler/random.rs
```diff
@@ -18,6 +18,33 @@ pub struct RandomScheduler {
...
+    current_seed: CurrentSeedDropGuard,
+}
+
+#[derive(Debug, Default)]
+struct CurrentSeedDropGuard {
+    inner: Option<u64>,
+}
+
+impl CurrentSeedDropGuard {
+    fn wipe(&mut self) {
```
jorajeev, line 30: Let's call this `clear` (or `reset`)  instead of `wipe`

funemy:
@sarsko @jorajeev all comments addressed

jorajeev (approved):

PR #163, opened 2024-09-23
Shuttle Explorer

Aurel300:
Added mechanism to emit annotated schedules from Shuttle. Added VS Code extension "Shuttle Explorer" to visualise such annotated schedules.

- [x] `cfg` flag to disable annotations completely (to avoid performance cost of checking a thread-local repeatedly)
- [ ] document and clean up extension

---

### DIFF
--- a/shuttle-explorer/README.md
+++ b/shuttle-explorer/README.md
@@ -1,0 +1,34 @@
+# Shuttle Explorer
+
+Timeline visualiser for [Shuttle](https://github.com/awslabs/shuttle).
+
+## Building
+
+To use the extension, it is currently required to build it from source. Pre-requisites include:
+
+- VS Code 1.92.0 or later
+- `node` 18.20 and `npm` 10.5 (other versions are probably fine too)
+- [esbuild Problem Matcher extension](https://marketplace.visualstudio.com/items?itemName=connor4312.esbuild-problem-matchers), installed either globally or in the Shuttle Explorer workspace. **Without this extension the builds will fail silently!**
+
+Steps:
+
+1. Clone this repository.
+2. Navigate to the Shuttle Explorer directory in a terminal, run `npm install`.
+3. Open the Shuttle Explorer directory in VS Code.
+4. Press F5 or find "Run > Start Debugging" in the menu. Accept a debug configuration if prompted.
+  - In the background, two tasks should spawn: a TypeScript watcher, and an esbuild watcher. You should be able to see both in the "terminal" window, and neither should be reporting an error. If the esbuild watcher is marked as "pending", you need the esbuild Problem Matcher extension.
+  - The TypeScript sources in `src` should be compiled down to JavaScript, output into the `dist` folder.
+5. A new VS Code window (the "extension host") should open.
+6. (Optionally,) open a workspace in the new window, to see the code cursors.
+7. From the command palette (cmd + shift + P), find "Shuttle Explorer: Focus on Home View" to reveal the extension.
+8. Press the "file" button in the top-left corner of the extension panel, or use "View annotated schedule" in the command palette to pick a JSON file to visualise.
+
+Notes:
+
+- Things may go wrong before the extension host window is open. Look at the problems panel, see if there are any problems. Look at the terminal window, to see if the background tasks are running and not reporting any errors.
+- Things may go wrong after the extension host window is open. In that window, look at the console (command palette: "Developer: Open Webview Developer Tools") to look for any errors.
+- After closing the Shuttle Explorer panel, re-opening it will probably not work. Reload the window (cmd + R or command palette: "Developer: Reload Window") to start again.
+
+During development:
+
+- While the extension is running, changing TypeScript files in `src` should trigger an automatic re-compilation. Simply reload the window (cmd + R or command palette: "Developer: Reload Window") with the extension to see the new changes. (Sometimes, the window spontaneously closes after reloading it. Re-start with F5 if this happens.)
--- a/src/annotations/mod.rs
+++ b/src/annotations/mod.rs
@@ -1,5 +1,5 @@
 //! Annotated schedules. When an execution is scheduled using the
-//! [`AnnotationScheduler`], Shuttle will produce a file that contains
+//! [`crate::scheduler::AnnotationScheduler`], Shuttle will produce a file that contains
 //! additional information about the execution, such as the kind of step that
 //! was taken (was a task created, were permits acquired from a semaphore, etc)
 //! as well as the task's vector clocks and thus any causal dependence between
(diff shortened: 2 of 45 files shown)


src/annotations/mod.rs
```diff
@@ -0,0 +1,447 @@
...
+
+        pub(crate) fn record_semaphore_try_acquire(object_id: ObjectId, num_permits: usize, successful: bool) {
+            record_event(AnnotationEvent::SemaphoreTryAcquire(object_id, num_permits, successful));
+        }
+
+        pub(crate) fn record_semaphore_release(object_id: ObjectId, num_permits: usize) {
+            record_event(AnnotationEvent::SemaphoreRelease(object_id, num_permits));
+        }
+
+        pub(crate) fn record_task_created(task_id: TaskId, future: bool) {
```
jorajeev, line 303: nit: rename argument to `is_future`

src/annotations/mod.rs
```diff
@@ -0,0 +1,447 @@
...
+                Lazy::new(|| Regex::new(r"([0-9]+): ([^\n]+)\n +at (\./src/[^:]+):([0-9]+):([0-9]+)\b").unwrap());
+            with_state(move |state| {
+                let task_id = state.last_task_id.expect("no last task ID?"); // unwrap_or(crate::current::me());
+                                                                             /*if let Some(last_task_id) = state.last_task_id {
+                                                                                 assert_eq!(task_id, last_task_id);
+                                                                             }*/
+
+                let task_id_num = usize::from(task_id);
+                assert!(task_id_num < state.tasks.len());
+                state.tasks[task_id_num].first_step = state.tasks[task_id_num].first_step.min(state.events.len());
```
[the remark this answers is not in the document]
Aurel300: This `min` will only do something once, yes. However, it should be the first step of *this* task, not its creator, so we can't do it when the task is created (because its first step will either be later, or it might even never take a step).

src/annotations/mod.rs
```diff
@@ -0,0 +1,447 @@
...
+                state.tasks[task_id_num].last_step = state.tasks[task_id_num].last_step.max(state.events.len());
+
+                use std::backtrace::Backtrace;
+                let bt = Backtrace::capture(); // TODO: config to disable
```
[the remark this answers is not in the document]
Aurel300: The option to disable is to disable backtraces. If we want this to be configurable separately from `RUST_BACKTRACE`, we can use `force_capture` and our own flag. I added a comment.

jorajeev (dismissed):

shuttle-explorer/README.md
```diff
@@ -0,0 +1,34 @@
+# Shuttle Explorer
+
+Timeline visualiser for [Shuttle](https://github.com/awslabs/shuttle).
+
+## Building
+
+To use the extension, it is currently required to build it from source. Pre-requisites include:
```
sarsko, line 7: Do you know how much hassle it is to make an extension available in the VS Code store?

shuttle-explorer/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+- VS Code 1.92.0 or later
+- `node` 18.20 and `npm` 10.5 (other versions are probably fine too)
+- [esbuild Problem Matcher extension](https://marketplace.visualstudio.com/items?itemName=connor4312.esbuild-problem-matchers), installed either globally or in the Shuttle Explorer workspace. **Without this extension the builds will fail silently!**
```
sarsko, line 11: Why do we need this?

src/annotations/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+        }
+
+        impl Serialize for TaskId {
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: serde::ser::Serializer,
+            {
+                usize::from(*self).serialize(serializer)
+            }
+        }
```
sarsko, line 160: Nit-ish:

Should all the serde stuff be moved somewhere else and put behind a feature flag "serde" (and "annotation" would enable "serde" as well)? It doesn't belong here and it also might be useful for other stuff to have a way to get serde for the types.

src/annotations/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+                // TODO: for now, "local to the project" is detected based on
+                //       the path starting with `./src/`. Find an alternative
+                //       way to do this?
+                // The following regex matches frames with local paths. We rely
+                // on the string format of the backtrace because there is no
+                // better API. At the time of writing, even the unstable feature
+                // `backtrace_frames` does not provide a better interface.
+                // https://doc.rust-lang.org/std/backtrace/struct.BacktraceFrame.html
+                static RE: Lazy<Regex> =
+                    Lazy::new(|| Regex::new(r"([0-9]+): ([^\n]+)\n +at (\./src/[^:]+):([0-9]+):([0-9]+)\b").unwrap());
```
sarsko, line 197: Would be nice to have a "step-by-step" comment sequence for the RegEx I think.

src/annotations/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+
+                // Whether or not the following call actually captures a backtrace
```
sarsko, line 199: I'm finding the code below hard to follow. Would be nice with some more comments, better names, more spacing etc.

src/scheduler/annotation.rs
```diff
@@ -0,0 +1,47 @@
+use crate::annotations::{record_random, record_schedule, start_annotations, stop_annotations};
+use crate::runtime::task::{Task, TaskId};
+use crate::scheduler::{Schedule, Scheduler};
+
+/// An `AnnotationScheduler` wraps an inner `Scheduler` and enables the
+/// creation of an annotated schedule (for use with Shuttle Explorer).
+/// Without enabling the feature `annotation`, this wrapper does nothing.
+///
+/// Importantly, only one `AnnotatedScheduler` instance should exist at a time.
```
sarsko, line 9: Nit: `AnnotationScheduler`

src/sync/mutex.rs
```diff
@@ -217,3 +217,17 @@ impl<T: Display + ?Sized> Display for MutexGuard<'_, T> {
         (**self).fmt(f)
     }
 }
+
+impl<T> crate::annotations::WithName for &Mutex<T> {
+    fn with_name_and_kind(self, name: Option<&str>, kind: Option<&str>) -> Self {
+        (&self.semaphore).with_name_and_kind(name, kind.or(Some("shuttle::sync::Mutex")));
```
sarsko, line 223: Why is this `kind.or`ed and the one below is not?

Aurel300: The one below is only forwarding the call to this implementation.

shuttle-explorer/README.md
Aurel300: As explained: without the "problem matcher", VS Code does not really understand that a background task is a background task. Some additional context [here](https://stackoverflow.com/a/49034830). This extension is also recommended by the VS Code guide for using `esbuild` for building extensions, [here](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#run-esbuild).

PR #167, opened 2024-12-03
Fix BatchSemaphore bug causing deadlocks

jorajeev:
This PR fixes a bug in BatchSemaphore that caused deadlocks in the presence of tasks that gave up waiting for requested permits, under the policy of strict fairness.

---

### DIFF
--- a/src/future/batch_semaphore.rs
+++ b/src/future/batch_semaphore.rs
@@ -238,6 +238,49 @@ impl BatchSemaphoreState {
             Err(TryAcquireError::NoPermits)
         }
     }
+
+    fn unblock_waiters_from_front(&mut self) {
+        while let Some(front) = self.waiters.front() {
+            if front.num_permits <= self.permits_available.available() {
+                let waiter = self.waiters.pop_front().unwrap();
+
+                crate::annotations::record_semaphore_acquire_unblocked(
+                    self.id.unwrap(),
+                    waiter.task_id,
+                    waiter.num_permits,
+                );
+
+                // The clock we pass into the semaphore is the clock of the
+                // waiter, corresponding to the point at which the waiter was
+                // enqueued. The clock we get in return corresponds to the
+                // join of the clocks of the acquired permits, used to update
+                // the waiter's clock to causally depend on the release events.
+                let clock = self
+                    .permits_available
+                    .acquire(waiter.num_permits, waiter.clock.clone())
+                    .unwrap();
+                trace!("granted {:?} permits to waiter {:?}", waiter.num_permits, waiter);
+
+                // Update waiter state as it is no longer in the queue
+                assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+                assert!(!waiter.has_permits.swap(true, Ordering::SeqCst));
+                ExecutionState::with(|s| {
+                    let task = s.get_mut(waiter.task_id);
+                    assert!(!task.finished());
+                    // The acquiry is causally dependent on the event
+                    // which released the acquired permits.
+                    task.clock.update(&clock);
+                    task.unblock();
+                });
+                let mut maybe_waker = waiter.waker.lock().unwrap();
+                if let Some(waker) = maybe_waker.take() {
+                    waker.wake();
+                }
+            } else {
+                return;
+            }
+        }
+    }
 }
 
 /// Counting semaphore
@@ -447,6 +490,20 @@ impl BatchSemaphore {
 
         state.waiters.remove(index).unwrap();
         assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+
+        match self.fairness {
+            Fairness::StrictlyFair => {
+                if index == 0 {
+                    // If the semaphore is strictly fair, and we removed the first waiter, check if its
+                    // removal unblocks remaining waiters.  This can happen in the following situation:
+                    // - the semahore has 1 permit available
+                    // - there are 2 waiters W1 and W2 where W1 wants 2 permits, and W2 wants 1 permit
+                    // - if W1 gives up and drops out, we want to ensure W2 is granted the semaphore
+                    state.unblock_waiters_from_front();
+                }
+            }
+            Fairness::Unfair => {}
+        }
     }
 
     /// Acquire the specified number of permits (async API)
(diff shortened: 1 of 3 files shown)


src/future/batch_semaphore.rs
```diff
@@ -447,6 +492,15 @@ impl BatchSemaphore {
 
         state.waiters.remove(index).unwrap();
         assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+
+        if matches!(self.fairness, Fairness::StrictlyFair) && index == 0 {
```
sarsko, line 496: Nit-ish: I don't think `matches!` is well suited here. We want to be forced to have a look at this if we change `Fairness`.

src/future/batch_semaphore.rs
```diff
@@ -447,6 +492,15 @@ impl BatchSemaphore {
...
+            // If the semaphore is strictly fair, and we removed the first waiter, check if its
+            // removal unblocks remaining waiters.  This can happen in the following situation:
+            // - the semahore has 2 permits available
+            // - there are 3 waiters W1, W2, W3 where W1 wants 3 permits, and W2,W3 want 1 permit each
```
sarsko, line 500: Nit: W2, W3 or W2 and W3 (ie, either add spacing or swap out , with and)

src/future/batch_semaphore.rs
```diff
@@ -447,6 +492,15 @@ impl BatchSemaphore {
...
+            // - if W3 is removed (because it drops out), we want to ensure W2,W3 are granted the semaphore
```
sarsko, line 501: Nit: W2, W3 or W2 and W3

Nit: "ensure that"

sarsko: Nit: "if W3 is removed" -> "if W1 is removed"

src/future/batch_semaphore.rs
```diff
@@ -447,6 +492,15 @@ impl BatchSemaphore {
 
         state.waiters.remove(index).unwrap();
         assert!(waiter.is_queued.swap(false, Ordering::SeqCst));
+
+        if matches!(self.fairness, Fairness::StrictlyFair) && index == 0 {
+            // If the semaphore is strictly fair, and we removed the first waiter, check if its
+            // removal unblocks remaining waiters.  This can happen in the following situation:
```
sarsko, line 498: Nit-ish: When I read "This can happen in the following situation", I usually expect the situation abstractly, and not an example.

Also: Example is bigger than it needs to be. If you have already decided that W1 is going to attempt to acquire an unacquireable amount, then we can make do with 1 permit, W1 wanting to acq 2, and W2 wanting to acq 1, then dropping W1.

Off-topic: do we have checks that we never try to acquire more than the maximum number of permits?

src/future/batch_semaphore.rs
```diff
@@ -501,48 +555,9 @@ impl BatchSemaphore {
...
-                        let mut maybe_waker = waiter.waker.lock().unwrap();
-                        if let Some(waker) = maybe_waker.take() {
-                            waker.wake();
-                        }
-                    } else {
-                        break;
-                    }
-                }
+                // in a strictly fair mode we will grant permits to waiters from the front
+                // of the queue, as long as there are enough permits available
```
sarsko, line 559: This comment is now repeated here and in `unblock_waiters_from_front`. We probably want it removed in `unblock_waiters_from_front` ?

tests/future/batch_semaphore.rs
```diff
@@ -512,52 +512,63 @@ fn bugged_cleanup_would_cause_deadlock() {
...
+// This test exercises scenarios to ensure that the BatchSemaphore behaves correctly in the presence
+// of tasks that drop an `Acquire` guard without waiting for the semaphore to become available.
 //
-// At this point, T2 should be woken up and should get the permit.
-// Unfortunately, in an earlier version of BatchSemaphore, this was not happening because
-// the Drop handler for Acquire was only returning permits to the semaphore, but not
-// waking up the next waiter in the queue.
+// The general idea is that there are 3 types of tasks: `EarlyDrop`, `Hold` and `Release` tasks
+// (determined by the `Behavior` enum).
+// S1. The semaphore initially has 0 permits.
```
sarsko, line 520: 1.-4. -> S1.-S4 ?. Short for step ?

tests/future/batch_semaphore.rs
```diff
@@ -512,52 +512,63 @@ fn bugged_cleanup_would_cause_deadlock() {
...
+// S2. The main task spawns a set of tasks, specifying the behavior of each, and the number of permits it should request
+// S3. Each task polls the semaphore once when it is created, in order to get added as a Waiter.
+// S4. The main task releases N semaphores which are sufficient to ensure that all tasks complete (see below)
```
sarsko, line 523: Nit: are -> is

tests/future/batch_semaphore.rs
```diff
@@ -512,52 +512,63 @@ fn bugged_cleanup_would_cause_deadlock() {
...
 //
-// At this point, T2 should be woken up and should get the permit.
-// Unfortunately, in an earlier version of BatchSemaphore, this was not happening because
-// the Drop handler for Acquire was only returning permits to the semaphore, but not
-// waking up the next waiter in the queue.
+// The general idea is that there are 3 types of tasks: `EarlyDrop`, `Hold` and `Release` tasks
+// (determined by the `Behavior` enum).
+// S1. The semaphore initially has 0 permits.
+// S2. The main task spawns a set of tasks, specifying the behavior of each, and the number of permits it should request
+// S3. Each task polls the semaphore once when it is created, in order to get added as a Waiter.
```
sarsko, line 522: Nit: "it" here refers to the semaphore, but we want it to refer to the task.

src/future/batch_semaphore.rs
jorajeev: Yeah, updated to use `match`

src/future/batch_semaphore.rs
jorajeev: Fixed

src/future/batch_semaphore.rs
jorajeev: I tried writing it abstractly, but found a concrete example was a lot easier to read.  Also, the shortest example requires weird behavior from W1 - it's woken up but then decides to drop.  I felt the example I have is more realistic (W2 gives up after waiting, e.g. with a timeout).

Also: I don't understand what you mean by `maximum number of permits`?  We can initialize a BatchSemaphore with 0 permits and then someone can release thousands of permits later on. (We use this for mpsc channels for instance.)

src/future/batch_semaphore.rs
jorajeev: Added spacing

src/future/batch_semaphore.rs
jorajeev: Oops - fixed.

tests/future/batch_semaphore.rs
jorajeev: Changed to 1,2,3...

tests/future/batch_semaphore.rs
jorajeev: Fixed

tests/future/batch_semaphore.rs
jorajeev: My favorite LLM says "are sufficient to ensure" is grammatically correct :-)

src/future/batch_semaphore.rs
sarsko: > tried writing it abstractly, but found a concrete example was a lot easier to read. 

Yes, I like the example, just more of a nit on writing "An example of this" or similar over "This can happen in the following situation".

> weird behavior from W1 - it's woken up but then decides to drop.

? I don't follow. The example is the same, but W1 wants to acquire 2 permits (with 1 available), W2 wants to acquire 1 permits, and W3 doesn't exist. Rest of the example pans out the same.

> Also: I don't understand ...

Yeah, nevermind.

sarsko (approved):
Thanks, looks good to me!

PR #172, opened 2025-03-06
Upgrade to edition 2024

sarsko:
The 2024 edition has been released. This PR upgrades to it.

I followed this guide:

https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html.

Most of the changes were automatically applied. The only ones that were not were the `let value = unsafe { &self.raw_load() };` changes I made, which are due to https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html.

One thing to note is that the `thread_local!` and `lazy_static!` macros now have `expr_2021`s, rather than `exprs`. This is correct, as std is still on 2021 (https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml) and lazy static is on 2015 (https://github.com/rust-lang-nursery/lazy-static.rs/blob/master/Cargo.toml).

I renamed a couple of places where `gen` was used as a variable name (as it is now a keyword). I'll just leave a heads up here that rand has gone away from `gen` (and all other `gen_` functions) in 0.9, in addition to a fair bit of other API changes. This is for sure going to be fun whenever people start migrating. The easiest here might be to provide both the new and the old APIs.

---


sarsko:
Closed because I can't be bothered rebasing and this is not an important PR.

PR #174, opened 2025-03-18
Make `SHUTTLE_RANDOM_SEED` always override.

sarsko:
Does two things:
1. Makes it so that if `SHUTTLE_RANDOM_SEED` is set, then that seed will be used.
2. Makes the PctScheduler also honor `SHUTTLE_RANDOM_SEED` (vs just RandomScheduler previously)

Regarding 2: We added the output seed and replay on seed in https://github.com/awslabs/shuttle/pull/161. @jorajeev originally wanted PCT as well, but this was for some reason left as is. I don't see why PCT shouldn't honor `SHUTTLE_RANDOM_SEED`.

Regarding 1: Currently it is kind of annoying and also just a bad user experience (because we say "set this env var", and then, depending on how the test is setup, might just ignore it) that the test has to use `check_random` for `SHUTTLE_RANDOM_SEED` to be honored. This is made worse by the fact that we mostly don't use `check_random`, and also usually recommend creating runners/schedulers explicitly (just because this is more versatile).

I believe the original arguments for putting it in `check_random` vs inside the constructor was that 1: one can always build a new constructor which honors the environment variable and 2: if you create a scheduler through `new_from_seed` then it seems fair to expect that you'll get a scheduler initialized with this seed. My argument against these points is that `SHUTTLE_RANDOM_SEED` doesn't get set by accident, and setting it and not getting the seed overridden is both worse and significantly more likely to occur. It also makes sense for the `new_from_seed` to be overridden because what is output is the seed used for the last execution, but what is input is the seed for the first execution. Also, `new_from_seed` mostly exists such that the seed can be generated by something like proptest, in which case you for sure will want to override the seed if a failing one is found.

---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -323,15 +323,14 @@ where
 
 /// Run the given function under a randomized concurrency scheduler for some number of iterations.
 /// Each iteration will run a (potentially) different randomized schedule.
-/// When the environment variable SHUTTLE_RANDOM_SEED is set to a u64, this number will be used
-/// as the seed to initialize the random scheduler.
 pub fn check_random<F>(f: F, iterations: usize)
 where
     F: Fn() + Send + Sync + 'static,
 {
     use crate::scheduler::RandomScheduler;
 
-    Runner::new(RandomScheduler::new(iterations), Default::default()).run(f)
+    let runner = Runner::new(RandomScheduler::new(iterations), Default::default());
+    runner.run(f);
 }
 
 /// Run function `f` using `RandomScheduler` initialized with the provided `seed` for the given
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -330,24 +330,8 @@ where
     F: Fn() + Send + Sync + 'static,
 {
     use crate::scheduler::RandomScheduler;
-    use tracing::info;
 
-    let seed_env = std::env::var("SHUTTLE_RANDOM_SEED");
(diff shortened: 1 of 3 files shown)


src/scheduler/pct.rs
```diff
@@ -42,6 +42,21 @@ impl PctScheduler {
     pub fn new_from_seed(seed: u64, max_depth: usize, max_iterations: usize) -> Self {
```
jorajeev, line 42: Need to update the comment to explain the effect of setting `SHUTTLE_RANDOM_SEED`

src/scheduler/random.rs
```diff
@@ -58,7 +58,23 @@ impl RandomScheduler {
     /// Two RandomSchedulers initialized with the same seed will make the same scheduling decisions
```
jorajeev, line 58: Need to update the comment to explain the effect of setting `SHUTTLE_RANDOM_SEED`

src/lib.rs
```diff
@@ -330,23 +330,8 @@ where
     F: Fn() + Send + Sync + 'static,
```
jorajeev, line 330: Remove the comment on lines 326-327.  (The comment should go in the `new` methods for the Random and PCT schedulers.)

jorajeev (changes requested):

jorajeev (approved):

PR #175, opened 2025-03-21
Add `reset_step_count`

sarsko:
Adds `reset_step_bound`.

Main motivation is to be able to scale up tests without having a huge step bound.

---

### DIFF
--- a/src/current.rs
+++ b/src/current.rs
@@ -95,6 +95,16 @@ pub fn me() -> TaskId {
     get_current_task().unwrap()
 }
 
+/// Sets the number of scheduling steps used (wrt. the step bound) to 0.
+///
+/// The idea behind this is to run the test with some step bound, and then call this function whenever it is known that progress has been made.
+/// This allows tests to run with tighter step bounds, and to scale a test up without also changing the step bound.
+///
+/// NOTE: Be careful when using this, as if used wrongly it can be used to make a test execute forever.
+pub fn reset_step_count() {
+    ExecutionState::with(|s| s.steps_reset_at = s.current_schedule.len());
+}
+
 /// Sets the `tag` field of the current task.
 /// Returns the `tag` which was there previously.
 #[deprecated]
--- a/src/runtime/execution.rs
+++ b/src/runtime/execution.rs
@@ -246,12 +246,14 @@ pub(crate) struct ExecutionState {
     has_yielded: bool,
     // the number of scheduling decisions made so far
     context_switches: usize,
+    // the schedule length last time `reset_stop_bound()` was called
+    pub(crate) steps_reset_at: usize,
 
     // static values for the current execution
     storage: StorageMap,
 
     scheduler: Rc<RefCell<dyn Scheduler>>,
-    pub(super) current_schedule: Schedule,
+    pub(crate) current_schedule: Schedule,
 
     in_cleanup: bool,
 
@@ -292,6 +294,7 @@ impl ExecutionState {
             next_task: ScheduledTask::None,
             has_yielded: false,
             context_switches: 0,
+            steps_reset_at: 0,
             storage: StorageMap::new(),
             scheduler,
             current_schedule: initial_schedule,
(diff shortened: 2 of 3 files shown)


jorajeev (dismissed):
No unit test?

tests/basic/execution.rs
```diff
@@ -244,3 +245,37 @@ fn failure_outside_execution() {
     let lock = shuttle::sync::Mutex::new(0u64);
     let _ = lock.lock().unwrap();
 }
+
+fn reset_step_bound(do_reset: bool) {
+    let mut config = Config::new();
+    config.max_steps = MaxSteps::FailAfter(20);
```
bkragl, line 251: Not absolutely necessary, but I'm wondering if 20 is tight here. Would be interesting to write this as a formula that has the number of threads and yields as parameters.

src/current.rs
```diff
@@ -95,6 +95,14 @@ pub fn me() -> TaskId {
     get_current_task().unwrap()
 }
 
+/// Resets the step bound.
+/// The idea behind this is to run the test with some step bound, and then call this function whenever it is known that progress has been made.
+/// This allows to have a tighter step bound, and to scale a test up without also changing the step bound.
+/// NOTE: Be careful when using this, as if used wrongly it can allow a test to execute forever.
+pub fn reset_step_bound() {
```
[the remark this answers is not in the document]
sarsko: > 1. We are resetting the count, not the bound, so I'd call this reset_step_count.

Sure.

> 2. "allows to" is grammatically incorrect in English ([bertrandmeyer.com/2018/01/16/before-i-start-screaming-once-again](https://bertrandmeyer.com/2018/01/16/before-i-start-screaming-once-again/)). As a native German speaker, I was surprised too when I learned about this.

TIL.

That said, it looks to me like my sentence was grammatically correct. [it] (the function) [can allow] [a test] [to execute forever]. Looks to me like "a test" is the "*someone* or *something*" from Prof. Meyer's article.

> 3. Blank line after summary line and before note.

Yup

tests/basic/execution.rs
sarsko: > I'm wondering if 20 is tight here?

What does tight mean? As small as possible?

We know that if you run the test without resetting step counts, then it will panic, and if you do reset, then it won't. That sounds like a good enough test to me, and any number other than 20 which satisfies that (one test failing, one test succeeding) is equivalent in its value.

> Would be interesting to write this as a formula that has the number of threads and yields as parameters.

Sounds like a waste of time.

tests/basic/execution.rs
```diff
@@ -244,3 +245,37 @@ fn failure_outside_execution() {
...
+                thread::spawn(move || {
+                    for _ in 0..3 {
+                        thread::yield_now();
+                    }
+                    if do_reset {
+                        shuttle::current::reset_step_bound();
+                    }
+                })
+            })
+            .for_each(move |jh| jh.join().unwrap())
```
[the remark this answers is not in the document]
sarsko: Nope, removed.

tests/basic/execution.rs
bkragl: Yeah, as small as possible. Basically the example shows that a user needs to think about the worst case, where all tasks are right before resetting the step count. Could have another test with `do_reset = true` and `max_steps` set to the smallest successful number - 1, which should panic.

src/runtime/execution.rs
```diff
@@ -644,14 +647,14 @@ impl ExecutionState {
         self.context_switches += 1;
 
         match self.config.max_steps {
-            MaxSteps::FailAfter(n) if self.current_schedule.len() >= n => {
+            MaxSteps::FailAfter(n) if self.current_schedule.len() - self.steps_reset_at >= n => {
```
bkragl, line 650: For readability, would be nice to extract the calculation (which also happens below) this into a function.

src/current.rs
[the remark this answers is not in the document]
sarsko: Ack fixed

src/runtime/execution.rs
sarsko: Ack done

tests/basic/execution.rs
sarsko: Ack, added

bkragl (approved):

bkragl (dismissed):

bkragl (approved):

PR #178, opened 2025-04-18
Add BatchSemaphore::{held_permits, is_queued}

sarsko:
Adds `BatchSemaphore::{held_permits, is_queued}`. The motivation is so that functionality building atop BatchSemaphore can protect against things like reentrancy or double_queueing.

---

### DIFF
--- a/shuttle-engine/Cargo.toml
+++ b/shuttle-engine/Cargo.toml
@@ -20,6 +20,8 @@ smallvec = { version = "1.11.2", features = ["const_new"] }
 tracing = { version = "0.1.36", default-features = false, features = ["std"] }
 corosensei = "0.3.1"
 const-siphasher = "1.0.2"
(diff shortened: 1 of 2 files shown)


shuttle/Cargo.toml
```diff
@@ -21,6 +21,8 @@ scoped-tls = "1.0.0"
 smallvec = { version = "1.11.2", features = ["const_new"] }
 tracing = { version = "0.1.36", default-features = false, features = ["std"] }
 corosensei = "0.3.1"
+indexmap = { version = "2.9", features = ["std"] }
```
jorajeev: Curious why we need to add a dependency on IndexMap here? Why isn't HashMap good enough?

sarsko: I did it to not have a latent footgun. `HashMap` does not have deterministic iteration order, but `IndexMap` does.

PR #180, opened 2025-05-18
Fix: make ThreadRng::default no longer give a Clippy error

sarsko:
Currently use of `ThreadRng::default` results in a Clippy error. After this change it does not.

Also fixes some Clippies

---


bkragl (commented):
two minor comments

bkragl (approved):

PR #181, opened 2025-06-07
Add thread::scope

### LINKED ISSUE #153: Implement thread::scope
Is there an equivalent to std::thread::scope? I see that Loom also doesn't implement this.

nrebei2:
Adds `thread::scope`, which allows threads to borrow non-'static data.
---

Hello, I’ve been using shuttle in my own projects and found it really useful! I prefer using thread::scope instead of thread::spawn in my code when I can, so I initially vendored Shuttle and added scoped thread support locally.

After seeing #153, I decided to fork and upstream my changes in this pull request. I've included some test cases as well.

One change which may be undesirable is the removal of `Send` bounds from task spawning and changing the functions to take a boxed closure directly. The former was due to scoped threads relying on `Rc<RefCell>` to track the number of running threads. The latter came about because I needed to box the closure early in `spawn_named` (now renamed to `spawn_named_unchecked`) to erase scoped lifetimes. I don't think removing `Send` is too much of an issue since the runtime is single-threaded, but correct me if I'm wrong. Anyways, this was the way I found to support scoped threads.

Resolves #153 

### DIFF
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -3,8 +3,9 @@
 use crate::runtime::execution::ExecutionState;
 use crate::runtime::task::TaskId;
 use crate::runtime::thread;
-use std::marker::PhantomData;
+use std::cell::RefCell;
 use std::time::Duration;
+use std::{marker::PhantomData, rc::Rc};
 
 pub use std::thread::{panicking, Result};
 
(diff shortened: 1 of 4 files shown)


src/thread.rs
```diff
@@ -60,22 +133,27 @@ where
     F: Send + 'static,
     T: Send + 'static,
 {
-    spawn_named(f, None, None)
+    // Safety: F is static so there are no captured references
+    unsafe {spawn_named_unchecked(f, None, None)}
```
sarsko: I think `spawn_named` should be reinstated (with the same signature as before) as a thin wrapper around `spawn_named_unchecked`. 

I also prefer SAFETY over Safety for safety comments.

src/thread.rs
```diff
@@ -50,6 +51,78 @@ impl Thread {
     }
 }
 
+/// A scope to spawn scoped threads in.
+///
+/// See [`scope`] for details.
+#[derive(Debug)]
```
sarsko: Should be swapped to the same impl as in `std`: https://doc.rust-lang.org/src/std/thread/scoped.rs.html#334-342.

src/thread.rs
```diff
@@ -120,22 +196,51 @@ where
...
+/// See [`Scope::spawn`] for details.
+#[derive(Debug)]
+pub struct ScopedJoinHandle<'scope, T> {
+    handle: JoinHandle<T>,
+    _marker: PhantomData<&'scope T>,
+}
+
+#[derive(Clone, Debug)]
+struct ScopeData {
+    num_running_threads: Rc<RefCell<usize>>,
```
sarsko: This cannot be `Rc<RefCell<usize>>` as `ScopeData` is a field in `Scope`, and `Scope` is supposed to be `Send` and `Sync`. You could of course just implement them, but I don't see why this cannot just be an `Arc<AtomicUsize>` ?

src/thread.rs
```diff
@@ -120,22 +196,51 @@ where
...
+    }
+}
+
 /// An owned permission to join on a thread (block on its termination).
 #[derive(Debug)]
 pub struct JoinHandle<T> {
     task_id: TaskId,
     thread: Thread,
-    result: std::sync::Arc<std::sync::Mutex<Option<Result<T>>>>,
+    result: Rc<RefCell<Option<Result<T>>>>,
```
sarsko: Why is this and the others like it being swapped to use `Rc<RefCell>`?

tests/basic/thread.rs
```diff
@@ -35,6 +35,80 @@ fn thread_yield_point() {
     assert_eq!(success_clone.load(Ordering::SeqCst), 0x3);
 }
 
+#[test]
+fn thread_scope() {
+    check_random(
```
sarsko: This should be `check_dfs`. I think the same applies to the two tests below as well.

src/thread.rs
```diff
@@ -50,6 +51,78 @@ impl Thread {
...
+            let mut num_threads = scope_data.num_running_threads.borrow_mut();
+            *num_threads -= 1;
+            if *num_threads == 0 {
+                ExecutionState::with(|s| s.get_mut(scope_data.main_task).unblock());
+            }
+
+            ret
+        };
+
+        // Safety: main task is blocked until all scoped closures complete so all captured references remain valid
```
sarsko: Safety -> SAFETY

src/thread.rs
```diff
@@ -257,7 +362,8 @@ impl Builder {
         F: Send + 'static,
         T: Send + 'static,
     {
-        Ok(spawn_named(f, self.name, self.stack_size))
+        // Safety: F is static so there are no captured references
```
sarsko: Safety -> SAFETY

src/thread.rs
```diff
@@ -120,22 +196,51 @@ where
...
+    /// Waits for the associated thread to finish.
+    pub fn join(self) -> Result<T> {
+        self.handle.join()
+    }
+
+    /// Extracts a handle to the underlying thread.
+    pub fn thread(&self) -> &Thread {
+        self.handle.thread()
+    }
+}
```
sarsko, line 256: Might as well add `is_finished` as well, no?

src/thread.rs
```diff
@@ -120,22 +196,51 @@ where
...
+pub struct ScopedJoinHandle<'scope, T> {
+    handle: JoinHandle<T>,
+    _marker: PhantomData<&'scope T>,
+}
+
+#[derive(Clone, Debug)]
+struct ScopeData {
+    num_running_threads: Rc<RefCell<usize>>,
+    main_task: TaskId,
+}
```
sarsko: I don't see the need for `ScopeData` at all, why cannot the fields just be inlined into `Scope`?

src/thread.rs
```diff
@@ -120,22 +196,51 @@ where
...
 /// An owned permission to join on a thread (block on its termination).
 #[derive(Debug)]
 pub struct JoinHandle<T> {
     task_id: TaskId,
     thread: Thread,
-    result: std::sync::Arc<std::sync::Mutex<Option<Result<T>>>>,
+    result: Rc<RefCell<Option<Result<T>>>>,
 }
 
+unsafe impl<T: Send> Send for JoinHandle<T> {}
```
sarsko: `JoinHandle` is `Send` and `Sync` (and is so regardless of `T`). I see now that we currently have done this incorrectly, put up a PR: https://github.com/awslabs/shuttle/pull/183.

sarsko (changes requested):
Thanks for the PR, much appreciated! I left some comments.

Regarding what you notes in the description: your belief about `Send` is correct. I think taking a boxed closure is perfectly fine (actually probably reduces the size of the final binary for around 0 cost), and I also don't immediately see a way of not doing it. Not sure I am a fan of the `Rc`-ification though, don't really see the point of it.

sarsko (approved):
Looks good to me, thanks!

jorajeev (approved):
Thanks for the contribution @nrebei2!

PR #187, opened 2025-07-28
Vector Clock Feature Flag

dylanjwolff:
This PR adds a feature flag for vector clocks to boost Shuttle's testing throughput when they are not needed.

This feature flag is *disabled* by default in production, but *always enabled* for tests because many tests leverage vector clocks for assertions to check correctness. In the micro-benchmarks for the Shuttle repository, this results in performance gains of 5-25%:

[criterion.tar.gz](https://github.com/user-attachments/files/21477778/criterion.tar.gz)

When disabled, the VectorClock struct is replaced by an empty struct for which all operations are i.e. a no-op.

---

### DIFF
--- a/shuttle/Cargo.toml
+++ b/shuttle/Cargo.toml
@@ -38,12 +38,16 @@ test-log = { version = "0.2.8", default-features = false, features = ["trace"] }
 tracing-subscriber = { version = "0.3.9", features = ["env-filter"] }
 trybuild = "1.0"
 pin-project = "1.1.3"
+# The following line is necessary to ensure vector-clocks are used for integration tests
+# To run performance tests without vector clocks using `cargo bench`, this line must be commented out
+shuttle = { path = ".", features = ["vector-clocks"] }
 
 [lib]
 bench = false
 
 [features]
 default = []
+vector-clocks = []
 annotation = ["dep:serde", "dep:serde_json", "dep:regex"]
 
 [[bench]]
--- a/shuttle/src/runtime/task/clock.rs
+++ b/shuttle/src/runtime/task/clock.rs
@@ -2,7 +2,7 @@ use crate::runtime::task::TaskId;
 use std::cmp::{Ordering, PartialOrd};
 
 #[cfg(any(test, feature = "vector-clocks"))]
-mod vc_enabled {
+mod vector_clock {
     use super::*;
     use crate::runtime::task::DEFAULT_INLINE_TASKS;
     use smallvec::{smallvec, SmallVec};
@@ -100,7 +100,7 @@ mod vc_enabled {
 /// A dummy VectorClock implementation which only provides no-op stubs to improve testing throughput when
 /// vector clocks are not necessary
 #[cfg(not(any(test, feature = "vector-clocks")))]
-mod vc_disabled {
+mod vector_clock {
     use super::*;
 
     #[derive(Clone, Debug, PartialEq, Eq)]
@@ -154,11 +154,7 @@ mod vc_disabled {
     }
 }
 
-#[cfg(any(test, feature = "vector-clocks"))]
-pub use vc_enabled::VectorClock;
-
-#[cfg(not(any(test, feature = "vector-clocks")))]
-pub use vc_disabled::VectorClock;
+pub use vector_clock::VectorClock;
 
 #[cfg(test)]
 mod test {
--- a/shuttle/src/runtime/task/clock.rs
+++ b/shuttle/src/runtime/task/clock.rs
@@ -1,95 +1,164 @@
-use crate::runtime::task::{TaskId, DEFAULT_INLINE_TASKS};
-use smallvec::{smallvec, SmallVec};
+use crate::runtime::task::TaskId;
 use std::cmp::{Ordering, PartialOrd};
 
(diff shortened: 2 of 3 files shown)


shuttle/src/runtime/task/clock.rs
```diff
@@ -1,12 +1,20 @@
...
+    crate::runtime::task::DEFAULT_INLINE_TASKS,
+    smallvec::{smallvec, SmallVec},
+};
+use {
+    crate::runtime::task::TaskId,
+    std::cmp::{Ordering, PartialOrd},
+};
+
+#[cfg(any(test, feature = "vector-clocks"))]
 #[derive(Clone, Debug, PartialEq, Eq)]
```
sarsko: How about `mod`ing all of this and having only one `cfg`?

shuttle/src/runtime/task/clock.rs
```diff
@@ -91,6 +104,64 @@ impl PartialOrd for VectorClock {
     }
 }
 
+/// A dummy VectorClock implementation which only provides no-op stubs to improve testing throughput when
+/// vector clocks are not necessary
+#[cfg(all(not(test), not(feature = "vector-clocks")))]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct VectorClock;
```
sarsko: My suggestion is to remove this and `cfg` all usages of `VectorClock`

shuttle/src/runtime/task/clock.rs
dylanjwolff: done, and I agree it makes the code a bit cleaner but fair warning that it also makes the diffs much harder to read

shuttle/src/runtime/task/clock.rs
dylanjwolff: it would be nice to not need a dummy implementation, but the issue with doing this is that there are other structs that have `VectorClock`'s (e.g. https://github.com/awslabs/shuttle/blob/41ae7ba0c589fbc2424bab602f10a6a723db87c1/shuttle/src/future/batch_semaphore.rs#L69) so the blast radius of this change is quite large and would result in the codebase being littered with `#[cfg...]`'s.

the dummy implementation keeps this PR quite small and gives us probably 99% of the performance benefits

shuttle/src/runtime/task/clock.rs
```diff
@@ -1,96 +1,165 @@
...
+            Some(Ordering::Equal)
+        }
     }
 }
 
+#[cfg(any(test, feature = "vector-clocks"))]
+pub use vc_enabled::VectorClock;
+
+#[cfg(not(any(test, feature = "vector-clocks")))]
+pub use vc_disabled::VectorClock;
```
sarsko: Instead of calling the mods `vc_enabled` and `vc_disabled`, just mod them under `vector_clock` and `pub use vector_clock`.

shuttle/Cargo.toml
```diff
@@ -38,12 +38,16 @@ test-log = { version = "0.2.8", default-features = false, features = ["trace"] }
 tracing-subscriber = { version = "0.3.9", features = ["env-filter"] }
 trybuild = "1.0"
 pin-project = "1.1.3"
+# The following line is necessary to ensure vector-clocks are used for integration tests
+# To run performance tests without vector clocks using `cargo bench`, this line must be commented out
```
sarsko, line 42: It is possible to do benching without vector clocks by introducing a feature (eg. `bench-no-vector-clocks`) and then gating vector clocks on `all(any(test, feature = "vector-clocks"), not(feature = "bench-no-vector-clocks")`

shuttle/src/runtime/task/clock.rs
dylanjwolff: whoops! yes, that is much better

shuttle/Cargo.toml
dylanjwolff: Yup, this would allow you to run the benchmarks without vector clocks by enabling the feature at the command line rather than commenting this line out, which I agree is better. I can add it, but I am getting worried that this setup is a little counter-intuitive (feature is disabled by default, but *always* enabled for tests and benchmarks *unless* you use this other feature flag).

I'm starting to think it might be better to enable vector clocks by default, add it as a "required feature" for tests, and add a gated compile error to prevent ever using `cargo test` without the feature. I think this is slightly more intuitive: if the feature is enabled its enabled everywhere and if it is disabled, it is disabled everywhere (but should result in compile errors for testing). For example:

https://github.com/awslabs/shuttle/commit/f9d5ced1b1f0097e951edd8e5106e3decd2d9fa9

Sorry for the back and forth on this!

sarsko: I think default enabling vector clocks (or most features tbh) is a mistake:
1. It is a "poweruser" feature with runtime performance costs.
2. The default offerings should be sensible. This is not a sensible default (noone has ever used it). 
3. The standard recommendation will be "oh and yeah, disable this feature" which is just bad UX.
4. Feature enablement is additive. Having it as a default makes accidental opt-in the standard, which is a pain for projects which span multiple packages.
5. One should opt-in to the functionality one wants, not opt-out of the functionality one does not want.

sarsko (approved):

PR #188, opened 2025-07-30
Fix: Preserve `Once` state across moves

lukenels:
`sync::Once` is a synchronization primitive to execute an initialization routine exactly once in the case of multiple threads. There's currently an issue in Shuttle's implementation of the `Once` primitive that gets triggered when a `once` is moved to a new location:

```rust
fn bar() {
    // Allocate a Once, initially is_completed = false
    let once = Box::new(Once::new());
    assert!(!once.is_completed());

    // Invoke call_once, now is_completed = true
    once.call_once(|| {});
    assert!(once.is_completed());

    // Move the Once to a new location, is_completed should still be true.
    // However, Shuttle's version resets back to is_completed = false
    // and this assertion fails.
    let once = Box::new(*once);
    assert!(once.is_completed());
}
```

The problem is that Shuttle's implementation of `Once` uses the _address_ of the `Once` as a unique key to look up the `Once`'s state in ExecutionState storage. But when the `Once` is moved, it gets a new address, and so it effectively becomes an entirely new object from Shuttle's perspective. For static uses of `Once`, this will never matter as the object lives for the duration of the program and cannot be moved from. But `Once`s need not be static, as shown above.

This commit introduces a new unit test case demonstrating the problem, and fixes the problem by adding a layer of indirection: instead of using the address of the `Once` itself as the storage key, the `Once` will (lazily) allocate a dummy `Box<usize>` and use the address of the allocation as the key. This fixes the problem because the inner address (and therefore, the storage key) is preserved even if the `Once` itself is moved.

---

### DIFF
--- a/shuttle/src/sync/once.rs
+++ b/shuttle/src/sync/once.rs
@@ -2,8 +2,9 @@ use crate::runtime::execution::ExecutionState;
 use crate::runtime::storage::StorageKey;
 use crate::runtime::task::clock::VectorClock;
 use crate::sync::Mutex;
-use std::cell::RefCell;
+use std::cell::{OnceCell, RefCell};
 use std::rc::Rc;
+use std::sync::atomic::{AtomicUsize as StdAtomicUsize, Ordering};
 use tracing::trace;
 
 /// A synchronization primitive which can be used to run a one-time global initialization. Useful
@@ -11,9 +12,8 @@ use tracing::trace;
 /// with [`Once::new()`].
 #[derive(Debug)]
 pub struct Once {
-    // We use the address of the `Once` as an identifier, so it can't be zero-sized even though all
-    // its state is stored in ExecutionState storage
-    _dummy: usize,
+    /// Unique identifier for this [`Once`], used as a key for the [`OnceInitState`] for this instance.
+    id: OnceCell<usize>,
 }
 
 /// A `Once` cell can either be `Running`, in which case a `Mutex` mediates racing threads trying to
@@ -33,12 +33,19 @@ impl std::fmt::Debug for OnceInitState {
     }
 }
 
+// Safety: Once is never actually passed across true threads, only across continuations. The
+// OnceCell type therefore can't be preempted mid-bookkeeping-operation.
+// TODO we shouldn't need to do this, but OnceCell is not Send, and anything we put within a Once
+// TODO needs to be Send.
+unsafe impl Send for Once {}
+unsafe impl Sync for Once {}
+
 impl Once {
     /// Creates a new `Once` value.
     #[must_use]
     #[allow(clippy::new_without_default)]
     pub const fn new() -> Self {
-        Self { _dummy: 0 }
+        Self { id: OnceCell::new() }
     }
 
     /// Performs an initialization routine once and only once. The given closure will be executed
@@ -142,6 +149,14 @@ impl Once {
         }
     }
 
+    fn id(&self) -> usize {
+        static NEXT_ID: StdAtomicUsize = StdAtomicUsize::new(1);
+
+        let id = *self.id.get_or_init(|| NEXT_ID.fetch_add(1, Ordering::Relaxed));
+        assert_ne!(id, 0, "id overflow");
+        id
+    }
+
     fn get_state<'a>(&self, from: &'a ExecutionState) -> Option<&'a RefCell<OnceInitState>> {
         from.get_storage::<_, RefCell<OnceInitState>>(self)
     }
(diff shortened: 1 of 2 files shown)


shuttle/src/sync/once.rs
```diff
@@ -11,9 +11,11 @@ use tracing::trace;
...
 #[derive(Debug)]
 pub struct Once {
-    // We use the address of the `Once` as an identifier, so it can't be zero-sized even though all
-    // its state is stored in ExecutionState storage
-    _dummy: usize,
+    // We need a unique identifier for this `Once` to look up its state in ExecutionState storage.
+    // We can't use the address of the `Once`, because it might be moved. Instead we'll (lazily)
+    // allocate a `Box<usize>` and use that pointer address to identify this once (the inner type
+    // doesn't actually matter, as long as it isn't zero-sized so we can ensure its address is unique.)
+    dummy: OnceCell<Box<usize>>,
```
sarsko: What am I missing? Why can't this just be a monotonically increasing `usize` (ie. `dummy` becomes `id`)?

lukenels: I considered that; I could do something like this:

```rust
fn id(&self) -> usize {
    static NEXT_ID: StdAtomicUsize = StdAtomicUsize::new(1);

    let id = *self.id.get_or_init(|| NEXT_ID.fetch_add(1, Ordering::Relaxed));
    assert_ne!(id, 0, "id overflow");
    id
}
```

The main reason I didn't is that I don't want to have to reason about the overflow case, using `Box` abuses the allocator to give a way to release identifiers when dropped. On platforms where usize is 64-bit, seems reasonable to assume it'll never overflow, but for 32-bit targets it seems possible.

sarsko: I'm not sure I share your worry; if someone is able to to allocate over 4 billion `Once`s in a single Shuttle test then I'll be mightily impressed.

If you really think this can happen, then surely an `AtomicU64` would be suited.

lukenels: I would use `AtomicU64` for the counter but `StorageKey` uses `usize` anyways so it wouldn't help. I agree it's unlikely to exhaust a the counter for 32-bit targets, but thought it's simpler to not have to make that argument at all. I'm happy to switch to the counter-based implementation though, I doubt there are many non-static `Once`s in the wild, and we'd have to run a lot of test case iterations in order to exhaust the id space.

shuttle/src/sync/once.rs
```diff
@@ -2,18 +2,18 @@ use crate::runtime::execution::ExecutionState;
...
 /// A synchronization primitive which can be used to run a one-time global initialization. Useful
 /// for one-time initialization for FFI or related functionality. This type can only be constructed
 /// with [`Once::new()`].
 #[derive(Debug)]
 pub struct Once {
-    // We use the address of the `Once` as an identifier, so it can't be zero-sized even though all
-    // its state is stored in ExecutionState storage
-    _dummy: usize,
+    /// Unique identifier for this [`Once`], used as a key for the [`OnceInitState`] for this instance.
+    id: OnceCell<usize>,
```
sarsko, line 16: Why do we need a `OnceCell` around the `usize`?

shuttle/tests/basic/once.rs
```diff
@@ -249,3 +249,25 @@ fn poison() {
...
+{
+    checker(Box::new(move || {
+        let once = Box::new(Once::new());
+        assert!(!once.is_completed());
+        once.call_once(|| {});
+        assert!(once.is_completed());
+
+        // The `Once` is still completed, even if its moved to another address.
+        let once = Box::new(*once);
+        assert!(once.is_completed());
```
sarsko, line 266: Might be pedantic, but should we create a second `Once` and assert that it is not completed (ie that `completed` does not apply to all `Once`s?)

shuttle/src/sync/once.rs
sarsko: Yeah, I getcha. I feel like an id-solution is less "foreign" and feels less like a trick, which I value more than theoretical overflow possibilities on 32 bit architectures.

I wonder if it is worth it to change from a static counter to a per-execution counter (a way to do it would be to add a call to `ExecutionState::cleanup` where we reset static counters), but that might be more hassle than it is worth.

shuttle/src/sync/once.rs
sarsko: Nevermind on this, we need it because `new` is `const` so we have to lazily get the `usize`.

shuttle/tests/basic/once.rs
lukenels: Seems reasonable, will update with that addition.

shuttle/src/sync/once.rs
lukenels: Yeah exactly, there's no good way to allocate unique identifiers at compile time in a `const` function yet as far as I can tell.

PR #189, opened 2025-08-01
Better Shuttle Benchmarks

dylanjwolff:
This PR improves the benchmarking suite of Shuttle.

(A) The benchmarks can now be run without vector-clock overhead by using the `bench-no-vector-clocks` feature. This feature is now enabled in the benchmarking Github action.

(B) The benchmarks have also been expanded and parameterized by the number of tasks and number of total events across all threads. More specifically:

1. **Creation and Startup:**`create.rs` has been added to give a rough gauge of thread creation time and Shuttle startup overhead
2. **Parameterization:** The existing benchmarks in `counter.rs` and `lock.rs` have been parameterized and set to run in a "narrow" [5 tasks] and "wide" [100 tasks] configuration. The number of total events/operations remains constant [10000 events] across these configurations (the number of events per task is computed from the tasks and the total number of events).
3. **Scaling** An additional set of benchmark groups has been added for showing scaling as the number of tasks and events increase. The sampling rate and warmup time has been greatly reduced for these to allow for the many new configurations (`3*5*2=30`) to be run in a reasonable amount of time. As such, statistical comparisons on individual data-points from these scaling benchmarks are not particularly informative. Unfortunately, Criterion doesn't have an automatic way to visualize trends, so the reports generated for these are not easy to interpret. In the future, it might be useful to generate custom reports from the raw data instead.

---


.github/workflows/tests.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Install Rust
         run: rustup update stable
       - name: cargo bench
-        run: cargo bench
+        run: cargo bench --features bench-no-vector-clocks
```
sarsko, line 57: Do we only wanna run without vector clocks?

sarsko:
Do we want an "even wider" bench (eg 1000 or 10000 tasks)? 100 is just not that many.

sarsko:
Just noting that `bench-no-vector-clocks` is fine by me (rumor has it that I suggested it), but it is also kinda bad style. Features are supposed to be additive, and this breaks that. We should have the feature separated from the rest, with a comment explaining, and also comment that this feature should never ever exist in a `Cargo.TOML` (or other config file), ie. it should only be invoked as we do here, as a part of a `cargo ...` command.

.github/workflows/tests.yml
dylanjwolff: I'm worried that the vector clock overhead will obscure the impact that other changes have on performance

sarsko: Yes, the question is if we wanna run with and without

dylanjwolff: how often is the vector clock code changing? we can always enable benchmarking for the vector-clock version if  there start being more changes to the vector clock implementations or how they are used. right now I suspect their usage and implementations aren't changing much, so probably it's not informative to benchmark them

sarsko: Yeah, no, it's mostly because I like being able to give an answer of the sort "the overhead of enabling them is X" instead of "yeah, we ran some benchmarks a couple of years ago, and I think it was around Y slower". 

We could put it as an event on whenever the `CHANGELOG` is changed (which happens whenever we cut a release)

dylanjwolff: CHANGELOG seems like a good option then

shuttle/src/sync/once.rs
sarsko, line 1: Looks like https://github.com/awslabs/shuttle/pull/188 is part of the diff as well.

sarsko: Ie that part shouldn't be in your commit and the actual change should be rebased wrt main

PR #191, opened 2025-08-05
Persistent Vec to avoid alloc'ing runnable tasks on each schedule

dylanjwolff:
This PR changes the internal implementation of the `schedule` function to improve performance without changing functionality. Instead of creating a new `SmallVec` on each call to schedule with the runnable tasks, we use a single `Vec` whose capacity is reused on each call. As a result we only have `O(logN)` allocations where `N` is the maximum number of runnable tasks. Notably there is some limited use of `unsafe` to avoid lifetime issues with references being iteratively added and removed to the persistent vector at each scheduling point.

There is also some other minor cleanup of `schedule` rolled into this change to avoid iterating over the set of tasks more than once etc.

Using the micro-benchmarks added in #189 we see performance improvements of ~10-20% for the scheduling heavy workloads (`counter` and `lock`):

[criterion-persistent-vec.tar.gz](https://github.com/user-attachments/files/21606388/criterion-persistent-vec.tar.gz)

On a more realistic private benchmark, the average improvement is ~10%

---

### DIFF
--- a/shuttle/src/runtime/execution.rs
+++ b/shuttle/src/runtime/execution.rs
@@ -262,6 +262,10 @@ pub(crate) struct ExecutionState {
 
     // The `Span` which the `ExecutionState` was created under. Will be the parent of all `Task` `Span`s
     pub(crate) top_level_span: Span,
+
+    // Persistent Vec used as a bump allocator for references to runnable tasks to avoid slow allocation
+    // on each scheduling decision. Should not be used outside of the `schedule` function
+    runnable_tasks: Vec<*const Task>,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
@@ -302,6 +306,7 @@ impl ExecutionState {
             #[cfg(debug_assertions)]
             has_cleaned_up: false,
             top_level_span: tracing::Span::current(),
+            runnable_tasks: Vec::with_capacity(DEFAULT_INLINE_TASKS),
         }
     }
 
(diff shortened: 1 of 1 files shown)


shuttle/src/runtime/execution.rs
```diff
@@ -262,6 +262,10 @@ pub(crate) struct ExecutionState {
 
     // The `Span` which the `ExecutionState` was created under. Will be the parent of all `Task` `Span`s
     pub(crate) top_level_span: Span,
+
+    // Persistent Vec used as a i.e. bump allocator for references to runnable tasks to avoid slow allocation
```
sarsko: Remove `i.e.`

shuttle/src/runtime/execution.rs
```diff
@@ -686,21 +699,38 @@ impl ExecutionState {
...
         let is_yielding = std::mem::replace(&mut self.has_yielded, false);
 
-        let runnable_tasks = runnable
-            .iter()
-            .map(|id| self.tasks.get(id.0).unwrap())
-            .collect::<SmallVec<[&Task; DEFAULT_INLINE_TASKS]>>();
+        // Cast the slice of raw pointers to a slice of references in place to provide schedulers with a safe API
+        // This is safe because the tasks themselves are only being accessed through this shared reference by the
+        // schedulers, and all references are always cleared from the runnable_tasks Vec at the end of this function.
+        let task_refs = unsafe { std::mem::transmute::<&[*const Task], &[&Task]>(&self.runnable_tasks) };
```
sarsko, line 710: Needs `SAFETY: ` comment

shuttle/src/runtime/execution.rs
```diff
@@ -686,21 +699,38 @@ impl ExecutionState {
...
+        //
+        // Note also that changing this trace! statement requires changing the test `basic::labels::test_tracing_with_label_fn`
+        // which relies on this trace reporting the `runnable` tasks.
+        self.top_level_span.in_scope(|| {
+            trace!(
+                i=self.current_schedule.len(),
+                next_task=?self.next_task,
+                runnable=?task_refs.iter()
+                    .filter(|t| t.runnable() || matches!(self.next_task.id(), Some(id) if id == t.id))
+                    .map(|t|t.id).collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>()
```
sarsko: This looks expensive. Can't we just log runnable directly?

sarsko:
Thanks for the PR. As discussed offline: I think I am fine with merging this (post fixing what I've commented), as it is an improvement, but I think the good solution is to keep a persistent runnable vector, and then update that on calls to `Task::sleep`, `Task::unblock`, `Task::block`.

shuttle/src/runtime/execution.rs
dylanjwolff: I removed the filter since we care about *both* runnable and spuriously wakeable tasks, not just the former

I think whether we collect to a vec or a string doesn't really matter here, so I'm leaving as is

shuttle/src/runtime/execution.rs
```diff
@@ -663,20 +668,28 @@ impl ExecutionState {
...
         let mut unfinished_attached = false;
-        let mut runnable = self
-            .tasks
-            .iter()
-            .inspect(|t| unfinished_attached = unfinished_attached || (!t.finished() && !t.detached))
-            .filter(|t| t.runnable())
-            .map(|t| t.id)
-            .collect::<SmallVec<[_; DEFAULT_INLINE_TASKS]>>();
+        let mut all_runnable_detached = true;
+        let mut has_runnable = false;
```
jorajeev: nit: this should be `some_runnable` or `any_runnable` (not sure what the `has` refers to)

shuttle/src/runtime/execution.rs
```diff
@@ -686,21 +699,39 @@ impl ExecutionState {
         // they won't contribute to the check on `runnable.is_empty()` if the only runnable tasks
```
jorajeev: Comment on lines 697-701 is now out of place

shuttle/src/runtime/execution.rs
```diff
@@ -686,21 +699,39 @@ impl ExecutionState {
...
-        runnable.extend(self.tasks.iter().filter(|t| t.can_spuriously_wakeup()).map(|t| t.id));
 
         let is_yielding = std::mem::replace(&mut self.has_yielded, false);
 
-        let runnable_tasks = runnable
-            .iter()
-            .map(|id| self.tasks.get(id.0).unwrap())
-            .collect::<SmallVec<[&Task; DEFAULT_INLINE_TASKS]>>();
+        // Cast the slice of raw pointers to a slice of references in place to provide schedulers with a safe API
+        // SAFETY: This is safe because the tasks themselves are only being accessed through this shared reference by the
```
jorajeev, line 706: nit: newline before SAFETY so it's more visible

shuttle/src/runtime/execution.rs
```diff
@@ -713,16 +744,7 @@ impl ExecutionState {
...
-        // 1) It is an action taken by the scheduler, and should thus be traced under the scheduler's span
-        // 2) It creates a visual separation of scheduling decisions and `Task`-induced tracing.
-        // Note that there is a case to be made for not `in_scope`-ing it, as that makes seeing the context
-        // of the context switch clearer.
-        //
-        // Note also that changing this trace! statement requires changing the test `basic::labels::test_tracing_with_label_fn`
-        // which relies on this trace reporting the `runnable` tasks.
-        self.top_level_span
-            .in_scope(|| trace!(i=self.current_schedule.len(), next_task=?self.next_task, ?runnable));
+        self.runnable_tasks.clear();
```
jorajeev, line 748: May be worth adding a comment here that this doesn't affect the capacity

jorajeev (dismissed):

PR #195, opened 2025-08-09
Adding GH workflow to automatically compare performance of benchmarks on PR

dylanjwolff:
This PR adds a Github workflow which runs `cargo bench --features bench-no-vector-clocks` on each PR for (a) the branch being PR'ed and (b) the main branch of Shuttle. The [Github action](https://github.com/boa-dev/criterion-compare-action) comments on the PR with the results of the comparison. As the benchmarks are run in CI, the numbers should be taken with a grain of salt, but they can give a quick, broad intuition for the performance effects of changes which can be investigated further if necessary.

Notably, there is a bug in the open-source action that is doing this (https://github.com/boa-dev/criterion-compare-action/issues/22) which causes throughput to be displayed incorrectly. Because throughput and latency are essentially the same in these closed, fixed workload benchmarks (`tput = constant workload / latency`) I have removed throughput reporting from the benchmarks so that the comparison renders properly.

See https://github.com/dylanjwolff/shuttle/pull/5 for an example of what this will look like.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
     name: Tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Rust
         run: rustup update stable
       - name: cargo test
@@ -26,7 +26,7 @@ jobs:
     name: rustfmt
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Rust
         run: rustup update stable
       - name: Install rustfmt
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -46,22 +46,25 @@ jobs:
       - name: clippy
         run: cargo clippy --all-targets -- -D clippy::all
 
-  bench:
-    name: Benchmarks
+  docs:
+    name: Docs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
```
jorajeev: Should we just upgrade to @v3 everywhere?

dylanjwolff: sure! I'll give it a try

jorajeev (approved):

PR #196, opened 2025-08-09
Adding stable identifiers to tasks across Shuttle test iterations

dylanjwolff:
This PR adds `TaskSignature`s to the `Task` struct in Shuttle to provide identifiers to tasks that don't change on each iteration of a Shuttle test. These identifiers can be used for debugging with multiple different schedules, for scheduling algorithms to denote specific tasks, and also are a prerequisite for stable *resource identifiers* across Shuttle test iterations.

A signature is intended to distinguish tasks with different behaviors, but is not guaranteed to be unique. To increase the flexibility of the signatures, they provide three methods to distinguish tasks at three levels of granularity: (1) static creation point, (2) dynamic single-task creation context, and (3) dynamic concurrency context. Currently these are instantiated with (1) the location given by the `track_caller` macro (2) a per-task counter that increments when multiple tasks are spawned from the same static location in a single parent task and (3) the task signature of the parent thread. Each corresponding method presents a u64, ostensibly a hash, so that the implementations of these components can change in the future. There is also a precomputed signature hash which combines (1) (2) and (3) for quick comparisons at runtime during scheduling.

I put together some basic tests using a Subscriber to check that it is working as expected. It's easy enough to add more if we want to be more comprehensive.

---

### DIFF
--- a/shuttle/src/runtime/execution.rs
+++ b/shuttle/src/runtime/execution.rs
@@ -2,7 +2,7 @@ use crate::runtime::failure::{init_panic_hook, persist_failure, persist_task_fai
 use crate::runtime::storage::{StorageKey, StorageMap};
 use crate::runtime::task::clock::VectorClock;
 use crate::runtime::task::labels::Labels;
-use crate::runtime::task::{ChildLabelFn, Task, TaskId, TaskName, DEFAULT_INLINE_TASKS};
+use crate::runtime::task::{ChildLabelFn, Task, TaskId, TaskName, TaskSignature, DEFAULT_INLINE_TASKS};
 use crate::runtime::thread::continuation::PooledContinuation;
 use crate::scheduler::{Schedule, Scheduler};
 use crate::thread::thread_fn;
@@ -14,7 +14,7 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
-use std::panic;
+use std::panic::{self, Location};
 use std::rc::Rc;
 use std::sync::Arc;
 use tracing::{trace, Span};
@@ -65,7 +65,7 @@ impl Execution {
     /// Run a function to be tested, taking control of scheduling it and any tasks it might spawn.
     /// This function runs until `f` and all tasks spawned by `f` have terminated, or until the
     /// scheduler returns `None`, indicating the execution should not be explored any further.
-    pub(crate) fn run<F>(mut self, config: &Config, f: F)
+    pub(crate) fn run<F>(mut self, config: &Config, f: F, caller: &'static Location<'static>)
     where
         F: FnOnce() + Send + 'static,
     {
@@ -79,11 +79,10 @@ impl Execution {
 
         EXECUTION_STATE.set(&state, move || {
             // Spawn `f` as the first task
-            ExecutionState::spawn_thread(
+            ExecutionState::spawn_main_thread(
                 Box::new(move || thread_fn(f, Default::default())),
                 config.stack_size,
-                Some("main-thread".to_string()),
-                Some(VectorClock::new()),
+                caller,
             );
 
             // Run the test to completion
@@ -373,9 +372,57 @@ impl ExecutionState {
         });
     }
 
+    // Note: `spawn_thread`, `spawn_main_thread`, and `spawn_future` share some similar logic.
+    // Changes to one of these functions likely need to be propagated to the other two as well.
+    pub(crate) fn spawn_main_thread(
+        f: Box<dyn FnOnce() + 'static>,
+        stack_size: usize,
+        caller: &'static Location<'static>,
+    ) -> TaskId {
+        let name = "main-thread".to_string();
+        let mut clock = VectorClock::new();
+
+        let task_id = Self::with(|state| {
+            let parent_span_id = state.top_level_span.id();
+            let task_id = TaskId(state.tasks.len());
+            let tag = state.get_tag_or_default_for_current_task();
+
+            Self::set_labels_for_new_task(state, task_id, Some(name.clone()));
+
+            clock.extend(task_id); // and extend it with an entry for the new thread
+
+            let schedule_len = state.current_schedule.len();
+
+            let task = Task::from_closure(
+                f,
+                stack_size,
+                task_id,
+                Some(name),
+                clock,
+                parent_span_id,
+                schedule_len,
+                tag,
+                None,
+                TaskSignature::new_parentless(caller),
+            );
+            state.tasks.push(task);
+
+            task_id
+        });
+        crate::annotations::record_task_created(task_id, false);
+        task_id
+    }
+
+    // Note: `spawn_thread`, `spawn_main_thread`, and `spawn_future` share some similar logic.
+    // Changes to one of these functions likely need to be propagated to the other two as well.
     /// Spawn a new task for a future. This doesn't create a yield point; the caller should do that
     /// if it wants to give the new task a chance to run immediately.
-    pub(crate) fn spawn_future<F>(future: F, stack_size: usize, name: Option<String>) -> TaskId
+    pub(crate) fn spawn_future<F>(
+        future: F,
+        stack_size: usize,
+        name: Option<String>,
+        caller: &'static Location<'static>,
+    ) -> TaskId
     where
         F: Future<Output = ()> + 'static,
     {
@@ -400,7 +447,8 @@ impl ExecutionState {
                 parent_span_id,
                 schedule_len,
                 tag,
-                state.try_current().map(|t| t.id()),
+                Some(state.current().id()),
+                state.current_mut().signature.new_child(caller),
             );
 
             state.tasks.push(task);
@@ -411,11 +459,14 @@ impl ExecutionState {
         task_id
     }
 
+    // Note: `spawn_thread`, `spawn_main_thread`, and `spawn_future` share some similar logic.
+    // Changes to one of these functions likely need to be propagated to the other two as well.
     pub(crate) fn spawn_thread(
         f: Box<dyn FnOnce() + 'static>,
         stack_size: usize,
         name: Option<String>,
         mut initial_clock: Option<VectorClock>,
+        caller: &'static Location<'static>,
     ) -> TaskId {
         let task_id = Self::with(|state| {
             let parent_span_id = state.top_level_span.id();
@@ -444,7 +495,8 @@ impl ExecutionState {
                 parent_span_id,
                 schedule_len,
                 tag,
-                state.try_current().map(|t| t.id()),
+                Some(state.current().id()),
+                state.current_mut().signature.new_child(caller),
             );
             state.tasks.push(task);
 
--- a/shuttle/src/runtime/task/mod.rs
+++ b/shuttle/src/runtime/task/mod.rs
@@ -9,8 +9,11 @@ use crate::thread::LocalKey;
 use bitvec::prelude::*;
 use std::any::Any;
 use std::cell::RefCell;
+use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
+use std::hash::{DefaultHasher, Hash, Hasher};
+use std::panic::Location;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::task::{Context, Waker};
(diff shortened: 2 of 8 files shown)


shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
     }
 }
 
+/// A task signature is an identifier that is intended to be *mostly* stable across executions
+/// and allow for categorizations of tasks according to how they were created. It provides three
+/// levels of granularity: static spawn location, dynamic single-task context, and dynamic concurrent
+/// context. These are each represented by a u64 so that the details of how they are computed can
```
sarsko: Not clear to me what "dynamic single-task context, and dynamic concurrent context" are

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
...
+            spawn_call_site_hash: h,
+            parent_signature_hash: 0,
+            signature_hash: h,
+            counter: 0,
+            child_counters: HashMap::new(),
+        }
+    }
+
+    fn new_child(
+        self: &mut TaskSignature,
```
sarsko: -> `&mut Self`

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
...
+    counter: u32,
+    signature_hash: u64,
+
+    child_counters: HashMap<&'static Location<'static>, u32>,
+}
+
+impl TaskSignature {
+    pub(crate) fn new(state: &mut ExecutionState, spawn_call_site: &'static Location<'static>) -> TaskSignature {
+        let mut hasher = DefaultHasher::new();
+        if let Some(t) = state.try_current_mut() {
```
sarsko: We always know the parent (with the exception of Task 0 which is parentless, and we already get all of this in the spawn and task creation functions. Why not have `new` take parent as `&mut` param instead of `ExecutionState`?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
...
+    }
+
+    /// Hash of the static location within the source code where the task was spawned
+    pub(crate) fn static_create_location(self: &TaskSignature) -> u64 {
+        self.spawn_call_site_hash
+    }
+
+    /// Hash of the dynamic single-task context in which the task was spawned
+    #[allow(unused)]
+    pub(crate) fn single_thread_ctxt(self: &TaskSignature) -> u64 {
```
sarsko: `ctxt` -> `ctx`

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
...
+        self.counter as u64
+    }
+
+    /// Hash of the dynamic concurrent context in which the task was spawned
+    #[allow(unused)]
+    pub(crate) fn multi_thread_ctxt(self: &TaskSignature) -> u64 {
```
sarsko: `ctxt` -> `ctx`

shuttle/src/runtime/task/mod.rs
sarsko: Why unused?

shuttle/src/runtime/task/mod.rs
sarsko: Why unused?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
...
+        self.parent_signature_hash
+    }
+
+    /// Combined signature of the static location, single-threaded context and threading context
+    /// where the task was spawned.
+    #[allow(unused)]
+    pub(crate) fn signature_hash(self: &TaskSignature) -> u64 {
```
sarsko, line 213: Why unused?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +150,102 @@ where
     }
 }
 
+/// A task signature is an identifier that is intended to be *mostly* stable across executions
+/// and allow for categorizations of tasks according to how they were created. It provides three
+/// levels of granularity: static spawn location, dynamic single-task context, and dynamic concurrent
+/// context. These are each represented by a u64 so that the details of how they are computed can
+/// be non-breaking changes in the future. All three are also combined into a single signature for
+/// fast checking of signatures at runtime
```
sarsko: I worry that since signatures take their parent's signature (rather than just their parent's call site for instance), any change of a parent will propagate to mean every child (and grandchild etc) will have a different signature.

shuttle/tests/basic/task.rs
```diff
@@ -0,0 +1,251 @@
...
+use test_log::test;
+use tracing::field::{Field, Visit};
+use tracing::{trace, Event, Id, Metadata, Subscriber};
+
+type SigCounterMap = Arc<Mutex<HashMap<u64, usize>>>;
+#[derive(Clone)]
+pub struct SignatureSubscriber {
+    pub signatures: SigCounterMap,
+    pub static_create_locations: SigCounterMap,
+}
```
sarsko, line 13: Nit but I don't think all of this has to be `pub`

shuttle/tests/basic/task.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    let scheduler = RandomScheduler::new(iterations);
+    let runner = Runner::new(scheduler, Default::default());
+    runner.run(test_fn);
+
+    (signatures, static_create_locations)
+}
+
+#[test]
+fn task_signatures_same_function() {
```
sarsko: Could you add a comment for each test explaining what they're checking?

shuttle/src/runtime/task/mod.rs
sarsko: Maybe I'm misremembering what we get and where. I just remember that we get parent in order to trace and to labels, and I think Signature fits in well with that.

jorajeev: +1 to not passing around the `ExecutionState`. The only parentless task is Task 0, right? We should have a special API for that, and pass in the parent for all the other calls.

jorajeev: Also, don't we need to implement PartialEq, Eq for TaskSignature?

shuttle/src/runtime/task/mod.rs
```diff
@@ -233,14 +339,17 @@ impl Task {
...
+            signature,
         };
 
         if let Some(tag) = tag {
             task.set_tag(tag);
         }
 
-        error_span!(parent: parent_span_id, "new_task", parent = ?parent_task_id, i = schedule_len)
-            .in_scope(|| event!(Level::INFO, "created task: {:?}", task.id));
+        // Note: the tests for the task signature depend on tracing the task signature and creation point here
```
jorajeev: nit: say which tests

jorajeev (commented):
As discussed offline, let's consider making the task signature a vector of Location, instead of only recording the parent hash in the child signature.

shuttle/src/runtime/task/mod.rs
dylanjwolff: Yeah, I am only passing the `ExecutionState` because we need it to check if the current task is parentless; `new` exists just so that we don't duplicate that check in both `spawn_future` and `spawn_thread`. 
 
I can rename `new` to be a helper in `execution.rs` instead so that `ExecutionState` isn't being passed around in the task module.

> Also, don't we need to implement PartialEq, Eq for TaskSignature?

Good point

shuttle/src/runtime/task/mod.rs
dylanjwolff: I've updated the doc comment to correspond to the changes and also hopefully be more clear

shuttle/src/runtime/task/mod.rs
dylanjwolff: yeah, this is my reasoning for the different levels of granularity -- sometimes we may want them to actually have those different signatures, sometimes we may not, depending on the use-case

shuttle/tests/basic/task.rs
dylanjwolff: I changed this test name and a couple of others to be more descriptive. I think it's better if they are understandable from the names since that is what shows up in the terminal/logs when they fail. But if any are still not clear lmk and I can add comments.

shuttle/src/runtime/task/mod.rs
dylanjwolff: Both of these points are now addressed in the current PR

shuttle/src/runtime/execution.rs
```diff
@@ -789,3 +809,13 @@ impl Drop for ExecutionState {
...
     }
 }
+
+fn create_signature(state: &mut ExecutionState, spawn_call_site: &'static Location<'static>) -> TaskSignature {
+    let mut hasher = DefaultHasher::new();
+    if let Some(t) = state.try_current_mut() {
+        t.signature.new_child(spawn_call_site, &mut hasher)
+    } else {
+        // If we can't find the current task, then the new child is parentless
+        TaskSignature::new_parentless(spawn_call_site, &mut hasher)
```
jorajeev: Is this ever called for any task other than Task 0?  If not, why not have a special API for Task 0 and do an `unwrap` on line 815?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +149,78 @@ where
     }
 }
 
+/// A task signature is an identifier that is intended to be *mostly* stable across executions
+/// and allow for categorization of tasks according to how they were created. It provides three
```
jorajeev: typo: "two levels" now, not 3

"static and dynamic"

shuttle/src/runtime/execution.rs
```diff
@@ -789,3 +809,13 @@ impl Drop for ExecutionState {
         assert!(self.has_cleaned_up || std::thread::panicking());
     }
 }
+
+fn create_signature(state: &mut ExecutionState, spawn_call_site: &'static Location<'static>) -> TaskSignature {
+    let mut hasher = DefaultHasher::new();
```
jorajeev: Why are we passing in a `Hasher` from the caller?  Why not move the Hasher creation into the `new_child` and `new_parentless` methods?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,6 +149,78 @@ where
...
+        Self {
+            task_creation_stack,
+            spawn_call_site_hash,
+            signature_hash: hasher.finish(),
+            child_counters: HashMap::new(),
+        }
+    }
+
+    /// Hash of the static location within the source code where the task was spawned
+    pub(crate) fn static_create_location_hash(self: &TaskSignature) -> u64 {
```
jorajeev: nit: argument should be `&self`

shuttle/src/runtime/execution.rs
dylanjwolff: no particular reason -- I've moved hasher creation in the latest version of this PR

shuttle/src/runtime/execution.rs
dylanjwolff: The special API for Task 0 *is* `new_parentless`, we are just checking which one to call here.

If you want, I can lift the choice of API to the `spawn_future` / `spawn_thread` level in `execution.rs`. In that case, we'd have a third function there `spawn_main_thread` and then everything called from there (including `create_signature`) can assume 0 / nonzero TaskId accordingly. Just lmk if you'd prefer this strongly enough to roll into this PR or if we want to do it as a separate smaller cleanup refactor later.

jorajeev: My preference would be to have a separate `spawn_main_thread` and then we don't need this if statement

dylanjwolff: OK, I've refactored to lift the API to that level

shuttle/src/runtime/execution.rs
```diff
@@ -373,9 +372,53 @@ impl ExecutionState {
         });
     }
 
+    pub(crate) fn spawn_main_thread(
```
jorajeev, line 377: Could we avoid copy/pasting the code for spawn_thread?  Is there another solution?

dylanjwolff: The old version of this PR had no duplicated code, just a single function call in both spawn functions, so we can always revert to that.

But I really don't think this is too bad in terms of duplication: only ~12 LoC are the same between them if you don't count the `from_closure` call that is split across many lines.

I can draw out some of the shared logic into a helper function or a `spawn_thread_inner`, but I (subjectively) think it's not be worth it; past a certain point, for me, the extra indirection starts to make it harder to understand what the code is actually doing.

dylanjwolff: @jorajeev, as discussed offline, I added comments above all three spawn functions to indicate that changes will likely need to be applied to all three

jorajeev (approved):

sarsko (approved):

PR #200, opened 2025-08-15
Add UniformRandomWalk Scheduler to Shuttle

dylanjwolff:
This PR adds UniformRandomScheduler to implement URW from the [Selectively Uniform Random Walk](https://dl.acm.org/doi/10.1145/3669940.3707214) (ASPLOS'25).

For event count estimates, the scheduler does a single trial run in the first iteration, and uses these estimates for all future iterations. This is a relatively naive approach, which we can likely improve with better heuristics in future changes. 

Notably, these changes are dependent on https://github.com/awslabs/shuttle/pull/196, so you can look at the commit directly to see [the changes for this PR in isolation](https://github.com/awslabs/shuttle/commit/02606b6b27272120adc7114a1c101138340c07cb) until that has been merged.

---

### DIFF
--- a/shuttle/src/future/mod.rs
+++ b/shuttle/src/future/mod.rs
@@ -11,6 +11,7 @@ use crate::runtime::thread;
 use std::error::Error;
 use std::fmt::{Display, Formatter};
 use std::future::Future;
+use std::panic::Location;
 use std::pin::Pin;
 use std::result::Result;
 use std::sync::Arc;
@@ -18,14 +19,14 @@ use std::task::{Context, Poll, Waker};
 
 pub mod batch_semaphore;
 
-fn spawn_inner<F>(fut: F) -> JoinHandle<F::Output>
+fn spawn_inner<F>(fut: F, caller: &'static Location<'static>) -> JoinHandle<F::Output>
 where
     F: Future + 'static,
     F::Output: 'static,
 {
     let stack_size = ExecutionState::with(|s| s.config.stack_size);
     let inner = Arc::new(std::sync::Mutex::new(JoinHandleInner::default()));
-    let task_id = ExecutionState::spawn_future(Wrapper::new(fut, inner.clone()), stack_size, None);
+    let task_id = ExecutionState::spawn_future(Wrapper::new(fut, inner.clone()), stack_size, None, caller);
 
     thread::switch();
 
@@ -33,22 +34,26 @@ where
 }
 
 /// Spawn a new async task that the executor will run to completion.
+#[track_caller]
 pub fn spawn<F>(fut: F) -> JoinHandle<F::Output>
 where
     F: Future + Send + 'static,
     F::Output: Send + 'static,
 {
-    spawn_inner(fut)
+    let caller = Location::caller();
+    spawn_inner(fut, caller)
 }
 
 /// Spawn a new async task that the executor will run to completion.
 /// This is just `spawn` without the `Send` bound, and it mirrors `spawn_local` from Tokio.
+#[track_caller]
 pub fn spawn_local<F>(fut: F) -> JoinHandle<F::Output>
 where
     F: Future + 'static,
     F::Output: 'static,
 {
-    spawn_inner(fut)
+    let caller = Location::caller();
+    spawn_inner(fut, caller)
 }
 
 /// An owned permission to abort a spawned task, without awaiting its completion.
--- a/shuttle/src/lib.rs
+++ b/shuttle/src/lib.rs
@@ -321,6 +321,18 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a *uniformly* random scheduler for some number of iterations.
+/// Each iteration will run a (potentially) different randomized schedule.
+pub fn check_urw<F>(f: F, iterations: usize)
+where
+    F: Fn() + Send + Sync + 'static,
+{
+    use crate::scheduler::UniformRandomScheduler;
+
+    let runner = Runner::new(UniformRandomScheduler::new(iterations), Default::default());
+    runner.run(f);
+}
+
 /// Run the given function under a randomized concurrency scheduler for some number of iterations.
 /// Each iteration will run a (potentially) different randomized schedule.
 pub fn check_random<F>(f: F, iterations: usize)
--- a/shuttle/src/runtime/execution.rs
+++ b/shuttle/src/runtime/execution.rs
@@ -2,7 +2,7 @@ use crate::runtime::failure::{init_panic_hook, persist_failure, persist_task_fai
 use crate::runtime::storage::{StorageKey, StorageMap};
 use crate::runtime::task::clock::VectorClock;
 use crate::runtime::task::labels::Labels;
-use crate::runtime::task::{ChildLabelFn, Task, TaskId, TaskName, DEFAULT_INLINE_TASKS};
+use crate::runtime::task::{ChildLabelFn, Task, TaskId, TaskName, TaskSignature, DEFAULT_INLINE_TASKS};
 use crate::runtime::thread::continuation::PooledContinuation;
 use crate::scheduler::{Schedule, Scheduler};
 use crate::thread::thread_fn;
@@ -14,7 +14,8 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
-use std::panic;
+use std::hash::DefaultHasher;
+use std::panic::{self, Location};
 use std::rc::Rc;
 use std::sync::Arc;
 use tracing::{trace, Span};
@@ -65,6 +66,7 @@ impl Execution {
     /// Run a function to be tested, taking control of scheduling it and any tasks it might spawn.
     /// This function runs until `f` and all tasks spawned by `f` have terminated, or until the
     /// scheduler returns `None`, indicating the execution should not be explored any further.
+    #[track_caller]
     pub(crate) fn run<F>(mut self, config: &Config, f: F)
     where
         F: FnOnce() + Send + 'static,
@@ -77,6 +79,7 @@ impl Execution {
 
         let _guard = init_panic_hook(config.clone());
 
+        let caller = Location::caller();
         EXECUTION_STATE.set(&state, move || {
             // Spawn `f` as the first task
             ExecutionState::spawn_thread(
@@ -84,6 +87,7 @@ impl Execution {
                 config.stack_size,
                 Some("main-thread".to_string()),
                 Some(VectorClock::new()),
+                caller,
             );
 
             // Run the test to completion
@@ -375,7 +379,12 @@ impl ExecutionState {
 
     /// Spawn a new task for a future. This doesn't create a yield point; the caller should do that
     /// if it wants to give the new task a chance to run immediately.
-    pub(crate) fn spawn_future<F>(future: F, stack_size: usize, name: Option<String>) -> TaskId
+    pub(crate) fn spawn_future<F>(
+        future: F,
+        stack_size: usize,
+        name: Option<String>,
+        caller: &'static Location<'static>,
+    ) -> TaskId
     where
         F: Future<Output = ()> + 'static,
     {
@@ -401,6 +410,7 @@ impl ExecutionState {
                 schedule_len,
                 tag,
                 state.try_current().map(|t| t.id()),
+                create_signature(state, caller),
             );
 
             state.tasks.push(task);
@@ -416,6 +426,7 @@ impl ExecutionState {
         stack_size: usize,
         name: Option<String>,
         mut initial_clock: Option<VectorClock>,
+        caller: &'static Location<'static>,
     ) -> TaskId {
         let task_id = Self::with(|state| {
             let parent_span_id = state.top_level_span.id();
@@ -445,6 +456,7 @@ impl ExecutionState {
                 schedule_len,
                 tag,
                 state.try_current().map(|t| t.id()),
+                create_signature(state, caller),
             );
             state.tasks.push(task);
 
@@ -577,6 +589,10 @@ impl ExecutionState {
         self.try_get(self.current_task.id()?)
     }
 
+    pub(crate) fn try_current_mut(&mut self) -> Option<&mut Task> {
+        self.try_get_mut(self.current_task.id()?)
+    }
+
     pub(crate) fn get(&self, id: TaskId) -> &Task {
         self.try_get(id).unwrap()
     }
@@ -587,6 +603,10 @@ impl ExecutionState {
 
     pub(crate) fn try_get(&self, id: TaskId) -> Option<&Task> {
         self.tasks.get(id.0)
+    }
+
+    pub(crate) fn try_get_mut(&mut self, id: TaskId) -> Option<&mut Task> {
+        self.tasks.get_mut(id.0)
     }
 
     pub(crate) fn in_cleanup(&self) -> bool {
@@ -787,5 +807,15 @@ impl ExecutionState {
 impl Drop for ExecutionState {
     fn drop(&mut self) {
         assert!(self.has_cleaned_up || std::thread::panicking());
+    }
+}
+
+fn create_signature(state: &mut ExecutionState, spawn_call_site: &'static Location<'static>) -> TaskSignature {
+    let mut hasher = DefaultHasher::new();
+    if let Some(t) = state.try_current_mut() {
+        t.signature.new_child(spawn_call_site, &mut hasher)
+    } else {
+        // If we can't find the current task, then the new child is parentless
+        TaskSignature::new_parentless(spawn_call_site, &mut hasher)
     }
 }
--- a/shuttle/src/runtime/task/mod.rs
+++ b/shuttle/src/runtime/task/mod.rs
@@ -9,8 +9,11 @@ use crate::thread::LocalKey;
 use bitvec::prelude::*;
 use std::any::Any;
 use std::cell::RefCell;
+use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
+use std::hash::{Hash, Hasher};
+use std::panic::Location;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::task::{Context, Waker};
@@ -146,6 +149,78 @@ where
     }
 }
 
+/// A task signature is an identifier that is intended to be *mostly* stable across executions
+/// and allow for categorization of tasks according to how they were created. It provides three
+/// levels of granularity: static (compile-time) spawn location, dynamic (run-time) context where
+/// that spawn location was reached. The static spawn location and signature are each represented
+/// by a u64 so that the details of how they are computed can be non-breaking changes in the future.
+/// Hashes are all pre-computed for fast checking of equality of signatures at runtime.
+#[derive(Debug, Clone)]
+pub(crate) struct TaskSignature {
+    /// The task creation stack is a tuple of (create location, number of tasks created at that location in the parent)
+    task_creation_stack: Vec<(&'static Location<'static>, u32)>,
+    spawn_call_site_hash: u64,
+    signature_hash: u64,
+    child_counters: HashMap<&'static Location<'static>, u32>,
+}
+
+impl TaskSignature {
+    pub(crate) fn new_parentless(spawn_call_site: &'static Location<'static>, hasher: &mut impl Hasher) -> TaskSignature {
+        let task_creation_stack = vec![(spawn_call_site, 0)];
+        task_creation_stack.hash(hasher);
+        Self {
+            task_creation_stack,
+            spawn_call_site_hash: 0,
+            signature_hash: hasher.finish(),
+            child_counters: HashMap::new(),
+        }
+    }
+
+    pub(crate) fn new_child(&mut self, spawn_call_site: &'static Location<'static>, hasher: &mut impl Hasher) -> Self {
+        let counter = self.child_counters.entry(spawn_call_site).or_insert(0);
+        *counter += 1;
+        let mut task_creation_stack = self.task_creation_stack.clone();
+        task_creation_stack.push((spawn_call_site, *counter));
+
+        spawn_call_site.hash(hasher);
+        let spawn_call_site_hash = hasher.finish();
+
+        task_creation_stack.hash(hasher);
+
+        Self {
+            task_creation_stack,
+            spawn_call_site_hash,
+            signature_hash: hasher.finish(),
+            child_counters: HashMap::new(),
+        }
+    }
+
+    /// Hash of the static location within the source code where the task was spawned
+    pub(crate) fn static_create_location_hash(self: &TaskSignature) -> u64 {
+        self.spawn_call_site_hash
+    }
+
+    /// Combined signature of the static location and dynamic context
+    /// context where the task was spawned.
+    pub(crate) fn signature_hash(self: &TaskSignature) -> u64 {
+        self.signature_hash
+    }
+}
+
+impl Hash for TaskSignature {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.task_creation_stack.hash(state);
+    }
+}
+
+impl PartialEq for TaskSignature {
+    fn eq(&self, other: &Self) -> bool {
+        self.signature_hash == other.signature_hash
+    }
+}
+
+impl Eq for TaskSignature {}
+
 /// A `Task` represents a user-level unit of concurrency. Each task has an `id` that is unique within
 /// the execution, and a `state` reflecting whether the task is runnable (enabled) or not.
 #[derive(Debug)]
@@ -188,6 +263,11 @@ pub struct Task {
     // Arbitrarily settable tag which is inherited from the parent.
     #[allow(deprecated)]
     tag: Option<Arc<dyn Tag>>,
+
+    /// The signature of a Task; this is an identifier that is *not* guaranteed to be unique but should be *mostly*
+    /// stable across iterations in a single Shuttle test. Tasks with the same signature are very likely to exhibit
+    /// similar behavior
+    pub(crate) signature: TaskSignature,
 }
 
 #[allow(deprecated)]
@@ -204,6 +284,7 @@ impl Task {
         schedule_len: usize,
         tag: Option<Arc<dyn Tag>>,
         parent_task_id: Option<TaskId>,
+        signature: TaskSignature,
     ) -> Self {
         #[cfg(all(any(test, feature = "vector-clocks"), not(feature = "bench-no-vector-clocks")))]
         assert!(id.0 < clock.time.len());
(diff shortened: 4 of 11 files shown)


shuttle/src/future/mod.rs
```diff
@@ -11,44 +11,49 @@ use crate::runtime::thread;
...
 /// This is just `spawn` without the `Send` bound, and it mirrors `spawn_local` from Tokio.
+#[track_caller]
 pub fn spawn_local<F>(fut: F) -> JoinHandle<F::Output>
 where
     F: Future + 'static,
     F::Output: 'static,
 {
-    spawn_inner(fut)
+    let caller = Location::caller();
+    spawn_inner(fut, caller)
```
sarsko: Nit on these two is to compact to `spawn_inner(fut, Location::caller())` (yes, very nitty, but it helps when reading the code to see quickly that the function is just a thin wrapper)

shuttle/src/lib.rs
```diff
@@ -321,6 +321,18 @@ where
     runner.run(f);
 }
 
+/// Run the given function under a *uniformly* random scheduler for some number of iterations.
+/// Each iteration will run a (potentially) different randomized schedule.
+pub fn check_urw<F>(f: F, iterations: usize)
+where
+    F: Fn() + Send + Sync + 'static,
+{
+    use crate::scheduler::UniformRandomScheduler;
```
sarsko: Unrelated to the PR, but I think all the `check` functions should have their `use` line moved to the top of the file.

shuttle/src/runtime/execution.rs
```diff
@@ -77,13 +79,15 @@ impl Execution {
...
 
+        let caller = Location::caller();
         EXECUTION_STATE.set(&state, move || {
             // Spawn `f` as the first task
             ExecutionState::spawn_thread(
                 Box::new(move || thread_fn(f, Default::default())),
                 config.stack_size,
                 Some("main-thread".to_string()),
                 Some(VectorClock::new()),
+                caller,
```
sarsko: Ditto on the nit here

shuttle/src/runtime/execution.rs
```diff
@@ -789,3 +809,13 @@ impl Drop for ExecutionState {
         assert!(self.has_cleaned_up || std::thread::panicking());
     }
 }
+
+fn create_signature(state: &mut ExecutionState, spawn_call_site: &'static Location<'static>) -> TaskSignature {
```
sarsko: `state: &mut ExecutionState` -> `&mut self`.

Not a big fan of taking `&mut ExecutionState`, but I'm also not a fan of having the function be on `Option<&mut Task>`. Hmm.

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
     }
 }
 
+/// A task signature is an identifier that is intended to be *mostly* stable across executions
+/// and allow for categorization of tasks according to how they were created. It provides three
+/// levels of granularity: static (compile-time) spawn location, dynamic (run-time) context where
```
sarsko: I'm not sure "It provides three levels of granularity" is correct? The task signature is composed of three parts, which together make up the entire signature, no?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
...
+/// that spawn location was reached. The static spawn location and signature are each represented
+/// by a u64 so that the details of how they are computed can be non-breaking changes in the future.
```
sarsko: This is a non-care. None of this is pub so all changes are non-breaking. The reason we have `u64`s is because we want to do efficient comparisons.

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
...
+    parent_signature_hash: u64,
+    signature_hash: u64,
+    child_counters: HashMap<&'static Location<'static>, u32>,
+}
+
+impl TaskSignature {
+    pub(crate) fn new_parentless(
+        spawn_call_site: &'static Location<'static>,
+        hasher: &mut impl Hasher,
+    ) -> TaskSignature {
```
sarsko: This applies to `new_child` as well: I think we want to remove the `&mut impl Hasher` parameter. We don't care about being generic over the hasher, we simply want a hash. Hash can either be constructed on each call or be stored in `TaskSignature`.

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
...
+        let task_creation_stack = vec![(spawn_call_site, 0)];
```
sarsko: Shouldn't count be 1?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
...
+        task_creation_stack.hash(hasher);
+        Self {
+            task_creation_stack,
+            spawn_call_site_hash: 0,
```
sarsko: Surely this should be a hash of `spawn_call_site` ?

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
...
+            parent_signature_hash: 0,
+            signature_hash: hasher.finish(),
+            child_counters: HashMap::new(),
+        }
+    }
+
+    pub(crate) fn new_child(&mut self, spawn_call_site: &'static Location<'static>, hasher: &mut impl Hasher) -> Self {
+        let counter = self.child_counters.entry(spawn_call_site).or_insert(0);
+        *counter += 1;
```
sarsko: Nit: Think these can be compacted into one line.

shuttle/src/runtime/task/mod.rs
```diff
@@ -146,11 +149,95 @@ where
...
+
+    /// Signature hash of the parent of this task
+    pub(crate) fn parent_signature_hash(self: &TaskSignature) -> u64 {
+        self.parent_signature_hash
+    }
+}
+
+impl Hash for TaskSignature {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.task_creation_stack.hash(state);
```
sarsko: This doesn't look correct. This will make multiple different `TaskSignature`s hash to the same thing. Isn't the correct thing to use the `signature_hash`?

shuttle/src/runtime/task/mod.rs
```diff
@@ -188,6 +275,11 @@ pub struct Task {
     // Arbitrarily settable tag which is inherited from the parent.
     #[allow(deprecated)]
     tag: Option<Arc<dyn Tag>>,
+
+    /// The signature of a Task; this is an identifier that is *not* guaranteed to be unique but should be *mostly*
+    /// stable across iterations in a single Shuttle test. Tasks with the same signature are very likely to exhibit
+    /// similar behavior
+    pub(crate) signature: TaskSignature,
```
sarsko: How does it not get unique? Aren't the `u32`s gonna differ?

shuttle/src/runtime/task/mod.rs
sarsko: For that matter we could probably even just derive `Hash`.

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,231 @@
...
+
+impl Drop for CurrentSeedDropGuard {
+    fn drop(&mut self) {
+        if let Some(s) = self.inner {
+            eprintln!(
+                "failing seed:\n\"\n{s}\n\"\nTo replay the failure, set the environment variable SHUTTLE_RANDOM_SEED to the seed and run `shuttle::check_urw`."
+            )
+        }
+    }
+}
```
sarsko: This could be moved from the `RandomScheduler` to `lib.rs` and `pub(crate)`d instead of reimplemented. `shuttle::check_urw` has to be taken as paramerer

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,231 @@
...
+        } else {
+            if self.events_counted_but_not_initialized() {
+                // If we have never initialized the event counts before, we need to aggregate the count estimates
+                // from each child task to its parents
+                info!("Finished estimation of event counts for URW");
+                info!(
+                    "Estimated event counts for URW (pre-parent subsumption): {:?}",
+                    self.signature_event_counts
+                );
+                // Iterating in reverse spawn order to ensure counts are propagated to grandparents correctly
```
sarsko: We should probably have some developer internal documentation on how URW works and what the counts are (for instance state somewhere that all counts are the tasks own scheduling decisions summed with that of its children)

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,231 @@
...
+            self.current_seed.update(seed);
+            Some(Schedule::new(seed))
+        }
+    }
+
+    fn next_task(&mut self, runnable: &[&Task], _current: Option<TaskId>, _is_yielding: bool) -> Option<TaskId> {
+        if let Some(task_event_counts) = self.task_event_counts.as_mut() {
+            // If the event counts have been estimated...
+
+            // We need to loop over the tasks to identify if there has been a `spawn` event
```
sarsko: This looks wildly inefficient and the code itself is not particularly easy to follow. Can we add stuff elsewhere to alleviate?

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    fn next_u64(&mut self) -> u64 {
+        self.data_source.next_u64()
+    }
+}
+
+#[inline]
+fn get_tid(task: &Task) -> usize {
+    task.id().into()
+}
```
sarsko, line 241: We call this function twice, not sure having a function for it is worth

shuttle/src/thread.rs
```diff
@@ -381,13 +396,15 @@ impl Builder {
...
+    #[track_caller]
     pub fn spawn<F, T>(self, f: F) -> std::io::Result<JoinHandle<T>>
     where
         F: FnOnce() -> T,
         F: Send + 'static,
         T: Send + 'static,
     {
-        Ok(spawn_named(f, self.name, self.stack_size))
+        let caller = Location::caller();
+        Ok(spawn_named(f, self.name, self.stack_size, caller))
```
sarsko: My caller nits apply to these as well

shuttle/tests/basic/task.rs
```diff
@@ -0,0 +1,250 @@
...
+use tracing::{trace, Event, Id, Metadata, Subscriber};
+
+type SigCounterMap = Arc<Mutex<HashMap<u64, usize>>>;
+#[derive(Clone)]
+struct SignatureSubscriber {
+    signatures: SigCounterMap,
+    static_create_locations: SigCounterMap,
+}
+
+impl SignatureSubscriber {
```
sarsko: Put in comment describing what this is and what it's looking for

shuttle/tests/demo/surw.rs
```diff
@@ -0,0 +1,168 @@
...
+            let t1 = future::spawn(future1);
+            let t2 = future::spawn(future2);
+
+            future::block_on(async {
+                t1.await.ok();
+                t2.await.ok();
+            });
+        },
+        RandomScheduler::new(NUM_SHUTTLE_ITERATIONS_URW * 10),
+        false, // Random scheduler cannot find this bug even with an order of magnitude more executions
```
sarsko, line 105: Surely this is just unlikely and thus the test is flaky?

shuttle/src/runtime/task/mod.rs
dylanjwolff: the task creation stack includes the current task's spawn context (with the counter) so that alone will make it unique

shuttle/src/scheduler/urw.rs
dylanjwolff: once the `next_event` stuff lands we can replace this with a simple match on the event enum for spawn events. I don't think it is worth hacking something else in the meantime

shuttle/src/scheduler/urw.rs
dylanjwolff: if it's not encapsulated you need to cast the `into` to usize when the type can't be inferred, so it ends up being a bit verbose and ugly IMO

shuttle/tests/demo/surw.rs
dylanjwolff: I've made the tests deterministic by initializing with a seed. This still could lead to a spurious failure in a very pathological case, but I think that is true of any bug-finding test that doesn't use DFS, so it is somewhat unavoidable

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::scheduler::{Schedule, Scheduler};
+use rand::rngs::OsRng;
+use rand::seq::SliceRandom;
+use rand::{RngCore, SeedableRng};
+use rand_pcg::Pcg64Mcg;
+use std::collections::{HashMap, HashSet};
+use tracing::{info, trace, warn};
+
+/// A scheduler which implements Uniform Random Walk (URW) from "Selectively Uniform Concurrency Testing" by
+/// Huan Zhao, Dylan Wolff, Umang Mathur, and Abhik Roychoudhury - ASPLOS '25. The URW algorithm samples
```
sarsko: Add link to paper ?

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+/// all interleavings *uniformly* given an accurate estimate of the number of events which will take place
+/// on each task. This implementation uses a single trial run of the program to generate these estimates.
+/// During the trial run, it uses vanilla random walk (the same algorithm as [`crate::scheduler::RandomScheduler`]).
+/// As discussed in the paper, parent Tasks subsume the count estimates of each child until the child is spawned.
```
sarsko: Somewhat cryptic. What is written here is that the count of a parent task is equal to its scheduling points added with the sum of the scheduling points of its children which have not yet been spawned.

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+///
+/// The RNG used is contained within the scheduler, allowing it to be reused across executions in
+/// order to get different random schedules each time. The scheduler has an optional bias towards
```
sarsko: Where is this optional bias?

dylanjwolff: whoops! copy paste error from the random scheduler

shuttle/src/scheduler/urw.rs
dylanjwolff: updated desc

shuttle/src/runtime/task/mod.rs
dylanjwolff: this is mostly an issue from the already merged task signature PR #196

I think actually they are unique for now, but that might change later. And the resource signatures are not unique

shuttle/src/runtime/task/mod.rs
dylanjwolff: also from #196, but it doesn't really matter if we start counting at 0 or 1

shuttle/src/lib.rs
dylanjwolff: I've made a note to do this in a small future PR

shuttle/src/runtime/task/mod.rs
```diff
@@ -214,6 +217,11 @@ impl TaskSignature {
     pub(crate) fn signature_hash(self: &TaskSignature) -> u64 {
         self.signature_hash
     }
+
+    /// Signature hash of the parent of this task
+    pub(crate) fn parent_signature_hash(self: &TaskSignature) -> u64 {
```
jorajeev: Could we avoid the inconsistent use of

`fn foo(&self) { .. }`

and

`fn foo(self: &Foo) { .. }`

Let's just use `fn foo(&self) { .. }`

shuttle/src/scheduler/mod.rs
```diff
@@ -23,6 +24,7 @@ pub use random::RandomScheduler;
 pub use replay::ReplayScheduler;
 pub use round_robin::RoundRobinScheduler;
 pub use uncontrolled_nondeterminism::UncontrolledNondeterminismCheckScheduler;
+pub use urw::UniformRandomScheduler;
```
jorajeev: nit: to be consistent with `pct` and `PctScheduler`, this should be `urw` and `UrwScheduler` ?

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::scheduler::data::DataSource;
+use crate::scheduler::{Schedule, Scheduler};
+use rand::rngs::OsRng;
+use rand::seq::SliceRandom;
+use rand::{RngCore, SeedableRng};
+use rand_pcg::Pcg64Mcg;
+use std::collections::{HashMap, HashSet};
+use tracing::{info, trace, warn};
+
+/// A scheduler which implements Uniform Random Walk (URW) from "Selectively Uniform Concurrency Testing" by Huan Zhao,
```
jorajeev: nit: reword this comment to match the format/style of the comment in the PctScheduler

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+/// Dylan Wolff, Umang Mathur, and Abhik Roychoudhury - [ASPLOS '25](https://dl.acm.org/doi/abs/10.1145/3669940.3707214).
+/// The URW algorithm samples all interleavings *uniformly* given an accurate estimate of the number of events which will
+/// take place on each task. This implementation uses a single trial run of the program to generate these estimates.
```
jorajeev: How hard would it be to generalize to using N initial trial runs?

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+#[derive(Debug)]
+pub struct UniformRandomScheduler {
+    max_iterations: usize,
+    rng: Pcg64Mcg,
+    iterations: usize,
+    data_source: RandomDataSource,
+    /// Number of events remaining on each task in the current execution
+    task_event_counts: Option<Vec<usize>>,
+    /// Event count estimates for each task by signature
+    signature_event_counts: HashMap<u64, usize>,
```
jorajeev: I would prefer to avoid naked `u64`s as signatures, and instead have a newtype `struct SignatureHash(u64)`, which is returned by `TaskSignature::signature_hash()`.

Then, these fields could be just `event_counts: HashMap<SignatureHash, usize>` etc

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+                    self.signature_event_counts
+                );
+                // Iterating in reverse spawn order to ensure counts are propagated to grandparents correctly
+                for (parent_sig, child_sig) in self.signature_parents.iter().rev() {
+                    let child_ct = *self.signature_event_counts.get(child_sig).unwrap();
+                    self.signature_event_counts
+                        .entry(*parent_sig)
+                        .and_modify(|parent_ct| *parent_ct += child_ct);
+                }
+                info!(
```
jorajeev: Don't think we want to dump this state (pre- and post- child count merging) at `info!` level.  For a large program, this could be a lot of output.  Should be `trace!`

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+                    "Estimated event counts for URW (post-parent subsumption): {:?}",
+                    self.signature_event_counts
+                );
+
+                // The implemenation of URW currently assumes that each task has a unique signature in a single execution.
+                // Thus the number of spawns events should equal the number of unique child signatures
+                debug_assert!(
```
jorajeev: Move this up so it fails before we do all the merging work on lines 97-102

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+            // simplified to a single conditional
+            for t in runnable {
+                let tid: usize = get_tid(t);
+                if tid == task_event_counts.len() {
+                    // Spawn: should remove child events from the parent and map remaining events to the correct task id
+                    let child_events = *self
+                        .signature_event_counts
+                        .get(&t.signature.signature_hash())
+                        .unwrap_or_else(|| {
+                            // TODO: we can probably handle unseen tasks less naively than estimating a single event
```
jorajeev: `1` seems like a bad default.  We could do the minimum, average, or (maybe best) the median (which needs to be computed only once)

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+                    task_event_counts.push(child_events);
+                    if let Some(ptid) = t.parent_task_id() {
+                        let ptid: usize = ptid.into();
+                        task_event_counts[ptid] = task_event_counts[ptid].saturating_sub(child_events).max(1);
+                    }
+                } else if tid > task_event_counts.len() {
+                    panic!("TID's expected to be spawned in ascending order in increments of 1");
+                }
+                // Any runnable task must have at least one remaining event
+                assert!(task_event_counts[tid] >= 1);
```
jorajeev: Add a comment that this is ensured by the code (esp line 171) below

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+            data_source: RandomDataSource::initialize(seed),
+            task_event_counts: None,
+            signature_event_counts: HashMap::new(),
+            signature_parents: Vec::new(),
+        }
+    }
+
+    #[inline]
+    fn events_counted_but_not_initialized(&self) -> bool {
+        self.task_event_counts.is_none() && !self.signature_event_counts.is_empty()
```
jorajeev: I'm not a big fan of using `None` and `is_empty` to track state.  Better to define an enum `UrwSchedulerState` that explicitly tracks what's going on.  I suspect that will also be easier to maintain if we want to use N initial iterations to determine event counts instead of a single random walk.

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,202 @@
...
+    fn next_task(&mut self, runnable: &[&Task], _current: Option<TaskId>, _is_yielding: bool) -> Option<TaskId> {
+        if let Some(task_event_counts) = self.task_event_counts.as_mut() {
+            // If the event counts have been estimated...
+
+            // We need to loop over the tasks to identify if there has been a `spawn` event
+            // In the future, if we embed a `next_event` enum field into the Task struct, this code can be
+            // simplified to a single conditional
+            for t in runnable {
+                let tid: usize = get_tid(t);
+                if tid == task_event_counts.len() {
```
jorajeev: Would this code be simpler if we just keep `task_event_counts` as a map instead of a vector?  You would still have to update parent weight when a child is spawned, and you would pay for the slightly higher cost of looking up the map (which is O(1) in general) vs vector access?  But maybe the code would be easier to understand and maintain?

shuttle/tests/demo/surw.rs
jorajeev: Not sure that initializing with a fixed seed is convincing though?  The point of random is that we run it with a different seed each time?

shuttle/tests/demo/surw.rs
```diff
@@ -0,0 +1,168 @@
...
+    assert_eq!(
+        expect_found,
+        has_found_schedule.load(std::sync::atomic::Ordering::SeqCst)
+    );
+}
+
+#[test]
+fn non_uniform_random_degrades_for_uneven_task_lengths() {
+    test_uneven_execution(
+        |counter1, counter2| {
```
jorajeev: Could we rewrite these tests to avoid repeating the closure on lines 50-67 and 75-93?

shuttle/src/scheduler/urw.rs
dylanjwolff: I've reworded it to look similar, but let me know if there is anything more specific you are looking for

shuttle/src/scheduler/urw.rs
dylanjwolff: Not hard, but I vote for doing it in a future PR. There are a few improvements to be done that we can roll into one PR later: generalizing to N trials, better estimation for unseen task signatures, and using the `next_task` info to detect spawn events

shuttle/src/scheduler/urw.rs
dylanjwolff: good idea!

shuttle/src/scheduler/urw.rs
dylanjwolff: I don't think it makes a meaningful difference in terms of clarity, so I think it's not worth the worse spatial locality and hashing overhead of a map

shuttle/src/scheduler/urw.rs
dylanjwolff: I've changed this to `min` for now. But this is still a placeholder. I think the proper solution is a kind of "fuzzy" match -- for example, if we don't have an exact signature match, was there another task at least spawned at the same location? If so, we can use that as an estimate instead.

shuttle/tests/demo/surw.rs
dylanjwolff: I think we should initialize with the fixed seed so that the tests won't randomly flake in CI after passing locally.

I agree that this is less convincing, but every time we make a change that affects the number of events there is a chance that these tests will start failing (deterministically) if the assertion is not reliable.

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,230 @@
...
+use crate::scheduler::data::DataSource;
+use crate::scheduler::{Schedule, Scheduler};
+use rand::rngs::OsRng;
+use rand::seq::SliceRandom;
+use rand::{RngCore, SeedableRng};
+use rand_pcg::Pcg64Mcg;
+use std::collections::{HashMap, HashSet};
+use tracing::{trace, warn};
+
+type SignatureHash = u64;
```
jorajeev: I'd prefer a newtype here `struct SignatureHash(u64)` with `From`/`Into` methods to convert from/to u64.  Type aliases don't give you any type protection.

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,230 @@
...
+            task_event_counts: None,
+            signature_event_counts: HashMap::new(),
+            min_event_count: usize::MAX,
+            signature_parents: Vec::new(),
+            state: UrwSchedulerState::PreEstimation,
+        }
+    }
+
+    fn initialize_estimates_from_observed_counts(&mut self) {
+        // If we have never initialized the event counts before, we need to aggregate the count estimates
```
jorajeev, line 100: Could add `assert_eq!(self.state, UrwSchedulerState::Estimating)` here

shuttle/src/scheduler/urw.rs
```diff
@@ -0,0 +1,230 @@
...
+impl Scheduler for UrwRandomScheduler {
+    fn new_execution(&mut self) -> Option<Schedule> {
+        if self.iterations >= self.max_iterations {
+            self.signature_event_counts.clear();
+            self.signature_parents.clear();
+            self.task_event_counts = None;
+            return None;
+        }
+
+        match self.state {
```
jorajeev, line 193: Thanks for adding this - makes it a lot clearer

PR #202, opened 2025-08-18
Serialize schedule even on aborting panics. Refactor panic path.

sarsko:
This PR does a few different things, which in a sense is poor PR etiquette, but they also fit together, and I'm not sure splitting it up would have been any better. The general problem being improved upon is that of test panics and aborts, where the goal is to reduce the frequency of test failures with no serialized schedule.

The list of what the PR does:

1. Moves the schedule out of `ExecutionState`. This removes the error mode of not being able to serialize the schedule while `ExecutionState` is borrowed.
2. Removes most of the panic hook and injecting the schedule into the panic payload mechanism. My opinion is that this was hard to follow and reason about, buggy, and didn't provide any significant benefits (the only benefit afaik is that we can read the panic message, ie. what `expected = ` in Rust tests reads to get the schedule. This is not a significant feature, as we can still read stderr (or a file) to get the same information).
   2a. Some tests have been `ignore`d. In a follow-up I'll make it so that Shuttle as a whole returns a `Result`, which will include the schedule. The tests will be updated and reenabled in that PR.
   2a. Changes the "have we serialized" into a thread local `AtomicUsize`. This is needed because the panic hook is a global resource, but whether we have serializes is a thread local resource. Previously we had the two following failure modes: 1. A test fails and serializes. **Any other test failure will not serialize after this, as this test will disable the panic hook for everyone**. 2. A test panics, but then keeps going (either because the panic was caught, or as part of cleanup). The test
4. Refactors `step` to not panic directly, but instead to return an error, and then have the caller return an error. The goal here is to make the code easier to read and reason about, and also to enable Shuttle to return a `Result` in the future, which I think is a better mechanism than relying on panicking as the messaging mechanism.
5. Adds `immediately_return_on_panic` as an experimental field to the `Config`. It's experimental as it might get stabilized as a (non-exhaustive) enum which also allows prioritized scheduling. It's also experimental simply because it has not been tested enough. My testing so far is that enabling this is an improvement. I will try enabling it in our bigger tests to see if there's some issue, there might be an issue with dropping the generators and then getting a panic.
6. Adds a new `advanced` test module to house the abort tests. My opinion is that `basic` has been overpopulated, and I often look for places to house more "advanced" functionality. The hope is to move more stuff there in the future.
7. Moves code into `leave_task_span` and `enter_task_span`. It was just poor code design from me initially to not have these as separate functions. Making these functions make the code more readable.

---

### DIFF
--- a/shuttle/src/runtime/execution.rs
+++ b/shuttle/src/runtime/execution.rs
@@ -11,7 +11,7 @@ use crate::{Config, MaxSteps};
 use scoped_tls::scoped_thread_local;
 use smallvec::SmallVec;
 use std::any::Any;
-use std::cell::{RefCell, UnsafeCell};
+use std::cell::RefCell;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
@@ -42,41 +42,32 @@ thread_local! {
 
 #[derive(Debug, Default)]
 pub struct CurrentSchedule {
-    // SAFETY: There are two ways having an `UnsafeCell` could be unsafe:
-    //         1: Multiple threads are accessing the `UnsafeCell` at the same time, and at least one thread does a write.
-    //         2: We give out a reference or a pointer to the `UnsafeCell`, and that reference is used after the contents have been moved or freed.
-    //         These cannot occur, as Shuttle is single threaded, and we never give out a reference or a raw pointer.
-    current_schedule: UnsafeCell<Schedule>,
+    current_schedule: RefCell<Schedule>,
 }
 
 impl CurrentSchedule {
     fn init(schedule: Schedule) {
-        // SAFETY: Shuttle is single threaded, and we never give out references to `current_schedule`
-        unsafe { CURRENT_SCHEDULE.with(|cs| (*cs.current_schedule.get() = schedule)) }
+        CURRENT_SCHEDULE.with(|cs| *cs.current_schedule.borrow_mut() = schedule)
     }
 
     /// Add the given task ID as the next step of the schedule.
     fn push_task(tid: TaskId) {
-        // SAFETY: Shuttle is single threaded, and we never give out references to `current_schedule`
-        unsafe { CURRENT_SCHEDULE.with(|cs| (&mut *cs.current_schedule.get()).push_task(tid)) }
+        CURRENT_SCHEDULE.with(|cs| cs.current_schedule.borrow_mut().push_task(tid))
     }
 
     /// Add a choice of a random u64 value as the next step of the schedule
     fn push_random() {
-        // SAFETY: Shuttle is single threaded, and we never give out references to `current_schedule`
-        unsafe { CURRENT_SCHEDULE.with(|cs| (&mut *cs.current_schedule.get()).push_random()) }
+        CURRENT_SCHEDULE.with(|cs| cs.current_schedule.borrow_mut().push_random())
     }
 
     /// Return the number of steps in the schedule
     pub(crate) fn len() -> usize {
-        // SAFETY: Shuttle is single threaded, and we never give out references to `current_schedule`
-        unsafe { CURRENT_SCHEDULE.with(|cs| (*cs.current_schedule.get()).len()) }
+        CURRENT_SCHEDULE.with(|cs| (*cs.current_schedule.borrow()).len())
     }
 
     /// Returns a clone of the inner schedule
     pub(crate) fn get_schedule() -> Schedule {
-        // SAFETY: Shuttle is single threaded, and we never give out references to `current_schedule`
-        unsafe { CURRENT_SCHEDULE.with(|cs| (*cs.current_schedule.get()).clone()) }
+        CURRENT_SCHEDULE.with(|cs| (*cs.current_schedule.borrow()).clone())
     }
 }
 
@@ -113,7 +104,7 @@ impl Execution {
     }
 }
 
-fn backtrace_enabled() -> bool {
+pub(crate) fn backtrace_enabled() -> bool {
     std::env::var("RUST_BACKTRACE").is_ok() || std::env::var("RUST_LIB_BACKTRACE").is_ok()
 }
 
@@ -129,6 +120,17 @@ enum StepError {
     StepBoundExceeded,
     // Task panic and `config.immediately_return_on_panic` is set to `true`.
     TaskPanicEarlyReturn,
+}
+
+impl StepError {
+    fn persist_failure(&self, config: &Config) {
+        if let StepError::StepBoundExceeded = self {
+            if let MaxSteps::ContinueAfter(_) = config.max_steps {
+                return;
+            }
+        }
+        persist_failure(config);
+    }
 }
 
 impl Execution {
@@ -156,11 +158,11 @@ impl Execution {
                 match self.run_to_competion(config.immediately_return_on_panic) {
                     Ok(()) => {},
                     Err(e) => {
+                        e.persist_failure(config);
+
                         match e {
                             StepError::TaskFailure(payload) => {
-                                let task_name = ExecutionState::failing_task();
-                                persist_failure(config);
-                                eprintln!("test panicked in task '{task_name}'");
+                                eprintln!("test panicked in task '{}'", ExecutionState::failing_task());
 
                                 panic::resume_unwind(payload);
                             }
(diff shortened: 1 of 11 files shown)


shuttle/src/runtime/execution.rs
```diff
@@ -28,6 +28,57 @@ scoped_thread_local! {
...
+    static CURRENT_SCHEDULE: CurrentSchedule = CurrentSchedule::default();
+}
+
+#[derive(Debug, Default)]
+pub struct CurrentSchedule {
+    // SAFETY: There are two ways having an `UnsafeCell` could be unsafe:
+    //         1: Multiple threads are accessing the `UnsafeCell` at the same time, and at least one thread does a write.
+    //         2: We give out a reference or a pointer to the `UnsafeCell`, and that reference is used after the contents have been moved or freed.
+    //         These cannot occur, as Shuttle is single threaded, and we never give out a reference or a raw pointer.
+    current_schedule: UnsafeCell<Schedule>,
```
jorajeev: I'm not convinced we need to use `UnsafeCell` here.  Why is `RefCell` not enough?

jorajeev: It may be okay today, but someone might change something tomorrow and break the SAFETY assumptions.  In general, I'd prefer to avoid having to think about this stuff unless absolutely necessary.

shuttle/src/runtime/execution.rs
```diff
@@ -89,159 +151,176 @@ impl Execution {
...
-            // Run the test to completion
-            while self.step(config) {}
+                // Run the test to completion
+                match self.run_to_competion(config.immediately_return_on_panic) {
+                    Ok(()) => {},
+                    Err(e) => {
+                        match e {
+                            StepError::TaskFailure(payload) => {
+                                let task_name = ExecutionState::failing_task();
+                                persist_failure(config);
```
jorajeev: Move this up to before setting `task_name`

shuttle/src/runtime/execution.rs
```diff
@@ -89,159 +151,176 @@ impl Execution {
...
+                                    state
+                                    .tasks
+                                    .iter()
+                                    .filter(|t| !t.finished())
+                                    .map(|t| {
+                                        format!(
+                                            "{} (task {:?}{}{}){}",
+                                            t.name().unwrap_or_else(|| "<unknown>".to_string()),
+                                            t.id(),
+                                            if t.detached { ", detached" } else { "" },
```
jorajeev: Instead of inlining all this code here, could just define a `format_for_deadlock`

shuttle/src/runtime/execution.rs
```diff
@@ -89,159 +151,176 @@ impl Execution {
...
+                                            if t.sleeping() { ", pending future" } else { "" },
+                                            if backtrace_enabled() {
+                                                format!("\nBacktrace:\n{:#?}\n", t.backtrace)
+                                            } else {
+                                                "".into()
+                                            }
+                                        )
+                                    })
+                                    .collect::<Vec<_>>());
+                                persist_failure(config);
```
jorajeev: move this before assigning `blocked_tasks`

shuttle/src/runtime/execution.rs
```diff
@@ -89,159 +151,176 @@ impl Execution {
                 caller,
             );
 
-            // Run the test to completion
-            while self.step(config) {}
+                // Run the test to completion
+                match self.run_to_competion(config.immediately_return_on_panic) {
+                    Ok(()) => {},
+                    Err(e) => {
+                        match e {
```
jorajeev, line 163: Could define a `should_persist_failure` on `StepError` and just call

`if e.should_persist_failure() { persist_failure(config); }`

or even `e.persist_failure(config)`

before the match, instead of calling it in each branch

shuttle/src/runtime/execution.rs
```diff
@@ -89,159 +151,176 @@ impl Execution {
...
+
+                if state.config.record_steps_in_span {
+                    state.current().step_span.record("i", CurrentSchedule::len());
                 }
-            }
+            });
+        });
+    }
+
+    fn leave_task_span() {
```
jorajeev: nit: rename to `exit_task_span` to match `enter_task_span`

shuttle/src/runtime/failure.rs
```diff
@@ -4,76 +4,54 @@
...
-pub fn persist_task_failure(schedule: &Schedule, task_name: String, config: &Config, print_if_fresh: bool) -> String {
-    persist_failure(
-        schedule,
-        format!("test panicked in task '{task_name}'"),
-        config,
-        print_if_fresh,
-    )
+// When we last persisted a schedule. Used so that we don't persist the same schedule twice.
+thread_local! {
+    static SCHEDULE_PERSITED_AT: Cell<usize> = const { Cell::new(0) };
```
jorajeev: typo: "PERSISTED" misspelled

Also: why store the length and not just a boolean?

shuttle/tests/advanced/abort_freedom.rs
```diff
@@ -0,0 +1,93 @@
+use shuttle::scheduler::DfsScheduler;
+use shuttle::sync::Mutex;
+use shuttle::{Config, Runner};
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+// check_dfs with immediately_return_on_panic set to true
+pub fn check_dfs<F>(f: F)
```
jorajeev: Why is this `pub`?  And why overload `check_dfs`?

Actually, why even define a function? It's only used once here?

shuttle/tests/basic/replay.rs
```diff
@@ -29,22 +29,26 @@ fn concurrent_increment_buggy() {
 }
 
 #[test]
+#[ignore = "replay mechanism is broken because the schedule is not emitted in the panic output. reintroduce once replay mechanism is fixed."]
```
jorajeev, line 32: What's the plan here? Could we define a `replay_with_expected_panic` for these tests at least?

shuttle/tests/advanced/abort_freedom.rs
sarsko: Copy-pasted it over, then didn't think to clean it up. I agree it's better to inline the function.

shuttle/src/runtime/execution.rs
sarsko: Don't care enough about this hill; I'll update to `RefCell`. I think the false safety of `RefCell` is much worse than the explicit contract of `UnsafeCell`. The contract above is hard to accidentally break and it is clear to see when one has departed from it.

shuttle/src/runtime/execution.rs
sarsko: True

shuttle/src/runtime/failure.rs
sarsko: So that we don't persist the same schedule twice. We want to persist longer schedules. An issue now is that we persist the schedule, then keep scheduling, and never persist the longer schedule. The second issue now is that a caught panic will trigger a schedule serialization but the actual test panic which comes later will go unserialized.

shuttle/tests/basic/replay.rs
sarsko: Plan is to change Shuttle to return a `Result` and have that contain the `StepError` and the schedule. These tests can then just do assertions on the `Result` and the whole replay mechanism can be removed.

> Could we define a `replay_with_expected_panic` for these tests at least?

What does this mean?

shuttle/src/runtime/execution.rs
```diff
@@ -565,6 +620,9 @@ impl ExecutionState {
     /// its execution.
     pub(crate) fn maybe_yield() -> bool {
         Self::with(|state| {
+            if std::thread::panicking() {
+                return true;
```
dylanjwolff, line 628: I think this should be `false`. I don't think it is a good idea to be scheduling and switching tasks when panicking. Notably the intermittent CI issue with #210 is resolved by adding this check and changing it to be false

sarsko: >  I don't think it is a good idea to be scheduling and switching tasks when panicking.

We're not.

This PR changes this to return `true` and also doesn't do any more scheduling and task switching after this point. We persist the schedule and return out of the test.

Having this be `false` means we stay on the current task. This has the advantage of fully cleaning up the panic, but the disadvantage of opening up the possibility of running forever (you are explicitly turning off scheduling if you're panicking after all) and the disadvantage of opening up for test aborts (this will probably be less of an issue if we swap to CoroSensei).

I think the approach of returning `true` and persisting schedule and returning all the way out is the better approach.

---
Also note that this PR should solve the issue you're seeing because we persist schedules multiple times, meaning we'll also persist the schedule which contains the scheduling point omitted in your PR.

jorajeev (approved):

PR #203, opened 2025-08-20
Add a wrapper for Rand 0.8

sarsko:
Adds a wrapper for https://crates.io/crates/rand/0.8.5 in order to make it more ergonomic to run a codebase under Shuttle

---


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "shuttle",
+    "shuttle_rand_08",
```
jorajeev: Why are we putting `_08` in the name?

sarsko: We're not, we're putting `_08` in the path. In the future we'll have an `_08` and an `_09`, both of which have `shuttle-rand` in their `Cargo.toml`, but one publishes a mirror of rand 0.8, and one publishes a mirror of rand 0.9.

bkragl: can we make it `_0_8`?

sarsko: Why?

bkragl: clarity. if we'd ever have `_111`, is it `1.11`, `11.1`, ...?

sarsko: Oh getcha

sarsko: I'd imagine we'd change the path if it ever became ambiguous. We're currently only interested in supporting 0.8 and 0.9 and I imagine that to remain the case for a few years

jorajeev (approved):

PR #204, opened 2025-08-22
Generators --> Corosensei

dylanjwolff:
This PR switches Shuttle continuations to use [corosensei](https://docs.rs/corosensei/latest/corosensei/) rather than the generators library. In the context-switch heavy micro-benchmarks (for example, `shuttle/benches/counter.rs` with the `RandomScheduler`), this results in 5-20% faster scheduling throughput because of reduced context-switching overhead.

This PR is marked as a draft to allow for thorough testing on client projects and for discussion of the implementation. 

Major points for discussion:

1) There is an API change with `corosensei` where we now need to pass an explicit `yielder` in order to suspend a coroutine.  Because the continuations are kept in a pool and reused, this requires us to exfiltrate the yielder from the closure passed in to `Coroutine::with_stack` and persist it for the lifetime of the Continuation. Currently, the yielder is carried around as a raw pointer on the Task struct, but I suspect there might be a nicer way to do this. Unfortunately, naively embedding it in the Continuation struct is not possible because the Continuation is already borrowed by its `resume` method when `switch` is called.

2) There is an issue where `corosensei` [expects to catch a ForcedUnwind when dropping coroutines](https://github.com/Amanieu/corosensei/blob/bf7e695f7043aa3c4d56b3797102c9008ac1245e/src/coroutine.rs#L384). Something in the way Shuttle handles panics is intercepting the ForcedUnwind and not propagating it up to where corosensei expects it, causing it to panic. This only arises on an initial panic, or during cleanup if some Tasks are detached. The current solution is to call `force_reset` instead of `force_unwind` in these cases, which does not attempt to unwind the coroutine, but therefore may leak resources allocated on the coroutine. In these two scenarios, we probably don't care about leaking resources, because the whole test is about to exit. 

TODOS before this is ready for merging:

- [ ] test on some client projects over a longer period of time
- [ ] update the ignored test (depends on how we handle (1))
- [ ] general cleanup (remove unnecessary pub, delete commented out println's etc.)

---

### DIFF
--- a/shuttle/src/lib.rs
+++ b/shuttle/src/lib.rs
@@ -246,7 +246,7 @@ impl Config {
     /// Create a new default configuration
     pub fn new() -> Self {
         Self {
-            stack_size: 0x8000,
+            stack_size: 0xf000,
             failure_persistence: FailurePersistence::Print,
             max_steps: MaxSteps::FailAfter(1_000_000),
             max_time: None,
--- a/shuttle/src/runtime/thread/continuation.rs
+++ b/shuttle/src/runtime/thread/continuation.rs
@@ -1,11 +1,6 @@
-// To use the scoped version of the `generator` API, we'd need a way to store each continuation's
-// `Scope` object locally. Normally that would be TLS but those use platform threads, so aren't
-// aware of `generator` threads. Instead we just fall back to using the unscoped API.
-// TODO: upgrade to the new scoped generator API
-#![allow(deprecated)]
-
 use crate::runtime::execution::ExecutionState;
-use generator::{Generator, Gn};
+use corosensei::Yielder;
+use corosensei::{stack::DefaultStack, Coroutine, CoroutineResult};
 use scoped_tls::scoped_thread_local;
 use std::cell::{Cell, RefCell};
 use std::collections::VecDeque;
@@ -24,9 +19,10 @@ scoped_thread_local! {
 /// to run via `initialize`. A continuation is only reusable if the previous function it was
 /// executing completed.
 pub(crate) struct Continuation {
-    generator: Generator<'static, ContinuationInput, ContinuationOutput>,
+    coroutine: Coroutine<ContinuationInput, ContinuationOutput, ContinuationOutput>,
     function: ContinuationFunction,
     state: ContinuationState,
+    pub yielder: *const Yielder<ContinuationInput, ContinuationOutput>,
 }
 
 /// A cell to pass functions into continuations
@@ -41,46 +37,54 @@ unsafe impl Send for ContinuationFunction {}
 
 /// Inputs that we can pass to a continuation.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
-enum ContinuationInput {
+pub enum ContinuationInput {
     Resume,
     Exit,
 }
 
 /// Outputs that a continuation can pass back to us
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
-enum ContinuationOutput {
+pub enum ContinuationOutput {
     Yielded,
-    Finished,
+    Finished(*const Yielder<ContinuationInput, ContinuationOutput>),
     Exited,
 }
 
 /// The current state of a continuation. Lifecycle runs from top to bottom.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 enum ContinuationState {
-    NotReady, // has no function in its cell; waiting for input about what to do next
-    Ready,    // has a function in its cell; waiting for input about what to do next
-    Running,  // currently inside a user-provided function
+    NotReady,          // has no function in its cell; waiting for input about what to do next
+    Initialized,       // has a function in its cell, but hasn't started running yet
+    Ready,             // has a suspended function in its cell; waiting for input about what to do next
+    Running,           // currently inside a user-provided function
+    FinishedIteration, // has finished the previous function, can be initialized with a new one
+    Exited,            // the internal coroutine has exited its loop and cannot receive new functions to execute
 }
 
 impl Continuation {
     pub fn new(stack_size: usize) -> Self {
         let function = ContinuationFunction(Rc::new(Cell::new(None)));
 
-        let mut gen = {
+        let mut coroutine = {
             let function = function.clone();
 
-            Gn::new_opt(stack_size, move || {
+            Coroutine::with_stack(DefaultStack::new(stack_size).unwrap(), move |yielder, input| {
+                if let ContinuationInput::Exit = input {
+                    return ContinuationOutput::Exited;
+                }
+
                 // Move the whole `ContinuationFunction`, not just its field (Rust 2021 thing)
                 let _ = &function;
 
+                // eprintln!("Yielder {:?}", yielder as *const _);
                 loop {
                     // Tell the caller we've finished the previous user function (or if this is our
                     // first time around the loop, the caller below expects us to pretend we've
                     // finished the previous function).
-                    match generator::yield_(ContinuationOutput::Finished) {
-                        None | Some(ContinuationInput::Exit) => break,
-                        _ => (),
-                    }
+                    match yielder.suspend(ContinuationOutput::Finished(yielder as *const _)) {
+                        ContinuationInput::Exit => break,
+                        ContinuationInput::Resume => {}
+                    };
 
                     let f = function.0.take().expect("must have a function to run");
 
@@ -91,12 +95,16 @@ impl Continuation {
             })
         };
 
-        // Resume the generator once to get it into the loop
-        let ret = gen.resume().unwrap();
-        debug_assert_eq!(ret, ContinuationOutput::Finished);
+        // Resume the coroutine once to get it into the loop
+        let yielder = match coroutine.resume(ContinuationInput::Resume) {
+            CoroutineResult::Yield(ContinuationOutput::Finished(yielder)) => yielder,
+            _ => panic!("Coroutine should yield the yielder on first resume"),
+        };
 
+        // eprintln!("Yielder ret {:?}", yielder as *const _);
         Self {
-            generator: gen,
+            coroutine,
+            yielder,
             function,
             state: ContinuationState::NotReady,
         }
@@ -105,21 +113,17 @@ impl Continuation {
     /// Provide a new function for the continuation to execute. The continuation must
     /// be in reusable state.
     pub fn initialize(&mut self, fun: Box<dyn FnOnce()>) {
-        debug_assert_eq!(
-            self.state,
-            ContinuationState::NotReady,
-            "shouldn't replace a function before it runs"
-        );
+        debug_assert!(self.reusable(), "shouldn't replace a function before it runs");
 
         let old = self.function.0.replace(Some(fun));
         debug_assert!(old.is_none(), "shouldn't replace a function before it runs");
 
-        self.state = ContinuationState::Ready;
+        self.state = ContinuationState::Initialized;
     }
 
     /// Resume the continuation, and returns true if the function it was executing has finished.
     pub fn resume(&mut self) -> bool {
-        debug_assert!(self.state == ContinuationState::Ready || self.state == ContinuationState::Running);
+        debug_assert!(self.state == ContinuationState::Ready || self.state == ContinuationState::Initialized);
 
         let ret = self.resume_with_input(ContinuationInput::Resume);
         debug_assert_ne!(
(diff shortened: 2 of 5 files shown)


sarsko:
Okay nevermind on the loop, as per https://github.com/Amanieu/corosensei/issues/57 only the first call matters

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -41,46 +37,54 @@ unsafe impl Send for ContinuationFunction {}
 
 /// Inputs that we can pass to a continuation.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
-enum ContinuationInput {
+pub enum ContinuationInput {
```
sarsko: Why are these `pub`? Should be `pub(crate)` no?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -41,46 +37,54 @@ unsafe impl Send for ContinuationFunction {}
...
-            Gn::new_opt(stack_size, move || {
+            Coroutine::with_stack(DefaultStack::new(stack_size).unwrap(), move |yielder, input| {
+                if let ContinuationInput::Exit = input {
+                    return ContinuationOutput::Exited;
+                }
+
                 // Move the whole `ContinuationFunction`, not just its field (Rust 2021 thing)
                 let _ = &function;
 
+                // eprintln!("Yielder {:?}", yielder as *const _);
```
sarsko: Leftover

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -91,12 +95,16 @@ impl Continuation {
             })
         };
 
-        // Resume the generator once to get it into the loop
-        let ret = gen.resume().unwrap();
-        debug_assert_eq!(ret, ContinuationOutput::Finished);
+        // Resume the coroutine once to get it into the loop
+        let yielder = match coroutine.resume(ContinuationInput::Resume) {
+            CoroutineResult::Yield(ContinuationOutput::Finished(yielder)) => yielder,
+            _ => panic!("Coroutine should yield the yielder on first resume"),
```
sarsko: Please elaborate `panic` message for the person who knows nothing about the continuation stuff

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -91,12 +95,16 @@ impl Continuation {
...
+        };
 
+        // eprintln!("Yielder ret {:?}", yielder as *const _);
```
sarsko: leftover

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -105,21 +113,17 @@ impl Continuation {
     /// Provide a new function for the continuation to execute. The continuation must
     /// be in reusable state.
     pub fn initialize(&mut self, fun: Box<dyn FnOnce()>) {
-        debug_assert_eq!(
-            self.state,
-            ContinuationState::NotReady,
-            "shouldn't replace a function before it runs"
-        );
+        debug_assert!(self.reusable(), "shouldn't replace a function before it runs");
```
sarsko: Why is the assertion weakened? Isn't the "continuation lifecycle" the same as before?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -41,46 +37,54 @@ unsafe impl Send for ContinuationFunction {}
...
 enum ContinuationState {
-    NotReady, // has no function in its cell; waiting for input about what to do next
-    Ready,    // has a function in its cell; waiting for input about what to do next
-    Running,  // currently inside a user-provided function
+    NotReady,          // has no function in its cell; waiting for input about what to do next
+    Initialized,       // has a function in its cell, but hasn't started running yet
+    Ready,             // has a suspended function in its cell; waiting for input about what to do next
+    Running,           // currently inside a user-provided function
+    FinishedIteration, // has finished the previous function, can be initialized with a new one
+    Exited,            // the internal coroutine has exited its loop and cannot receive new functions to execute
```
sarsko, line 61: Why do we need `FinishedIteration` and `Exited`?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -128,24 +132,33 @@ impl Continuation {
...
-        ret
     }
 
     /// A continuation is reusable if it has completed running a user function and is waiting
     /// to resume. A continuation isn't reusable if it's still inside the user function `f`
     /// (for example, if the DFS scheduler terminated a path early, a function might not have
     /// completed, and resuming it will take us to somewhere arbitrary in user code).
     fn reusable(&self) -> bool {
-        self.state == ContinuationState::NotReady
+        self.state == ContinuationState::NotReady || self.state == ContinuationState::FinishedIteration
```
sarsko: Isn't `Exited` and `Ready` also `reusable`?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -128,24 +132,33 @@ impl Continuation {
...
+            }
+            CoroutineResult::Return(output) => {
+                self.state = ContinuationState::Exited;
+                output
+            }
         }
-        ret
     }
 
     /// A continuation is reusable if it has completed running a user function and is waiting
```
sarsko, line 157: Not sure I think this comment is great, maybe update it?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -258,8 +271,12 @@ unsafe impl Send for PooledContinuation {}
 pub(crate) fn switch() {
     crate::annotations::record_tick();
     if ExecutionState::maybe_yield() {
-        let r = generator::yield_(ContinuationOutput::Yielded).unwrap();
-        assert!(matches!(r, ContinuationInput::Resume));
+        let yielder = ExecutionState::with(|state| state.current().yielder);
+        // println!("switch @ {:?}", yielder);
```
sarsko: leftover

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -258,8 +271,12 @@ unsafe impl Send for PooledContinuation {}
...
+        match unsafe { &(*yielder) }.suspend(ContinuationOutput::Yielded) {
```
sarsko, line 288: SAFETY

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -269,6 +286,7 @@ mod tests {
     use crate::Config;
 
     #[test]
+    #[ignore = "needs to update for corosensei"]
```
sarsko: ?

shuttle/src/lib.rs
```diff
@@ -256,7 +256,7 @@ impl Config {
     /// Create a new default configuration
     pub fn new() -> Self {
         Self {
-            stack_size: 0x8000,
+            stack_size: 0xf000,
```
sarsko, line 259: ?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -199,26 +228,10 @@ impl ContinuationPool {
...
-            c.state = ContinuationState::Running;
-        }
-    }
-}
-
 /// A thin wrapper around a `Continuation` that returns it to a `ContinuationPool`
 /// when dropped, but only if it's reusable.
 pub(crate) struct PooledContinuation {
-    continuation: Option<Continuation>,
+    pub continuation: Option<Continuation>,
```
sarsko: Why has this been `pub`bed

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -199,26 +228,10 @@ impl ContinuationPool {
     }
 }
 
-impl Drop for ContinuationPool {
-    fn drop(&mut self) {
```
sarsko, line 203: Why is is fine to remove this?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -156,9 +169,24 @@ impl Drop for Continuation {
...
+        match self.state {
+            ContinuationState::Initialized | ContinuationState::FinishedIteration | ContinuationState::NotReady => {
+                let ret = self.resume_with_input(ContinuationInput::Exit);
+                debug_assert_eq!(ret, ContinuationOutput::Exited);
+            }
+            ContinuationState::Running | ContinuationState::Ready => {
+                // If already panicking or at the end of the execution, don't worry about cleaning up resources
+                // on individual coroutines which are still in-flight
+                if std::thread::panicking() {
+                    unsafe {
```
sarsko, line 191: SAFETY

shuttle/src/runtime/thread/continuation.rs
dylanjwolff: This "cheat" was necessary because of an internal implementation detail of the Generators library. As far as I can see `corosensei` does not internally use ThreadLocals in the same way. I think the "cheat" is a bit difficult to reason about (overloading a state, messing with drop behavior) and so if it is not necessary I believe it should be removed.

shuttle/src/runtime/thread/continuation.rs
dylanjwolff: It is not weakened. I created additional, more granular states because I found the old state-machine to be confusing. What was formerly `NotReady` is now represented by two more granular states:`NotReady || Finished`

shuttle/src/runtime/thread/continuation.rs
dylanjwolff: FinishedIteration is similar to NotReady, except that it has an *old* function in its cell, instead of no function at all.

Exited indicates that the coroutine has exited the inner loop and cannot be reused or resumed.

I think the new states make the state-machine easier to understand, even if they aren't strictly necessary (obviously they aren't because we got away with only three states before).

shuttle/src/lib.rs
dylanjwolff: In my testing the `corosensei` coroutines were running out of stack space occasionally. I guess there is some slight constant memory usage overhead as compared to generators

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -128,24 +133,36 @@ impl Continuation {
...
-        ret
     }
 
     /// A continuation is reusable if it has completed running a user function and is waiting
-    /// to resume. A continuation isn't reusable if it's still inside the user function `f`
-    /// (for example, if the DFS scheduler terminated a path early, a function might not have
-    /// completed, and resuming it will take us to somewhere arbitrary in user code).
+    /// to be initialized with a new one. A continuation isn't reusable if it's still inside the user
+    /// function `f` (Ready or Running), as changing it's inner function would leak the currently
+    /// executing context. We also do not consider Initialized coroutines to be reusable to avoid
```
sarsko, line 160: But surely reusing an `Initialized` coroutine is valid? It is valid for a scheduler to not schedule a spawned task thus leaving it in `Initialized`. Or am I missing something?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -289,8 +317,10 @@ mod tests {
         );
 
         let mut c = pool.acquire_inner(config.stack_size);
-        c.initialize(Box::new(|| {
-            generator::yield_with(ContinuationOutput::Yielded);
+        c.initialize(Box::new(move || {
+            // We use the yielder directly here because we are not in a Shuttle execution
+            // for `continuation::switch`, which requires access to `ExecutionState`
+            unsafe { &(*yielder) }.suspend(ContinuationOutput::Yielded);
```
sarsko, line 327: SAFETY

dylanjwolff: Do we need safety comments in test code?

shuttle/src/runtime/thread/continuation.rs
dylanjwolff: It is valid in that swapping out the function when Initialized won't break anything.

I omitted it since I couldn't think of a real use case for initializing and then immediately overwriting with a different function. 

Wrt. your propose scenario: I don't think there is a scheduling point between continuation initialization and running the continuation -- we have many tests using DFS which would trigger this scenario if that were the case.

sarsko: Hmm, maybe I'm missing something? Isn't the flow: "spawn task start, yield, spawn task finish, return JoinHandle", and at that point the continuation is initialized but hasn't been run?

shuttle/src/runtime/thread/continuation.rs
sarsko: I didn't see that we were in test code, but I generally think yes, both because it is good to show that one has thought about it and is able to give the argument for why its fine, and because people like me (or code review bots) sometimes look for `unsafe` without SAFETY comment and make a fuzz about it.

shuttle/src/runtime/thread/continuation.rs
sarsko: Also re what is `reusable`: it is not super important, but I'm generally pro having functions be correct, as in return `true` iff the continuation can be reused, and then enforce "state machine" invariants elsewhere (ie do a `debug_assert` that we're not reusing an `Initialized` continuation if that's what we wanna enforce)

shuttle/src/lib.rs
dylanjwolff: per the comment on the longer benchmarking time with backtraces, this is probably because corosensei produces deeper callstacks than generators

PR #205, opened 2025-08-26
Print task backtraces on deadlock

sarsko:
Prints a backtrace for each task on failure. Uses `RUST_BACKTRACE`/`RUST_LIB_BACKTRACE` to decide whether to capture backtraces (as this is somewhat expensive), which is the same signaling mechanism used by `Backtrace::capture()`.

---

### DIFF
--- a/shuttle/src/runtime/execution.rs
+++ b/shuttle/src/runtime/execution.rs
@@ -61,6 +61,10 @@ impl Execution {
     }
 }
 
+fn backtrace_enabled() -> bool {
+    std::env::var("RUST_BACKTRACE").is_ok() || std::env::var("RUST_LIB_BACKTRACE").is_ok()
+}
+
 impl Execution {
     /// Run a function to be tested, taking control of scheduling it and any tasks it might spawn.
     /// This function runs until `f` and all tasks spawned by `f` have terminated, or until the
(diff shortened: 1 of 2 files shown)


shuttle/src/runtime/execution.rs
```diff
@@ -125,14 +129,25 @@ impl Execution {
...
+                                    if backtrace_enabled() {
+                                        format!("\nBacktrace:\n{:#?}\n", t.backtrace)
+                                    } else {
+                                        "".into()
+                                    }
                                 )
                             })
                             .collect::<Vec<_>>();
+
+                        // Collecting backtraces is expensive, so we only want to do it if the use opts in to collecting them.
```
jorajeev: nit: "use" -> "user"

shuttle/src/runtime/task/mod.rs
```diff
@@ -233,6 +237,7 @@ impl Task {
             span_stack,
             local_storage: StorageMap::new(),
             tag: None,
+            backtrace: Backtrace::disabled(),
```
jorajeev, line 329: The documentation on `Backtrace::disabled` is super confusing, because it states "Forcibly captures a disabled backtrace, regardless of environment variable configuration."

But that seems like a cut-n-paste error, because that's not what the implementation does: https://doc.rust-lang.org/beta/src/std/backtrace.rs.html#319-321
🤷‍♂️

jorajeev (dismissed):

shuttle/src/runtime/task/mod.rs
sarsko: Yeah I also noticed that

PR #206, opened 2025-08-26
Record task names in the step span

sarsko:
Logs task names in the step span

---


shuttle/src/current.rs
```diff
@@ -82,6 +82,17 @@ pub fn get_name_for_task(task_id: TaskId) -> Option<TaskName> {
...
+    // Do note that `record` simply appends the new name as a field, meaning the step span will end up looking something like this:
+    // step{task="main-thread(2)" task="Child"}
+    // when running with something like the `tracing_subscriber::fmt` subscriber.
+    // This either has to be lived with, or the task name should be set via the `ChildLabelFn` mechanism, or a different subscriber should be used
+    // (if this is done, then `record_steps_in_span` should be set to true as well), or Shuttle will have to be chanegd to recreate the Span
+    ExecutionState::try_with(|state| {
+        state
+            .get_mut(task_id)
+            .step_span
+            .record("task", format!("{:?}", task_name));
```
jorajeev, line 94: Why not `.record("task-name", ...)`.  That is, why call it "task"?

sarsko: In part because the field already was named `task`, and in part because when the name is not set the id is listed.

jorajeev (approved):

PR #207, opened 2025-08-27
Stable Synchronization-Resource Identifiers Across Executions

dylanjwolff:
This PR adds resource "signatures", which allows the same synchronization primitives to be identified within and across Shuttle iterations. These signatures can be used for debugging with multiple different schedules and also are a prerequisite for many scheduling algorithms such as [Partial Order Sampling](https://link.springer.com/content/pdf/10.1007/978-3-319-96142-2_20.pdf). The signatures are similar conceptually to `Task` signatures: https://github.com/awslabs/shuttle/pull/196

Each synchonization primitive now carries with it a `TypedResourceSignature` enum which represents its type and contains the `ResourceSignature`. The `ResourceSignature` itself is a combination of (1) the static source location where the resource was created and (2) the parent task signature hash and (3) a per-task counter which is incremented every time the parent task creates a resource at that location. We eagerly hash the `ResourceSignature` to allow for fast comparisons between resources during scheduling. Signatures are kept on the underlying `BatchSempahore` for higher level resources such as `Mutex`.

Note that some `#[allow(unused)]` annotations are necessary because the signatures aren't yet being consumed anywhere outside of unit tests. They will be removed in the near future when we provide information about the `next_event` on the `Task` struct for scheduling algorithms.

---

### DIFF
--- a/shuttle/src/sync/mod.rs
+++ b/shuttle/src/sync/mod.rs
@@ -22,6 +22,7 @@ pub use rwlock::RwLock;
 pub use rwlock::RwLockReadGuard;
 pub use rwlock::RwLockWriteGuard;
 
+use std::fmt::{Debug, Display};
 use std::hash::{Hash, Hasher};
 use std::panic::Location;
 pub use std::sync::{LockResult, PoisonError, TryLockError, TryLockResult};
@@ -50,7 +51,7 @@ pub(crate) enum ResourceType {
 }
 
 #[allow(unused)]
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub(crate) struct ResourceSignature {
     resource_type: ResourceType,
     static_create_location: &'static Location<'static>,
@@ -103,6 +104,39 @@ impl ResourceSignature {
     /// context where the task was spawned.
     pub(crate) fn signature_hash(&self) -> u64 {
         self.signature_hash
+    }
+}
+
+impl Debug for ResourceSignature {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ResourceSignature")
+            .field("type", &self.resource_type)
+            .field("static_create_location", &self.static_create_location)
+            .field("parent_task_signature", &self.parent_task_signature)
+            .field("create_location_counter", &self.create_location_counter)
+            .finish()
+    }
+}
+
+impl Display for ResourceSignature {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let suffix = match self.create_location_counter {
+            1 => "st",
+            2 => "nd",
+            3 => "rd",
+            _ => "th",
+        };
+        write!(
+            f,
+            "{:?}[ {:?}{}@{}:{:?}:{:?} on {} ]",
+            self.resource_type,
+            self.create_location_counter,
+            suffix,
+            self.static_create_location.file(),
+            self.static_create_location.line(),
+            self.static_create_location.column(),
+            self.parent_task_signature,
+        )
     }
 }
 


shuttle/Cargo.toml
```diff
@@ -26,6 +26,7 @@ tracing = { version = "0.1.36", default-features = false, features = ["std"] }
 regex = { version = "1.10.6", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { version = "1.0", optional = true }
+const-siphasher = "1.0.2"
```
dylanjwolff, line 29: If we don't want to take on a new dependency I can implement a `const` hasher myself

shuttle/src/runtime/task/mod.rs
```diff
@@ -204,6 +205,16 @@ impl TaskSignature {
         }
     }
 
+    pub(crate) fn new_resource(&mut self, static_create_location: &'static Location<'static>) -> ResourceSignature {
```
sarsko: A general heads up which I forgot about on your previous PR as well: from the docs for [`caller`](https://doc.rust-lang.org/std/panic/struct.Location.html#method.caller): `Returns the source location of the caller of this function. If that function’s caller is annotated then its call location will be returned, and so on up the stack to the first call within a non-tracked function body.`, AKA: I don't think we have to pass the `Location` everywhere, we could just `#[track_caller]` all the way to the public API.

shuttle/Cargo.toml
sarsko: I'm fine with adding a dependency

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,117 @@ pub use rwlock::RwLock;
...
+    Condvar(ResourceSignature),
+    Mutex(ResourceSignature),
+    RwLock(ResourceSignature),
+    Once(ResourceSignature),
+    MpscChannel(ResourceSignature),
+}
+
+#[derive(Clone, Debug)]
+/// A unique signature for identifying resources in Shuttle tests.
+/// This is used internally to track and differentiate between different resource instances.
```
sarsko: Move doc comment above `derive`

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,117 @@ pub use rwlock::RwLock;
...
+pub(crate) struct ResourceSignature {
```
sarsko, line 55: This is not the scheme I expected when I saw `TypedResourceSignature` and `ResourceSignature` in the imports above (I expected them to both be instances of resource signatures, not one to be the metadata of the other). Maybe refactor such that `TypedResourceSignature` -> `ResourceSignature` and `ResourceSignature` -> `Signature` or something like that?

shuttle/src/sync/atomic/mod.rs
```diff
@@ -235,4 +239,76 @@ impl<T: Copy + Eq> Atomic<T> {
...
+            atomic_u16.signature(),
+            atomic_u32.signature(),
+            atomic_u64.signature(),
+            atomic_usize.signature(),
+            atomic_bool.signature(),
+            atomic_ptr.signature(),
+        ];
+
+        // Check all signatures are unique
+        for i in 0..signatures.len() {
```
sarsko: Just compare length of the set with the length of the vec no?

shuttle/src/sync/atomic/mod.rs
```diff
@@ -235,4 +239,76 @@ impl<T: Copy + Eq> Atomic<T> {
...
+                let atomic1 = AtomicBool::new(false);
+                let atomic2 = AtomicBool::new(true);
+
+                all_signatures_clone.lock().unwrap().insert(atomic1.signature());
+                all_signatures_clone.lock().unwrap().insert(atomic2.signature());
+            },
+            10,
+        );
+
+        // Should have exactly 3 unique signatures across all iterations
```
sarsko: Typo: 3 -> 2

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,117 @@ pub use rwlock::RwLock;
...
+    static_create_location: &'static Location<'static>,
+    parent_task_signature: u64,
+    create_location_counter: u32,
+    static_create_location_hash: u64,
+    signature_hash: u64,
+}
+
+const fn hash_bytes<'a>(hasher: &'a mut SipHasher, bytes: &'static [u8], index: usize) -> &'a mut SipHasher {
```
sarsko: Huh surely this exists in the `SipHasher` api?

shuttle/src/runtime/task/mod.rs
sarsko: Please refactor to do this

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,117 @@ pub use rwlock::RwLock;
...
+        Self {
+            static_create_location,
+            parent_task_signature: CONST_CONTEXT_SIGNATURE,
+            static_create_location_hash,
+            create_location_counter: 1,
+            signature_hash,
+        }
+    }
+
+    pub(crate) fn new(
```
sarsko: What's the non-const part? Just always use the SipHasher and compact into one function no?

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,117 @@ pub use rwlock::RwLock;
...
+
+    /// Combined signature of the static location and dynamic context
+    /// context where the task was spawned.
+    pub(crate) fn signature_hash(&self) -> u64 {
+        self.signature_hash
+    }
+}
+
+impl Hash for ResourceSignature {
+    fn hash<H: Hasher>(&self, state: &mut H) {
```
sarsko, line 144: Surely hash should just be the `signature_hash` no?

shuttle/src/sync/mod.rs
dylanjwolff: some resources have non-const constructors. in these cases, they could be created in a loop etc. so we need a non-const constructor to give us unique signatures

shuttle/src/sync/mod.rs
dylanjwolff: they are the same. the signature is created by the hash function in the constructor. I can change it to directly return the signature if you'd prefer -- as of right now, I think we don't have any use case where we are hashing lots of signatures and it would matter

shuttle/src/sync/mod.rs
dylanjwolff: I cannot believe I didn't see it. It's because all of the others are `write_<T>` and the bytes API is just `hash`

shuttle/src/sync/mod.rs
dylanjwolff: changed to ResourceSignatureData

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,108 @@ pub use rwlock::RwLock;
...
+use std::panic::Location;
 pub use std::sync::{LockResult, PoisonError, TryLockError, TryLockResult};
 
 // TODO implement true support for `Arc`
 pub use std::sync::{Arc, Weak};
+
+const CONST_CONTEXT_SIGNATURE: u64 = 0;
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub(crate) enum TypedResourceSignature {
```
sarsko: Why is it called `TypedResourceSignature` and not just `ResourceSignature`?

shuttle/src/sync/mod.rs
sarsko: I am not following, afaict the only part about this function which is not const is the hasher? So change it to use the SipHasher and have `new` be const and `new_const` be a wrapper on that ?

shuttle/src/sync/mod.rs
dylanjwolff: no particular reason, happy to change the name

shuttle/src/sync/mod.rs
dylanjwolff: the non-const version has two additional arguments:

```
        parent_task_signature: u64,
        create_location_counter: u32
```

these represent the *runtime context* in which the resource was created (which task created the resource and how many resources has that Task created from that source location). This information is thus not available at compile time.

If we make everything use a constructor without these arguments, then the ResourceSignatures for two different resources created in a loop will be the same, even though the objects themselves are different.

sarsko: I get all that, but I still don't see why the function cannot be `const`? The function signature would be:

```rust
pub(crate) const fn new(static_create_location: &'static Location<'static>, parent_task_signature: u64, create_location_counter: u32)
```

And then the function currently called `new_const` would call `new(Location::caller(), CONST_CONTEXT_SIGNATURE, 1)`

dylanjwolff: what is the benefit of making this const? is it just to avoid duplicating functionality in both versions? 

in practice, when you call this function, it should always depend on runtime information. if it doesn't depend on runtime information, you should use `new_const`. IMO making this `const` makes the API more confusing and that doesn't seem worth shaving a few lines of similar code in `new_const` and `new` to me.

sarsko: It's to avoid duplicating the implementation.

>in practice, when you call this function, it should always depend on runtime information. 

I get what you mean, but we already know this is not true, as `new_const` can be implemented on top of `new`.

> MO making this `const` makes the API more confusing

`const` is a contract on the implementation of the function (namely that the body can be evaluated at compile time), and doesn't say anything about how the function should be used. Obviously if you are not able to get the parameters to call the function at compile time then you will have to evaluate the function at runtime.


If you want to keep new non-const, then it can be implemented as such:

```rust
pub(crate) const fn new_const() -> Self {
    new_inner(Location::caller(), CONST_CONTEXT_SIGNATURE, 1)
}

pub(crate) fn new(static_create_location: &'static Location<'static>, parent_task_signature: u64, create_location_counter: u32) -> Self {
     new_inner(static_create_location, parent_task_signature, create_location_counter)
}

const fn new_inner(static_create_location: &'static Location<'static>, parent_task_signature: u64, create_location_counter: u32) -> Self {
	...
}

```

dylanjwolff: I've made the change so that code is shared for both constructors and updated the doc-comment to state that const constructed signatures are not unique

shuttle/src/future/batch_semaphore.rs
```diff
@@ -323,25 +326,55 @@ impl std::error::Error for AcquireError {}
...
+    #[track_caller]
     pub fn new(num_permits: usize, fairness: Fairness) -> Self {
+        Self::new_internal(
+            num_permits,
+            fairness,
+            ResourceSignature::BatchSemaphore(ExecutionState::new_resource_signature()),
+        )
+    }
+
+    pub(crate) fn new_internal(num_permits: usize, fairness: Fairness, signature: ResourceSignature) -> Self {
```
jorajeev: nit: would name this `new_with_signature`

shuttle/src/future/batch_semaphore.rs
```diff
@@ -323,25 +326,55 @@ impl std::error::Error for AcquireError {}
...
+    #[track_caller]
     pub const fn const_new(num_permits: usize, fairness: Fairness) -> Self {
+        Self::const_new_internal(
+            num_permits,
+            fairness,
+            ResourceSignature::BatchSemaphore(ResourceSignatureData::new_const()),
+        )
+    }
+
+    pub(crate) const fn const_new_internal(
```
jorajeev: nit: `const_new_with_signature`

shuttle/src/future/batch_semaphore.rs
```diff
@@ -596,8 +629,13 @@ unsafe impl Send for BatchSemaphore {}
 unsafe impl Sync for BatchSemaphore {}
 
 impl Default for BatchSemaphore {
+    #[track_caller]
     fn default() -> Self {
-        Self::new(Default::default(), Fairness::StrictlyFair)
+        Self::new_internal(
```
jorajeev: Why did the body of this method change?

shuttle/src/future/batch_semaphore.rs
```diff
@@ -752,3 +790,50 @@ impl crate::annotations::WithName for BatchSemaphore {
...
+                let sem1 = BatchSemaphore::new(1, Fairness::StrictlyFair);
+                let sem2 = BatchSemaphore::new(2, Fairness::Unfair);
+
+                all_signatures_clone.lock().unwrap().insert(sem1.signature().clone());
+                all_signatures_clone.lock().unwrap().insert(sem2.signature().clone());
+            },
+            10,
+        );
+
+        // Should have exactly 2 unique signatures across all iterations
```
jorajeev: Could we assert the stronger condition that sem1 and sem2 keep their signatures across iterations?  Could we store (number_of_available_permits, signature) in the HashSet instead?  Since you initialize them with different #permits?

shuttle/src/runtime/task/mod.rs
```diff
@@ -204,6 +205,18 @@ impl TaskSignature {
         }
     }
 
+    #[track_caller]
+    pub(crate) fn new_resource(&mut self) -> ResourceSignatureData {
+        let static_create_location = Location::caller();
+        let counter = self
+            .child_counters
```
jorajeev, line 212: Should we use separate counters for spawned children and spawned resources?

shuttle/src/sync/atomic/mod.rs
```diff
@@ -235,4 +240,75 @@ impl<T: Copy + Eq> Atomic<T> {
...
+        let atomic_u8 = AtomicU8::new(0);
+        let atomic_u16 = AtomicU16::new(0);
+        let atomic_u32 = AtomicU32::new(0);
+        let atomic_u64 = AtomicU64::new(0);
+        let atomic_usize = AtomicUsize::new(0);
+        let atomic_bool = AtomicBool::new(false);
+        let atomic_ptr = AtomicPtr::new(std::ptr::null_mut::<i32>());
+
+        // All atomics should have unique signatures
+        let signatures = vec![
```
jorajeev: Why not just create a HashSet of the signatures and assert size of set == 12 ?

shuttle/src/sync/atomic/mod.rs
```diff
@@ -235,4 +240,75 @@ impl<T: Copy + Eq> Atomic<T> {
...
+                let atomic1 = AtomicBool::new(false);
+                let atomic2 = AtomicBool::new(true);
+
+                all_signatures_clone.lock().unwrap().insert(atomic1.signature());
+                all_signatures_clone.lock().unwrap().insert(atomic2.signature());
+            },
+            10,
+        );
+
+        // Should have exactly 2 unique signatures across all iterations
```
jorajeev: Could do a stronger check by loading the atomic and distinguishing between them since they carry two different boolean values

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,99 @@ pub use rwlock::RwLock;
...
+const CONST_CONTEXT_SIGNATURE: u64 = 0;
+
+/// A stable signature for identifying resources in Shuttle tests.
+/// This is used internally to track and differentiate between different resource instances
+/// *across* shuttle iterations. This signature hash is not *guaranteed* to be unique (two
+/// different resources may have the same hash). In particular, resources which provide
+/// `const` constructors cannot provide any *runtime* context to their signatures, so any
+/// two resources created at the same source location will collide.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub(crate) enum ResourceSignature {
```
jorajeev: Instead of having this enum, how about having instead

`enum ResourceSignatureType {
    Atomic,
    BatchSemaphore,
    Barrier,
    ...
}`

and then making `ResourceSignature` a wrapper around `ResourceSignatureType` and `ResourceSignatureData`.

This would also make it possible to use separate counters per resource type when generating the signatures, so when debugging, you would know "this is the 3rd RwLock created by this task"

shuttle/src/runtime/task/mod.rs
dylanjwolff: I don't think there is any reason to. Resources and tasks cannot be created at the same source `(file, line, column)`. Might as well keep them all together

shuttle/src/future/batch_semaphore.rs
dylanjwolff: yup, this did not need to be changed. this was a holdover from before I found out `#[track_caller]` is transitive

shuttle/src/sync/mod.rs
dylanjwolff: I don't have a strong preference between an enum containing a struct or a struct containing an enum. I think it doesn't really make a large difference either way.

> This would also make it possible to use separate counters per resource type when generating the signatures, so when debugging, you would know "this is the 3rd RwLock created by this task"

Actually, the current implementation gives even more information than this -- if the counter is 3 for a RWLock resource, that means, for example, "this is the 3rd RwLock created by this task *at that source location*". 

That being said, a lot of the constructors for resources are `const` constructors. We actually don't get accurate counters for these cases anyways because there can be no parent task or counters in a `const fn`.

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,99 @@ pub use rwlock::RwLock;
...
+    Barrier(ResourceSignatureData),
+    Condvar(ResourceSignatureData),
+    Mutex(ResourceSignatureData),
+    RwLock(ResourceSignatureData),
+    Once(ResourceSignatureData),
+    MpscChannel(ResourceSignatureData),
+}
+
+#[allow(unused)]
+#[derive(Clone, Debug)]
```
jorajeev: Do we need to include the `static_create_location_hash` and the `signature_hash` in the `Debug` output? Those are derived from the other fields.

shuttle/src/sync/mod.rs
jorajeev: If we ever want to have schedulers (or monitors) that want to look at the resource type, they have to pass around the entire `ResourceSignature`.  Having a separate `ResourceType` (better name than `ResourceSignatureType`, which I suggested earlier) would allow passing just that around.

Not a big deal, since you can match on either, but just seems clearer to me to separate the type of resource, and the unique identifier to it.

dylanjwolff: ab21caf539

shuttle/src/sync/mod.rs
dylanjwolff: I've added manual implementations of both Display and Debug (19c21d57), example outputs below

Display:
`Mutex[ 1st@shuttle/tests/basic/clocks.rs:40:24 on 0 ]`

Debug:
`ResourceSignature { type: Mutex, static_create_location: Location { file: "shuttle/tests/basic/clocks.rs", line: 40, column: 24 }, parent_task_signature: 0, create_location_counter: 1 } `

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,136 @@ pub use rwlock::RwLock;
...
+const CONST_CONTEXT_SIGNATURE: u64 = 0;
+
+/// A stable signature for identifying resources in Shuttle tests.
+/// This is used internally to track and differentiate between different resource instances
+/// *across* shuttle iterations. This signature hash is not *guaranteed* to be unique (two
+/// different resources may have the same hash). In particular, resources which provide
+/// `const` constructors cannot provide any *runtime* context to their signatures, so any
+/// two resources created at the same source location will collide.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub(crate) enum ResourceType {
```
bkragl, line 42: Seems like this should be the doc comment on `ResourceSignature`.

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,136 @@ pub use rwlock::RwLock;
...
+    Condvar,
+    Mutex,
+    RwLock,
+    Once,
+    MpscChannel,
+}
+
+#[allow(unused)]
+#[derive(Clone)]
+pub(crate) struct ResourceSignature {
```
bkragl, line 55: Are we planning to expose task and resource signatures in the future?

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,136 @@ pub use rwlock::RwLock;
...
+
+    pub(crate) const fn new(
+        resource_type: ResourceType,
+        static_create_location: &'static Location<'static>,
+        parent_task_signature: u64,
+        create_location_counter: u32,
+    ) -> Self {
+        let mut hasher = SipHasher::new();
+        let file: &'static str = static_create_location.file();
+        hasher.hash(file.as_bytes());
```
bkragl, line 79: What's the "scope of portability" we want for signatures? I'm thinking that on different machines the file location can be different, we don't fix endianness, etc.

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,136 @@ pub use rwlock::RwLock;
...
+        hasher.write_u32(static_create_location.line());
+        hasher.write_u32(static_create_location.column());
+        let static_create_location_hash = hasher.finish();
+        hasher.write_u64(static_create_location_hash);
```
bkragl, line 83: Why are we sticking `static_create_location_hash` back into the hasher? `finish` does not reset the hasher state.

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,136 @@ pub use rwlock::RwLock;
...
+        };
+        write!(
+            f,
+            "{:?}[ {:?}{}@{}:{:?}:{:?} on {} ]",
+            self.resource_type,
+            self.create_location_counter,
+            suffix,
+            self.static_create_location.file(),
+            self.static_create_location.line(),
+            self.static_create_location.column(),
```
bkragl, line 137: nit: could just use the `Display` implementation of `Location`

shuttle/src/sync/mod.rs
```diff
@@ -21,7 +22,136 @@ pub use rwlock::RwLock;
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.static_create_location_hash().hash(state);
+        self.parent_task_signature.hash(state);
+        self.create_location_counter.hash(state);
+    }
+}
+
+impl PartialEq for ResourceSignature {
+    fn eq(&self, other: &Self) -> bool {
+        self.signature_hash() == other.signature_hash()
```
bkragl, line 153: So we consider the hash as the "identity" of the signature, and don't care about collisions? I guess there is not much point since the parent info is just a hash.

shuttle/src/sync/mod.rs
dylanjwolff: wow that is very counterintuitive to me that `finish` doesn't reset hasher state. good catch! I'll put together a cleanup PR for this and some of your other comments. thanks for the feedback!

shuttle/src/sync/mod.rs
dylanjwolff: This is a fair point -- schedules produced by schedulers which use these signatures *potentially* won't be replayable (from seeds) across different machines.

It's not clear to me that there is an easy/good way to solve this, and I think in practice it isn't a big issue; there is always the total ordering replay mode to fall-back on.

shuttle/src/sync/mod.rs
dylanjwolff: Unfortunately there can be collisions of identical signatures for different resources even ignoring hash collisions because some resources are const constructed.

I am actually leaning towards un-implementing Eq/PartialEq in the follow up PR for the Signatures to communicate that they cannot naively be compared. Instead it is up to the caller/use case to decide what constitutes equality wrt. signatures.

PR #208, opened 2025-08-27
Fix: make Once thread safe

sarsko:
Been seeing a few of the following issues after #188:

```
thread '<unnamed>' panicked at /local/TodWorker-nobody/testruns/57c7ab17-05bc-4f8a-b6b2-d92ed4639d1a/env/brazil-rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell/once.rs:296:66:
reentrant init
stack backtrace:
   0:     0x55ea5b824f92 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h67e4df710c8f676f
   1:     0x55ea5b84ff93 - core::fmt::write::hda312f67a39af703
   2:     0x55ea5b820c13 - std::io::Write::write_fmt::h8e8a8b40c8cd6968
   3:     0x55ea5b824de2 - std::sys::backtrace::BacktraceLock::print::h9ec8308d922515db
   4:     0x55ea5b82686c - std::panicking::default_hook::{{closure}}::hd92428497bc2c49c
   5:     0x55ea5b82666f - std::panicking::default_hook::hc6676cbc3d2b1b51
   6:     0x55ea5b7dc086 - shuttle::runtime::failure::init_panic_hook::{{closure}}::{{closure}}::h46ac9a410e94a15e
   7:     0x55ea5b827363 - std::panicking::rust_panic_with_hook::h8132dac0081efe2c
   8:     0x55ea5b826fe6 - std::panicking::begin_panic_handler::{{closure}}::h090bfbc752a41996
   9:     0x55ea5b825489 - std::sys::backtrace::__rust_end_short_backtrace::h66453f58cb486cef
  10:     0x55ea5b826cad - __rustc[d0726284cd839be0]::rust_begin_unwind
  11:     0x55ea5b84cef0 - core::panicking::panic_fmt::h8a4c917d9814701b
  12:     0x55ea5b7dd4ca - core::cell::once::OnceCell<T>::try_init::h00f67c7b1b66c84e
  13:     0x55ea5b7e317f - shuttle:runtime::execution::ExecutionState::get_storage::h21ea056bd9457d2f
  14:     0x55ea556f4b69 - scoped_tls::ScopedKey<T>::with::hf5d9b9da4b91b4f3
  15:     0x55ea5ae2afd3 - shuttle::sync::once::Once::call_once_inner::h1c72d955e106d601
  16:     0x55ea59eb001f - shuttle::lazy_static::Lazy<T>::get::h5aa4ee4b1f3deaae
  17:     0x55ea58db91b7 - FUNCTION_WHICH_ACCESSES_LAZY_STATIC
  18:     0x55ea58a71168 - shuttle::future::block_on::h927b353e98ba51e0
  19:     0x55ea564817ea -  shuttle::thread::thread_fn::hbffc719e845f61c1
  20:     0x55ea5b7dc903 - generator::stack::StackBox<F>::call_once::hb5e423b7998a3b0d
  21:     0x55ea5b7ee19b - generator::detail::gen::gen_init_impl::h86c106ff57f2fbed
  22:     0x55ea5b7ee0d9 - generator::detail::asm::gen_init::hb78bc15cd5f139cf
  23:                0x0 - <unknown>
test panicked in task 'main-thread
```

I think it is due to `Lazy<T>` being built upon `Once`, and #188 made `Once` no longer thread safe. Thus this change makes `Once` actually thread safe buy building it around `LazyLock` instead of `OnceCell`.

---

### DIFF
--- a/shuttle/src/sync/once.rs
+++ b/shuttle/src/sync/once.rs
@@ -2,9 +2,10 @@ use crate::runtime::execution::ExecutionState;
 use crate::runtime::storage::StorageKey;
 use crate::runtime::task::clock::VectorClock;
 use crate::sync::Mutex;
-use std::cell::{OnceCell, RefCell};
+use std::cell::RefCell;
 use std::rc::Rc;
 use std::sync::atomic::{AtomicUsize as StdAtomicUsize, Ordering};
+use std::sync::LazyLock;
 use tracing::trace;
 
 /// A synchronization primitive which can be used to run a one-time global initialization. Useful
@@ -12,8 +13,24 @@ use tracing::trace;
 /// with [`Once::new()`].
 #[derive(Debug)]
 pub struct Once {
+    // Note that we need to use a `LazyLock` here (or similar) so that tests don't interfere with each other when multiple
+    // tests try to interact with a `Once` at the same time. This happens in our `Lazy` implementation, and also happens
+    // whenever there is a `static Once`, such as in the following:
+    // ```
+    // static O: Once = Once::new();
+    //
+    // #[test]
+    // fn foo1() {
+    //     check_dfs(|| O.call_once(|| {}), None);
+    // }
+    //
+    //  #[test]
+    // fn foo2() {
+    //     check_dfs(|| O.call_once(|| {}), None);
+    // }
+    // ```
     /// Unique identifier for this [`Once`], used as a key for the [`OnceInitState`] for this instance.
-    id: OnceCell<usize>,
+    id: LazyLock<usize>,
 }
 
 /// A `Once` cell can either be `Running`, in which case a `Mutex` mediates racing threads trying to
@@ -33,19 +50,20 @@ impl std::fmt::Debug for OnceInitState {
     }
 }
 
-// Safety: Once is never actually passed across true threads, only across continuations. The
-// OnceCell type therefore can't be preempted mid-bookkeeping-operation.
-// TODO we shouldn't need to do this, but OnceCell is not Send, and anything we put within a Once
-// TODO needs to be Send.
-unsafe impl Send for Once {}
-unsafe impl Sync for Once {}
-
 impl Once {
     /// Creates a new `Once` value.
     #[must_use]
     #[allow(clippy::new_without_default)]
     pub const fn new() -> Self {
-        Self { id: OnceCell::new() }
+        static NEXT_ID: StdAtomicUsize = StdAtomicUsize::new(1);
+
+        Self {
+            id: LazyLock::new(|| {
+                let id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
+                assert_ne!(id, 0, "id overflow");
+                id
+            }),
+        }
     }
 
     /// Performs an initialization routine once and only once. The given closure will be executed
@@ -150,11 +168,7 @@ impl Once {
     }
 
     fn id(&self) -> usize {
-        static NEXT_ID: StdAtomicUsize = StdAtomicUsize::new(1);
-
-        let id = *self.id.get_or_init(|| NEXT_ID.fetch_add(1, Ordering::Relaxed));
-        assert_ne!(id, 0, "id overflow");
-        id
+        *self.id
     }
 
     fn get_state<'a>(&self, from: &'a ExecutionState) -> Option<&'a RefCell<OnceInitState>> {


sarsko:
I reproed the issue (pre fix) with the following:

```rust
struct Mt {}

impl Mt {
    fn new() -> Self {
        println!("yup");
        Mt {}
    }

    fn m(&self) {
        println!("m");
    }
}

lazy_static! {
    static ref MEOW: Mt = Mt::new();
}

#[test]
fn s12() {
    check_dfs(|| MEOW.m(), None);
}

#[test]
fn s123() {
    check_dfs(
        || {
            MEOW.m();
        },
        None,
    );
}
```
Which by running a few times gives the following:
```
cargo test s12 -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running unittests src/lib.rs (target/debug/deps/shuttle-42d4eb921e9ef897)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.00s

     Running tests/mod.rs (target/debug/deps/mod-da9801a4d256efc5)

running 2 tests
2025-08-28T17:35:38.107900Z  INFO execution{i=0}:new_task{parent=None i=0}: shuttle::runtime::task: created task task_id=main-thread(0) signature=406307610674748927 static_create_location=12665329965471106439
2025-08-28T17:35:38.107909Z  INFO execution{i=0}:new_task{parent=None i=0}: shuttle::runtime::task: created task task_id=main-thread(0) signature=406307610674748927 static_create_location=12665329965471106439

thread 'future::basic::s12' panicked at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/cell/once.rs:296:66:
reentrant init
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
yup
m
WARNING: Shuttle runs the `Drop` method of `lazy_static` values at the end of an execution, unlike the actual `lazy_static` implementation. This difference may cause false positives. See https://docs.rs/shuttle/*/shuttle/lazy_static/index.html#warning-about-drop-behavior for details or to disable this warning.
2025-08-28T17:35:38.108845Z ERROR execution{i=0}:step{task="main-thread(0)"}: generator::detail::gen: set panic inside generator    
2025-08-28T17:35:38.108894Z  INFO shuttle::scheduler::metrics: run finished iterations=1 steps=[min=3, max=3, avg=3.0] context_switches=[min=0, max=0, avg=0.0] preemptions=[min=0, max=0, avg=0.0] random_choices=[min=0, max=0, avg=0.0]
test future::basic::s123 ... ok
test panicked in task 'main-thread'
failing schedule:
"
910101f8acd1910100
"
pass that string to `shuttle::replay` to replay the failure
2025-08-28T17:35:38.109442Z  INFO shuttle::scheduler::metrics: run finished iterations=1 steps=[min=1, max=1, avg=1.0] context_switches=[min=0, max=0, avg=0.0] preemptions=[min=0, max=0, avg=0.0] random_choices=[min=0, max=0, avg=0.0]
test future::basic::s12 ... FAILED

failures:

failures:
    future::basic::s12

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 410 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p shuttle --test mod`
```

shuttle/src/sync/once.rs
```diff
@@ -150,9 +148,7 @@ impl Once {
     }
 
     fn id(&self) -> usize {
-        static NEXT_ID: StdAtomicUsize = StdAtomicUsize::new(1);
-
-        let id = *self.id.get_or_init(|| NEXT_ID.fetch_add(1, Ordering::Relaxed));
+        let id = *self.id;
         assert_ne!(id, 0, "id overflow");
```
bkragl: If we stick with this solution, can we move the assertion right after `fetch_add`?

bkragl (commented):
The issue is that any declared lazy_static desugars to a single static `Lazy`, with a singly underlying `Once`, with a single underlying `OnceCell`, that independent Shuttle executions in different threads unsafely race on initializing. This change makes the race safe. Note that it is fine for a static `Once` to get a "global" ID for all Shuttle executions, since the ID is used to index into execution-local storage where the actual `Once` state lives, so there is no overlap.

I don't love that we need to use `LazyLock` due to this particular use case, but I'm also not sure it there is an easy other solution. If we go with this, we should document all of the above.

shuttle/src/sync/once.rs
sarsko: I dunno if I think it's better to have this inside the closure passed to `LazyLock`. This is right after the `fetch_add`

bkragl: The thing is that we execute the assertion every time `id` is called, which is unnecessary.

PR #210, opened 2025-09-03
Changing scheduling points to always *precede* visible operations

dylanjwolff:
Currently, the scheduling point placement in Shuttle is somewhat ad-hoc. Atomics have scheduling points both before and after operations, most other primitives have scheduling points after the operation has concluded, but then there are some special cases for `BatchSemaphore` to allow `try_acquire`s to fail. Notably, this inconsistency likely contributed to some completeness issues with `Channel` because of missing scheduling points (addressed in this PR).

This PR makes all such scheduling point placements consistent. By ensuring each visible operation has a scheduling point before it, there will be a potential context switch in between each visible operation, which is a necessary condition for Shuttle's completeness. Putting these scheduling points before the operation, rather than after, allows scheduling algorithms to decide what to do proactively based on what operations are “in-flight”. In contrast, if the scheduling points are only after each operation, the algorithms can only be retroactive, making decisions based only on what has already happened.

However, placing scheduling points before each operation naively can lead to yielding twice on blocking operations (once before the operation, then again when it blocks). 

As an optimization, the switch before the blocking operation can be conditionally omitted while maintaining completeness iff (1) the operation will block in the current context and (2) if the act of blocking itself does not affect other tasks. (1) ensures that there will still be a context switch to make the previous operation visible. (2) ensures that there is no loss of completeness in the current operation. For example, a blocking Channel send satisfies (1), but does not satisfy (2) if there are not already other blocking senders on the channel — prior to the sender blocking, try_recv will fail, but after the sender blocks, try_recv succeeds. Thus the act of blocking itself is a visible operation, meaning that we need a scheduling point before blocking for complete exploration of all possible behaviors. The checks (1) and (2) are done dynamically during execution by checking the state of the corresponding primitive.

---

### DIFF
--- a/shuttle/src/future/batch_semaphore.rs
+++ b/shuttle/src/future/batch_semaphore.rs
@@ -360,6 +360,8 @@ impl BatchSemaphore {
     /// Closes the semaphore. This prevents the semaphore from issuing new
     /// permits and notifies all pending waiters.
     pub fn close(&self) {
+        thread::switch();
+
         self.init_object_id();
         let mut state = self.state.borrow_mut();
         if state.closed {
@@ -402,6 +404,8 @@ impl BatchSemaphore {
     /// If the semaphore is closed, returns `Err(TryAcquireError::Closed)`
     /// If there aren't enough permits, returns `Err(TryAcquireError::NoPermits)`
     pub fn try_acquire(&self, num_permits: usize) -> Result<(), TryAcquireError> {
+        thread::switch();
+
         self.init_object_id();
         let mut state = self.state.borrow_mut();
         let id = state.id.unwrap();
@@ -432,12 +436,6 @@ impl BatchSemaphore {
         }
 
         crate::annotations::record_semaphore_try_acquire(id, num_permits, res.is_ok());
-
-        // We context switch here whether we acquired any permits or not. If
-        // we have, this is to let other threads fail their `try_acquire`;
-        // if we have not, we yield so that the current thread can try again
-        // after other threads have worked.
-        thread::switch();
 
         res
     }
@@ -508,18 +506,22 @@ impl BatchSemaphore {
 
     /// Acquire the specified number of permits (async API)
     pub fn acquire(&self, num_permits: usize) -> Acquire<'_> {
+        // No switch here; switch should be triggered on polling future
         self.init_object_id();
         Acquire::new(self, num_permits)
     }
 
     /// Acquire the specified number of permits (blocking API)
     pub fn acquire_blocking(&self, num_permits: usize) -> Result<(), AcquireError> {
+        // No switch here; switch should be triggered on polling future
         self.init_object_id();
         crate::future::block_on(self.acquire(num_permits))
     }
 
     /// Release `num_permits` back to the Semaphore
     pub fn release(&self, num_permits: usize) {
+        thread::switch();
+
         self.init_object_id();
         if num_permits == 0 {
             return;
@@ -582,9 +584,6 @@ impl BatchSemaphore {
             }
         }
         drop(state);
-
-        // Releasing a semaphore is a yield point
-        thread::switch();
     }
 }
 
@@ -608,6 +607,7 @@ pub struct Acquire<'a> {
     waiter: Arc<Waiter>,
     semaphore: &'a BatchSemaphore,
     completed: bool, // Has the future completed yet?
+    has_polled: bool,
 }
 
 impl<'a> Acquire<'a> {
@@ -617,6 +617,7 @@ impl<'a> Acquire<'a> {
             waiter,
             semaphore,
             completed: false,
+            has_polled: false,
         }
     }
 }
@@ -626,6 +627,22 @@ impl Future for Acquire<'_> {
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         assert!(!self.completed);
+        let will_succeed = self.waiter.has_permits.load(Ordering::SeqCst)
+            || self.semaphore.is_closed()
+            || self.semaphore.available_permits() >= self.waiter.num_permits;
+
+        let blocking_changes_state = self.semaphore.fairness == Fairness::StrictlyFair;
+        // If the acquire will succeed on the first try, we need to context switch once to allow the previous
+        // event to become visible. If we won't succeed, then we still need to context switch if the act of
+        // blocking is a visible operation. This is true for fair semaphores because the order of the waiter
+        // queue is affected by blocking. In the case of unfair semaphores, if the semaphore has no permits
+        // available, the extra waiter doesn't affect other tasks -- all waiters and active tasks will race
+        // to acquire permits when the current holder releases.
+        if !self.has_polled && (will_succeed || blocking_changes_state) {
+            thread::switch();
+        }
+        self.has_polled = true;
+
         if self.waiter.has_permits.load(Ordering::SeqCst) {
             assert!(!self.waiter.is_queued.load(Ordering::SeqCst));
             self.completed = true;
@@ -700,8 +717,6 @@ impl Future for Acquire<'_> {
                         // threads that can no longer succeed.
                         self.semaphore.reblock_if_unfair();
 
-                        // Yield so other threads can fail a `try_acquire`.
-                        thread::switch();
                         Poll::Ready(Ok(()))
                     }
                     Err(TryAcquireError::NoPermits) => {
--- a/shuttle/src/future/mod.rs
+++ b/shuttle/src/future/mod.rs
@@ -28,8 +28,6 @@ where
     let inner = Arc::new(std::sync::Mutex::new(JoinHandleInner::default()));
     let task_id = ExecutionState::spawn_future(Wrapper::new(fut, inner.clone()), stack_size, None, caller);
 
-    thread::switch();
-
     JoinHandle { task_id, inner }
 }
 
@@ -242,15 +240,20 @@ pub fn block_on<F: Future>(future: F) -> F::Output {
     let waker = ExecutionState::with(|state| state.current_mut().waker());
     let cx = &mut Context::from_waker(&waker);
 
+    // Note: we only switch on poll pending, since this blocks the current task. This means that *internal*
+    // Shuttle futures which do not use other Shuttle primitives such as `batch_semaphore::Acquire` must
+    // have a scheduling point prior to first poll if that poll will be successful and can affect other tasks.
+    // For example, an uncontested acquire makes other threads block or fail try-acquires, so there must be
+    // a scheduling point for scheduling completeness. For *external* futures, this is a non-issue because they
+    // should use other Shuttle primitives inside of `poll` if polling can affect other threads.
     loop {
         match future.as_mut().poll(cx) {
             Poll::Ready(result) => break result,
             Poll::Pending => {
                 ExecutionState::with(|state| state.current_mut().sleep_unless_woken());
+                thread::switch();
             }
         }
-
-        thread::switch();
     }
 }
 
--- a/shuttle/src/runtime/execution.rs
+++ b/shuttle/src/runtime/execution.rs
@@ -3,6 +3,7 @@ use crate::runtime::storage::{StorageKey, StorageMap};
 use crate::runtime::task::clock::VectorClock;
 use crate::runtime::task::labels::Labels;
 use crate::runtime::task::{ChildLabelFn, Task, TaskId, TaskName, TaskSignature, DEFAULT_INLINE_TASKS};
+use crate::runtime::thread;
 use crate::runtime::thread::continuation::PooledContinuation;
 use crate::scheduler::{Schedule, Scheduler};
 use crate::thread::thread_fn;
@@ -360,6 +361,36 @@ impl ExecutionState {
         Self::with(|s| s.current().id())
     }
 
+    /// If there is only one attached, unfinished task and there is at least one detached, unfinished task
+    /// then exiting the attached task will truncate the events on the detached task(s)
+    pub(crate) fn exit_current_truncates_execution(&self) -> bool {
+        // Strictly speaking, this is only true if there are other runnable detached tasks, but always making the main thread
+        // exit a scheduling point is simpler conceptually
+        if self.current().id() == TaskId::from(0) {
+            return true;
+        }
+
+        // If the current task is detached, then it definitely doesn't truncate the execution
+        if self.current().is_detached() {
+            return false;
+        }
+
+        let mut single_unfinished_attached = false;
+        let mut has_unfinished_detached = false;
+        for t in self.tasks.iter() {
+            let unfinished_attached = !t.finished() && !t.detached;
+            if single_unfinished_attached && unfinished_attached {
+                // there are more than one unfinished attached tasks, so one exiting won't truncate
+                return false;
+            } else {
+                single_unfinished_attached = true;
+            }
+
+            has_unfinished_detached |= !t.finished() && t.detached;
+        }
+        has_unfinished_detached && single_unfinished_attached
+    }
+
     fn set_labels_for_new_task(state: &ExecutionState, task_id: TaskId, name: Option<String>) {
         LABELS.with(|cell| {
             let mut map = cell.borrow_mut();
@@ -441,6 +472,7 @@ impl ExecutionState {
     where
         F: Future<Output = ()> + 'static,
     {
+        thread::switch();
         let task_id = Self::with(|state| {
             let schedule_len = state.current_schedule.len();
             let parent_span_id = state.top_level_span.id();
@@ -483,6 +515,7 @@ impl ExecutionState {
         mut initial_clock: Option<VectorClock>,
         caller: &'static Location<'static>,
     ) -> TaskId {
+        thread::switch();
         let task_id = Self::with(|state| {
             let parent_span_id = state.top_level_span.id();
             let task_id = TaskId(state.tasks.len());
--- a/shuttle/src/runtime/task/mod.rs
+++ b/shuttle/src/runtime/task/mod.rs
@@ -436,6 +436,10 @@ impl Task {
         self.state == TaskState::Finished
     }
 
+    pub(crate) fn is_detached(&self) -> bool {
+        self.detached
+    }
+
     pub(crate) fn detach(&mut self) {
         self.detached = true;
     }
@@ -564,6 +568,9 @@ impl Task {
         self.local_storage.pop()
     }
 
+    pub(crate) fn park_token_is_available(&self) -> bool {
+        self.park_state.token_available
+    }
     /// Park the task if its park token is unavailable. If the task blocks, then it will be woken up
     /// when the token becomes available or spuriously without consuming the token (see the
     /// documentation for [`std::thread::park`], which says that "it may also return spuriously,
--- a/shuttle/src/runtime/thread/continuation.rs
+++ b/shuttle/src/runtime/thread/continuation.rs
@@ -11,7 +11,9 @@ use std::cell::{Cell, RefCell};
 use std::collections::VecDeque;
 use std::ops::Deref;
 use std::ops::DerefMut;
+use std::panic::Location;
 use std::rc::Rc;
+use tracing::debug;
 
 scoped_thread_local! {
     pub(crate) static CONTINUATION_POOL: ContinuationPool
@@ -255,8 +257,29 @@ impl std::fmt::Debug for PooledContinuation {
 unsafe impl Send for PooledContinuation {}
 
 /// Possibly yield back to the executor to perform a context switch.
+/// This function should be called *before* any visible operation.
+/// If each visible operation has a scheduling point before it, then there will
+/// be a potential context switch *in between* each visible operation, which
+/// is a necessary condition for Shuttle's completeness.
+///
+/// Putting scheduling points before visible operations, rather than after, has the
+/// advantage of giving the scheduling algorithm *maximum information* to make scheduling
+/// decisions based on what is about-to-happen on each task. The disadvantage of this
+/// approach is that it can lead to double-yields for blocking operations, explained below.
+///
+/// Blocking operations will result in an additional switch when the current thread blocks.
+/// As an optimization, the switch *before* the blocking operation can be conditionally
+/// omitted to avoid switching twice for the same operation iff (1) the operation *will*
+/// block in the current context and (2) if the act of blocking does not affect other tasks.
+/// For example, a blocking Channel send does not satisfy (2) if there are not already
+/// other blocking senders on the channel -- prior to the sender blocking, `try_recv` will
+/// fail, but after the sender blocks, `try_recv` succeeds. Thus the act of blocking itself
+/// is a visible operation, meaning that both scheduling points are necessary for complete
+/// exploration of all possible behaviors.
+#[track_caller]
 pub(crate) fn switch() {
     crate::annotations::record_tick();
+    debug!("switch from {}", Location::caller());
     if ExecutionState::maybe_yield() {
         let r = generator::yield_(ContinuationOutput::Yielded).unwrap();
         assert!(matches!(r, ContinuationInput::Resume));
--- a/shuttle/src/sync/barrier.rs
+++ b/shuttle/src/sync/barrier.rs
@@ -93,6 +93,14 @@ impl Barrier {
 
     /// Blocks the current thread until all threads have rendezvoused here.
     pub fn wait(&self) -> BarrierWaitResult {
+        let state = self.state.borrow_mut();
+        let will_block = state.waiters.len() < state.bound;
+        drop(state);
+        // If all tasks have already rendezvoused, we need to context switch once to allow the previous
+        // event to become visible. Otherwise it will become visible when we block
+        if !will_block {
+            thread::switch();
+        }
         let mut state = self.state.borrow_mut();
         let my_epoch = state.epoch;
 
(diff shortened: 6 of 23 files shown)


shuttle/src/sync/mpsc.rs
```diff
@@ -166,10 +157,38 @@ impl<T> Channel<T> {
...
         //    this is a rendezvous channel and there are no waiting receivers
-        let sender_should_block =
-            is_full || !state.waiting_senders.is_empty() || (is_rendezvous && state.waiting_receivers.is_empty());
+        is_full || !state.waiting_senders.is_empty() || (is_rendezvous && state.waiting_receivers.is_empty())
+    }
+
+    fn send_internal(&self, message: T, can_block: bool) -> Result<(), TrySendError<T>> {
+        let mut should_block = self.sender_must_block();
+        let blocking_send_changes_state =
+            self.state.borrow().waiting_receivers.is_empty() || self.state.borrow().waiting_senders.is_empty();
```
dylanjwolff: This might just be a tautology. If so then I think the double-yield optimization is not possible for Channels

shuttle/src/sync/mpsc.rs
```diff
@@ -253,7 +272,28 @@ impl<T> Channel<T> {
...
+        // The receiver should block in any of the following situations:
+        //    the channel is empty
+        //    there are waiting receivers
+        state.messages.is_empty() || !state.waiting_receivers.is_empty()
+    }
+
     fn recv_internal(&self, can_block: bool) -> Result<T, TryRecvError> {
+        let mut should_block = self.receiver_must_block();
+        let blocking_recv_changes_state =
+            self.state.borrow().waiting_receivers.is_empty() || self.state.borrow().waiting_senders.is_empty();
```
dylanjwolff: same as for sender, I'm not sure if this is ever false

shuttle/src/sync/mutex.rs
```diff
@@ -60,6 +61,9 @@ impl<T: ?Sized> Mutex<T> {
             drop(state);
 
             self.semaphore.acquire_blocking(1).unwrap();
+        } else {
+            // we always need to allow for a context switch to make the previous event visible for completeness
+            thread::switch();
```
dylanjwolff: This may not be necessary. If the semaphore is closed then a failed acquire probably isn't really a visible operation.

shuttle/src/sync/rwlock.rs
```diff
@@ -209,6 +210,9 @@ impl<T: ?Sized> RwLock<T> {
             drop(state);
 
             self.semaphore.acquire_blocking(typ.num_permits()).unwrap();
+        } else {
+            // we always need to allow for a context switch to make the previous event visible for completeness
+            thread::switch();
         }
```
dylanjwolff, line 222: same as for mutex, maybe not necessary to have a switch here

shuttle/src/thread.rs
```diff
@@ -282,16 +284,26 @@ unsafe impl<T> Sync for JoinHandle<T> {}
 impl<T> JoinHandle<T> {
     /// Waits for the associated thread to finish.
     pub fn join(self) -> Result<T> {
-        ExecutionState::with(|state| {
+        // The switch before joining ensures that the preceding operation on the joiner is visible to be returned by the joinee
+        let will_block = !ExecutionState::with(|state| state.get(self.task_id).finished());
+        if !will_block {
+            thread::switch();
+        }
```
dylanjwolff: As discussed with @sarsko, `Join` probably isn't actually a visible operation -- it can't affect any other tasks. As such, we can maybe omit the switch here

shuttle/tests/basic/condvar.rs
```diff
@@ -277,6 +277,7 @@ fn check_producer_consumer_broken1() {
     check_random(producer_consumer_broken1, 5000)
 }
 
+#[ignore]
```
dylanjwolff, line 280: tests which replay from a seed have been `#[ignore]`ed, as any further changes to this PR might cause them to break again. once the PR is approved, I will regenerate all the seeds for these tests

shuttle/tests/basic/dfs.rs
```diff
@@ -49,82 +49,28 @@ fn trivial_two_threads() {
...
-///   "SarA" -> "SarAf" [label="f"];
-///   "SarAf" -> "SarAfR" [label="R"];
-///   "SarAfR" -> "SarAfRF" [label="F"];
-///   "Sar" -> "Sarf" [label="f"];
-///   "Sarf" -> "SarfA" [label="A"];
-///   "SarfA" -> "SarfAR" [label="R"];
-///   "SarfAR" -> "SarfARF" [label="F"];
-/// }
+/// isareIARE
+/// isarIeARE
```
dylanjwolff, line 60: I can generate the graph in the prior version of this comment once the PR is approved. Until then modifications to the PR could change the graph, so I'd prefer not to

shuttle/tests/basic/mpsc.rs
```diff
@@ -546,7 +578,6 @@ fn mpsc_try_recv_iter_sync() {
                 let mut acc = 0;
                 for x in rx.try_iter() {
                     acc += x;
-                    thread::yield_now();
```
dylanjwolff, line 549: in general, `yield_now` is probably a code-smell for our tests, as they can mask completeness issues. I've added comments as to why they are necessary in some of the tests for this PR.

shuttle/tests/basic/mpsc.rs
```diff
@@ -567,8 +597,11 @@ fn mpsc_try_recv_iter_sync() {
         None,
     );
     let observed_values = Arc::try_unwrap(observed_values).unwrap().into_inner().unwrap();
-    // Could fail to rendezvous at any step of the sequence
-    assert_eq!(observed_values, HashSet::from([0, 3, 4, 6]));
+    // Could fail to rendezvous at any step of the sequence (0, 3, 4, 6)
+    //
+    // Additionally, the second send can fail if the first send has woken the receiver, but has not been yet read by the woken receiver (5)
+    // The same is true for the third send, but this also leads to (4)
+    assert_eq!(observed_values, HashSet::from([0, 3, 4, 6, 5]));
```
dylanjwolff, line 604: this appears to have been a completeness issue with the old Channel implementation

shuttle/src/thread.rs
```diff
@@ -213,6 +210,11 @@ where
 {
     let ret = f();
 
+    if ExecutionState::with(|s| s.exit_current_truncates_execution()) {
+        // Exiting the last attached task can truncate the execution. To make the previous
+        // event visible before truncation, we need a scheduling point before exiting.
+        thread::switch();
+    }
```
dylanjwolff, line 221: I believe that we only need to make the previous operation visible on exit if the exit truncates other tasks.

The other case to consider is ensuring that the previous operation is visible before the return value is available, but this should be true naturally for the context switch when the thread finishes, but hasn't been joined yet.

shuttle/src/future/mod.rs
```diff
@@ -242,15 +240,20 @@ pub fn block_on<F: Future>(future: F) -> F::Output {
     let waker = ExecutionState::with(|state| state.current_mut().waker());
     let cx = &mut Context::from_waker(&waker);
 
+    // Note: we only switch on poll pending, since this blocks the current task. This means that *internal*
+    // Shuttle futures which do not use other Shuttle primitives such as `batch_semaphore::Acquire` must
+    // have a scheduling point prior to first poll if that poll will be successful and can affect other tasks.
+    // For example, an uncontested acquire makes other threads block or fail try-acquires, so there must be
+    // a scheduling point for scheduling completeness. For *external* futures, this is a non-issue because they
+    // should use other Shuttle primitives inside of `poll` if polling can affect other threads.
```
sarsko, line 248: Not following this note at all.

Also don't see why we don't just always `switch`. It matches developers mental model (namely that any `await` point can be a switch point) and is correct (even though for instance tokio will happily starve an application on always-ready futures.

shuttle/src/runtime/task/mod.rs
```diff
@@ -564,6 +568,9 @@ impl Task {
         self.local_storage.pop()
     }
 
+    pub(crate) fn park_token_is_available(&self) -> bool {
```
sarsko: Nit: move `is` to start of name?

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -255,8 +257,29 @@ impl std::fmt::Debug for PooledContinuation {
...
+/// block in the current context and (2) if the act of blocking does not affect other tasks.
+/// For example, a blocking Channel send does not satisfy (2) if there are not already
+/// other blocking senders on the channel -- prior to the sender blocking, `try_recv` will
+/// fail, but after the sender blocks, `try_recv` succeeds. Thus the act of blocking itself
+/// is a visible operation, meaning that both scheduling points are necessary for complete
+/// exploration of all possible behaviors.
+#[track_caller]
 pub(crate) fn switch() {
     crate::annotations::record_tick();
+    debug!("switch from {}", Location::caller());
```
sarsko: Should this be `debug!` or `trace!`? @jorajeev

shuttle/src/future/batch_semaphore.rs
```diff
@@ -626,6 +627,22 @@ impl Future for Acquire<'_> {
...
+            || self.semaphore.available_permits() >= self.waiter.num_permits;
+
+        let blocking_changes_state = self.semaphore.fairness == Fairness::StrictlyFair;
+        // If the acquire will succeed on the first try, we need to context switch once to allow the previous
+        // event to become visible. If we won't succeed, then we still need to context switch if the act of
+        // blocking is a visible operation. This is true for fair semaphores because the order of the waiter
+        // queue is affected by blocking. In the case of unfair semaphores, if the semaphore has no permits
+        // available, the extra waiter doesn't affect other tasks -- all waiters and active tasks will race
+        // to acquire permits when the current holder releases.
+        if !self.has_polled && (will_succeed || blocking_changes_state) {
```
sarsko: I think I am in camp "don't be smart, always yield even if the block is gonna yield as well".  In my head the nature of the `switches` are different ; the initial one is to invite maximum concurrency and to tell the scheduler, the second is a consequence of being blocked. @jorajeev @dylanjwolff

shuttle/src/runtime/execution.rs
```diff
@@ -360,6 +361,36 @@ impl ExecutionState {
         Self::with(|s| s.current().id())
     }
 
+    /// If there is only one attached, unfinished task and there is at least one detached, unfinished task
+    /// then exiting the attached task will truncate the events on the detached task(s)
+    pub(crate) fn exit_current_truncates_execution(&self) -> bool {
```
sarsko: Huh what does truncating the events on the detached tasks mean?

dylanjwolff: we terminate the shuttle iteration when the last attached task exits. as a result, any detached tasks which are still alive will not finish their executions, thus they are truncated

shuttle/src/future/mod.rs
dylanjwolff: `poll` is only visible if (1) it changes global state on first `poll` or (2) it resolves to `Pending`, causing the current task to be blocked.

For (1), user-code that changes global state in the `poll` of a future should use some other shuttle primitive to do so (for example, write to an atomic or acquire a mutex). Since these primitives all introduce their own context switches, we don't need an additional switch before the `poll`. But Shuttle primitives themselves that are futures and change state without using another primitive need to introduce a scheduling point before the state change.

shuttle/src/future/batch_semaphore.rs
```diff
@@ -660,6 +661,13 @@ impl Future for Acquire<'_> {
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         assert!(!self.completed);
+
+        // Always switch at least once
+        if !self.has_polled {
+            thread::switch_task();
+        }
+        self.has_polled = true;
```
sarsko: Nit: this can be moved into the `if`

shuttle/src/sync/barrier.rs
```diff
@@ -98,6 +98,11 @@ impl Barrier {
 
     /// Blocks the current thread until all threads have rendezvoused here.
     pub fn wait(&self) -> BarrierWaitResult {
+        let state = self.state.borrow_mut();
+        drop(state);
```
sarsko: Huh? borrow state, drop state, switch, borrow state again? Why are we borrowing then dropping?

shuttle/src/sync/condvar.rs
```diff
@@ -134,18 +134,20 @@ impl Condvar {
     pub fn wait<'a, T>(&self, guard: MutexGuard<'a, T>) -> LockResult<MutexGuard<'a, T>> {
         let me = ExecutionState::me();
 
+        // Release the lock, which allows for a switch *before* unlocking, but not after
+        let mutex = guard.unlock();
```
sarsko, line 143: I feel like this warrants a bigger comment, as the correctness of `wait` relies on switching before

shuttle/src/future/batch_semaphore.rs
```diff
@@ -642,6 +641,7 @@ pub struct Acquire<'a> {
     waiter: Arc<Waiter>,
     semaphore: &'a BatchSemaphore,
     completed: bool, // Has the future completed yet?
+    has_polled: bool,
```
jorajeev: nit: maybe the negation `never_polled` would be clearer (this flag means "has polled at least once")

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -255,8 +257,20 @@ impl std::fmt::Debug for PooledContinuation {
 unsafe impl Send for PooledContinuation {}
 
 /// Possibly yield back to the executor to perform a context switch.
-pub(crate) fn switch() {
+/// This function should be called *before* any visible operation.
+/// If each visible operation has a scheduling point before it, then there will
+/// be a potential context switch *in between* each visible operation, which
```
jorajeev: nit: `*in between* any pair of visible operations`

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -255,8 +257,20 @@ impl std::fmt::Debug for PooledContinuation {
...
+/// is a necessary condition for Shuttle's completeness.
+///
+/// Putting scheduling points before visible operations, rather than after, has the
+/// advantage of giving the scheduling algorithm *maximum information* to make scheduling
```
jorajeev: nit: replace `maximum information` -> `additional information`

(there's no reasonable metric by which this is `maximum`)

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -255,8 +257,20 @@ impl std::fmt::Debug for PooledContinuation {
...
+/// decisions based on what is about-to-happen on each task. The disadvantage of this
```
jorajeev: nit: no hyphenation in `about-to-happen`

shuttle/src/sync/barrier.rs
jorajeev: Agree with Sarek - this needs more explanation

shuttle/src/sync/condvar.rs
jorajeev: Would it be equivalent to just call `thread::switch_task` on line 138, and then `guard.unlock` on line 150?

shuttle/src/sync/mpsc.rs
```diff
@@ -260,7 +270,19 @@ impl<T> Channel<T> {
...
+        // The receiver should block in any of the following situations:
+        //    the channel is empty
+        //    there are waiting receivers
+        state.messages.is_empty() || !state.waiting_receivers.is_empty()
+    }
+
     fn recv_internal(&self, can_block: bool) -> Result<T, TryRecvError> {
+        thread::switch_task();
+
+        let should_block = self.receiver_must_block();
```
jorajeev: Why are we assigning `should_block` here, when we only need it on line 331?  (And since we only use the variable once, why even introduce the local, and not just write `if self.receiver_must_block() { .. }`

shuttle/src/sync/rwlock.rs
jorajeev: Don't understand this remark? The comment in the code on line 220 gives a reason for the switch?

shuttle/src/thread.rs
```diff
@@ -205,14 +209,7 @@ where
...
 
-/// Body of a Shuttle thread, that runs the given closure, handles thread-local destructors, and
-/// stores the result of the thread in the given lock.
-pub(crate) fn thread_fn<F, T>(f: F, result: std::sync::Arc<std::sync::Mutex<Option<Result<T>>>>)
-where
-    F: FnOnce() -> T,
-{
-    let ret = f();
-
+pub(crate) fn thread_fn_cleanup<T>(ret: T, result: std::sync::Arc<std::sync::Mutex<Option<Result<T>>>>) {
```
jorajeev: nit: a better name might be `thread_fn_inner` ? Not sure what the `_cleanup` suggests.

Actually, why do we need three variants of `thread_fn_*`? Why not just pass `switch_before_exit` as an argument to `thread_fn` ?

shuttle/tests/basic/barrier.rs
```diff
@@ -79,7 +79,7 @@ fn barrier_test(n: usize, c: usize) {
 
 #[test]
 fn barrier_test_many_ok() {
-    check_dfs(|| barrier_test(4, 4), None);
+    check_dfs(|| barrier_test(3, 3), None);
```
jorajeev, line 82: Why did we reduce this number? Do the tests take much longer now?

shuttle/tests/basic/dfs.rs
```diff
@@ -45,86 +45,32 @@ fn trivial_two_threads() {
         });
     }
 
-    assert_eq!(iterations.load(Ordering::SeqCst), 2);
+    assert_eq!(iterations.load(Ordering::SeqCst), 3);
```
jorajeev, line 48: Can you enumerate the 3 iterations? Unclear why this changed

shuttle/tests/basic/dfs.rs
```diff
@@ -45,86 +45,32 @@ fn trivial_two_threads() {
...
 }
 
 /// We have two threads T0 and T1 with the following lifecycle (with letter denoting each step):
-/// * T0: spawns T1 (S), acquires lock (A), releases lock (R), finishes (F)
-/// * T1:                acquires lock (a), releases lock (r), finishes (f)
+/// * T0: (i) initializes, spawns T1 (s), acquires lock (a), releases lock (r), finishes (e)
```
jorajeev: nit: to be consistent, letter should come after description in all cases

`T0: initializes (i), spawns T1 (s), ..`

shuttle/tests/basic/dfs.rs
jorajeev: Which is the one interleaving we don't get now? (We went from 16 to 15.)

shuttle/tests/basic/dfs.rs
```diff
@@ -155,7 +101,7 @@ fn two_threads() {
     }
 
     // See `two_threads_work` for an illustration of all 16 interleavings.
-    assert_eq!(iterations.load(Ordering::SeqCst), 16);
+    assert_eq!(iterations.load(Ordering::SeqCst), 29);
```
jorajeev, line 104: Why did this jump to 29? Is this due to double switch which the follow-up PR fixes?  What's the number with that in place?

shuttle/tests/basic/dfs.rs
```diff
@@ -296,7 +242,7 @@ fn yield_loop_three_threads() {
         });
     }
 
-    assert_eq!(iterations.load(Ordering::SeqCst), 50050);
+    assert_eq!(iterations.load(Ordering::SeqCst), 378378);
```
jorajeev, line 245: That's a big jump.  What's the number without double switching?

shuttle/tests/basic/execution.rs
```diff
@@ -204,7 +203,7 @@ fn context_switches_atomic() {
                 thread.join().unwrap();
             }
 
-            assert_eq!(current::context_switches(), EXPECTED_CONTEXT_SWITCHES);
+            assert!(current::context_switches() <= EXPECTED_CONTEXT_SWITCHES);
```
jorajeev, line 206: Can we not assert an exact equality here?

shuttle/tests/basic/thread.rs
```diff
@@ -12,11 +12,12 @@ fn thread_yield_point() {
     let success = Arc::new(AtomicU8::new(0));
     let success_clone = Arc::clone(&success);
 
-    // We want to see executions that include both threads running first, otherwise we have
-    // messed up the yieldpoints around spawn.
+    // We want to see executions that include both threads running first.
+    // Spawn no longer itself produces a switch, but accessing any Shuttle primitive will allow
+    // either thread to run it's visible operations first.
```
jorajeev: nit: "it's" -> "its"

shuttle/tests/basic/thread.rs
```diff
@@ -12,11 +12,12 @@ fn thread_yield_point() {
...
     check_random(
         move || {
-            let flag = Arc::new(AtomicBool::new(false));
+            let flag = Arc::new(shuttle::sync::atomic::AtomicBool::new(false));
```
jorajeev, line 20: Why did we change to using Shuttle `AtomicBool` here?

shuttle/src/sync/rwlock.rs
dylanjwolff:
```
1. operation A
2. closed semaphore acquire
3. operation B
```

Both of these comments refer to the above situation. there is always a scheduling point before (1) and in between (2) and (3). The question is whether there needs to be a scheduling point between (1) and (2). 

If we omit the scheduling point between (1) and (2), this means it is impossible to observe any schedules where something happens in between (1) and (2) *if the semaphore was already closed when we started to do (1)*. 

The question is whether a failed acquire on a closed semaphore can be affected by anything else. For example, if closed semaphores can be reopened, omitting the scheduling point makes it impossible to observe `1 reopen 2 3`.

shuttle/tests/basic/dfs.rs
dylanjwolff: the jump to 29 is because of the double switch which will be fixed in the follow up PR. The number is 15 in that one

shuttle/tests/basic/dfs.rs
dylanjwolff: without double switching this one stays the same

shuttle/tests/basic/thread.rs
dylanjwolff: in prior versions of shuttle, *after* a spawn there is a scheduling point. In the new changes, this scheduling point is *before* the spawn. As a result, if the parent thread has no *visible* operations, it will run through to completion before the child thread can be run.

By switching to a shuttle atomic, we guarantee at least one context switch on the main thread before it begins doing irreversible work (flag.load).

shuttle/src/sync/condvar.rs
dylanjwolff: The mutex guard unlock calls batch semaphore release internally when it drops itself:

```
impl<T: ?Sized> Drop for MutexGuard<'_, T> {
    fn drop(&mut self) {
        // Release a permit (this is a yield point)
        self.mutex.semaphore.release(1);
```

... which has a switch before doing anything visible:

```
    /// Release `num_permits` back to the Semaphore
    pub fn release(&self, num_permits: usize) {
        thread::switch_task();
```

> Would it be equivalent to just call thread::switch_task on line 138, and then guard.unlock on line 150?

No this would be a double-yield

shuttle/src/sync/mpsc.rs
dylanjwolff: in the follow up PR which does the double-yield optimization, we need to check if the receiver/sender should block twice, once at entry and once at 331. this is just a vestige from the PR before the optimizations were removed

shuttle/tests/basic/barrier.rs
dylanjwolff: yes, if we don't reduce the size of some of the DFS tests they wont complete in any reasonable amount of time

dylanjwolff: with the double-yield optimizations this is not necessary

shuttle/tests/basic/dfs.rs
dylanjwolff: let's move this discussion to the later PR, since without the double-yield optimizations the number is actually 29.

I'll look into it in the meantime, I'm quite certain it is something either thread creation or thread destruction, but it's a bit difficult because the definitions of events are slightly different

shuttle/tests/basic/dfs.rs
dylanjwolff: again, without the double-yield optimizations we have several places where there are extra scheduling points, which result in extra interleavings. 

This one is three because there is one spawn and two exits. however, the child thread exiting should not be visible because it is detached

shuttle/tests/basic/thread.rs
dylanjwolff: this is also what the comment is saying at lines 15-17, let me know if you have suggestions to communicate that more clearly

shuttle/tests/basic/execution.rs
dylanjwolff: The variability is related to `join`. There are two possible context switches for `join`:

1. The scheduling point before starting to `join`. 
2. When joining, if the thread being joined has not finished, the current thread must block (causing another context switch).

Thus in some schedules where the the thread being joined is ready, we only see (1). In other schedules, we see both (1) and (2).

The double-yield optimization will remove this variability because it will only do (1) if the joined thread is already finished, so (2) will be skipped. Likewise it will always do (2) if (1) is skipped. So we always get exactly one scheduling point.

shuttle/src/sync/barrier.rs
dylanjwolff: Oops! this is leftover from the optimized version where we need the state at entry

shuttle/src/sync/mpsc.rs
dylanjwolff: I've moved the call to be in the conditional for this PR so the logic isn't as difficult to follow

shuttle/src/thread.rs
dylanjwolff: this is cleaned up to be a boolean parameter to `thread_fn` now

jorajeev (approved):

PR #211, opened 2025-09-03
Tuning down log-level for spawn events to reduce mem. usage of default test config

dylanjwolff:
The `test-log` crate uses `INFO` level logging by default. Thus anything at that level will be logged by default when running tests. The logs are also buffered in memory until the test has completed, which can lead to high memory consumption for tests producing verbose logs, in addition to increasing the test runtime.

While logging has already been turned off by default in CI before this PR (`RUST_LOG=off`), the default behavior with no additional environment variables causes the tests to run extremely slowly due to verbose logging. This is not a good experience for new developers of Shuttle and generally seems like a bad default. 

The primary culprit is logging task creation at `INFO` level; we have several tests which do millions of shuttle iterations, each containing multiple task spawns. This PR moves that logging to `DEBUG`, so it is off by default during testing, but can be easily re-enabled by setting `RUST_LOG=debug`. With this change, running the shuttle test suite on my machine goes from:

`cargo test --release`: from ~244s and >20 GB of RAM to ~183s and <1GB of RAM

---

### DIFF
--- a/shuttle/src/runtime/task/mod.rs
+++ b/shuttle/src/runtime/task/mod.rs
@@ -336,7 +336,7 @@ impl Task {
 
         // Note: the tests for the task signature in [`crate::tests::basic::task`] depend on tracing the task signature and creation point here
         error_span!(parent: parent_span_id, "new_task", parent = ?parent_task_id, i = schedule_len).in_scope(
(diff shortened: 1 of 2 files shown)


sarsko:
Its wild to me that this one change causes a 20x reduction in memory usage and a 25-ish% perf improvement. I am fine with the change, though I do like this event being INFO, as it is an important event. I’m wondering if the correct solution is to vend a better tracing subscriber (and our own test macro which installs it)

bkragl (approved):

PR #212, opened 2025-09-06
Next Event on Task struct

dylanjwolff:
Currently, Shuttle schedulers have no visibility into the system under test. The algorithms only see which tasks are enabled at a given time, not what those tasks are actually doing. This lack of semantic information greatly limits what we can do with Shuttle schedulers. Many algorithms, such as Partial Order Sampling (POS) and Selective Uniform Random Walk (SURW) leverage information about the events being executed to make better scheduling decisions (e.g. make a different decision if two events conflict, or if a particular task is accessing a resource “of interest”). Such information is also useful for developing a general notion of behavioral coverage — for example, tracking new reads-from edges.

This PR adds a `next_event` field to the Task struct, which is accessible to scheduling algorithms. The Event enum gives the schedulers access to the signature of the resource being accessed, as well as the source location of the access (via `track_caller`).

As a small "demo" application, I also modified the Metrics scheduler to count the number and kind of events observed during a Shuttle test. As part of this demo, I did make some light changes to the Scheduler API so that it returns a &Task instead of a TaskId. This change allows nested schedulers such as the Metrics scheduler to directly lookup information about the task chosen by inner schedulers without iterating over the runnable tasks.

Note: this PR does use some unsafe lifetime casting to allow Task to hold a reference to the Task/Resource signatures. Just outright cloning the signatures is maybe  ~5-8% slower in the worst case. I'm open to discussion/suggestions here, but I think the other approaches to avoid cloning without unsafe might end up being significantly more complicated than the unsafe code without much real benefit. In particular, using `Rc<_>` for signatures requires heap allocations on initialization, while many resource constructors are `const`. Note that this is similar to the unsafe added for the runnable tasks slice, and, as in that case, we are providing a safe API to the scheduling algorithms.

I am marking this as a draft for now because it depends on ~~two other open PRs (#207 and #210)~~ #216.

~~It currently includes the changes from these other PRs. To look at the changes unique to this PR, see 4778a31461c74077ea6ab49ebec82bbd69393dfd~~

I've rebased this on #216, will mark as ready for review as soon as that is merged

---

### DIFF
--- a/shuttle-engine/src/future/batch_semaphore.rs
+++ b/shuttle-engine/src/future/batch_semaphore.rs
@@ -1,7 +1,7 @@
 //! A counting semaphore supporting both async and sync operations.
 use crate::current;
 use crate::runtime::execution::ExecutionState;
-use crate::runtime::task::{clock::VectorClock, TaskId};
+use crate::runtime::task::{clock::VectorClock, Event, TaskId};
 use crate::runtime::thread;
 use crate::sync_types::{ResourceSignature, ResourceType};
 use std::cell::RefCell;
@@ -342,7 +342,6 @@ impl BatchSemaphoreState {
 pub struct BatchSemaphore {
     state: RefCell<BatchSemaphoreState>,
     fairness: Fairness,
-    #[allow(unused)]
     signature: ResourceSignature,
 }
 
@@ -445,8 +444,9 @@ impl BatchSemaphore {
 
     /// Closes the semaphore. This prevents the semaphore from issuing new
     /// permits and notifies all pending waiters.
+    #[track_caller]
     pub fn close(&self) {
-        thread::switch();
+        thread::switch(Event::batch_semaphore_rel(&self.signature));
         self.close_no_scheduling_point();
     }
 
@@ -495,8 +495,9 @@ impl BatchSemaphore {
     /// If the permits are available, returns Ok(())
     /// If the semaphore is closed, returns `Err(TryAcquireError::Closed)`
     /// If there aren't enough permits, returns `Err(TryAcquireError::NoPermits)`
+    #[track_caller]
     pub fn try_acquire(&self, num_permits: usize) -> Result<(), TryAcquireError> {
-        thread::switch();
+        thread::switch(Event::batch_semaphore_acq(&self.signature));
 
         self.init_object_id();
         let mut state = self.state.borrow_mut();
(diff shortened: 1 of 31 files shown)


shuttle-engine/src/future/batch_semaphore.rs
```diff
@@ -778,12 +783,25 @@ impl Future for Acquire<'_> {
                             self.waiter.is_queued.store(true, Ordering::SeqCst);
                         }
                         trace!("Acquire::poll for waiter {:?} that is enqueued", self.waiter);
+
+                        let event = Event::batch_semaphore_acq(&self.semaphore.signature);
+                        ExecutionState::with(|s| unsafe { s.current_mut().set_next_event(event) });
+                        // SAFETY: This is safe because the current task immediately suspends after this future
```
sarsko, line 888: SAFETY should be on the line above

shuttle/src/runtime/task/mod.rs
```diff
@@ -664,6 +793,19 @@ impl Task {
             }
         )
     }
+
+    /// Get the next_event with a downcast lifetime tied to self.
+    pub(crate) fn next_event(&self) -> &Event<'_> {
+        unsafe { std::mem::transmute(&self.next_event) }
+    }
+
+    pub(crate) unsafe fn set_next_event(&mut self, event: Event<'_>) {
```
sarsko: SAFETY comment describing how to make it safe. Also describe why unsafe is needed

shuttle-engine/src/runtime/task/mod.rs
```diff
@@ -664,6 +793,19 @@ impl Task {
             }
         )
     }
+
+    /// Get the next_event with a downcast lifetime tied to self.
+    pub(crate) fn next_event(&self) -> &Event<'_> {
+        unsafe { std::mem::transmute(&self.next_event) }
```
sarsko, line 820: SAFETY

shuttle/src/runtime/thread/continuation.rs
```diff
@@ -291,13 +291,28 @@ unsafe impl Send for PooledContinuation {}
...
+    // SAFETY we cast the lifetime of the Event to 'static when embedding it into the current Task
+    // This is safe because (1) we have a valid reference to the Event's data for the scope of this function
+    // and (2) the static reference is dropped at the end of the scope of this function when the next_event
+    // is set back to Unknown in `switch_keep_event`
+    ExecutionState::with(|s| s.try_current_mut().map(|c| unsafe { c.set_next_event(event) }));
+    switch_keep_event()
+}
+
+#[track_caller]
+pub(crate) fn switch_keep_event() {
```
sarsko: Should have a comment documenting this function. Also it looks like it loses the event on 315?

PR #213, opened 2025-09-15
Introduce SHUTTLE_BACKTRACE & centralize env vars

sarsko:
It's clear that relying on `RUST_LOG` is a poor scheme as it is set in a lot of testing environments, meaning slow tests on accidents. This PR thus introduces `SHUTTLE_BACKTRACE`.

In the process I also centralized and pubbed the various env vars.

---


shuttle/src/lib.rs
```diff
@@ -452,6 +452,31 @@ where
     runner.run(f);
 }
 
+/// If this environment variable is set, then Shuttle will capture the backtrace of each task and display
+/// the backtraces in the panic message.
+/// Capturing backtraces is quite expensive, so this should onluy be set wehn debugging a failing test.
```
jorajeev: typos: "only" , "when"

shuttle/src/lazy_static.rs
```diff
@@ -124,7 +125,7 @@ fn maybe_warn_about_drop(silence_warnings: bool) {
         .compare_exchange(false, true, Ordering::Relaxed, Ordering::Relaxed)
         .is_ok()
     {
-        if silence_warnings || std::env::var("SHUTTLE_SILENCE_WARNINGS").is_ok() {
+        if silence_warnings || std::env::var(SILENCE_WARNINGS).is_ok() {
```
jorajeev: Instead of checking the env var directly here, adopt the pattern for annotations etc, and define `fn silence_warnings()` in `lib.rs` ?

shuttle/src/runtime/task/mod.rs
```diff
@@ -453,16 +456,22 @@ impl Task {
     /// permitted to spuriously wake up the thread (though it will still not count as a live thread
     /// for deadlock detection purposes for as long as it remains blocked).
     pub(crate) fn block(&mut self, allow_spurious_wakeups: bool) {
-        // `Backtrace::capture()` is a noop (it returns the constant `disabled()`) if `RUST_BACKTRACE`/`RUST_LIB_BACKTRACE` is not set.
-        self.backtrace = Backtrace::capture();
+        self.backtrace = if backtrace_enabled() {
```
jorajeev, line 487: nit: this can be written more concisely as

```
backtrace_enabled().then(|| Backtrace::force_capture())
```

shuttle/src/runtime/task/mod.rs
```diff
@@ -453,16 +456,22 @@ impl Task {
...
+        };
 
         assert!(self.state != TaskState::Finished);
         self.state = TaskState::Blocked { allow_spurious_wakeups };
     }
 
     pub(crate) fn sleep(&mut self) {
-        // `Backtrace::capture()` is a noop (it returns the constant `disabled()`) if `RUST_BACKTRACE`/`RUST_LIB_BACKTRACE` is not set.
-        self.backtrace = Backtrace::capture();
+        self.backtrace = if backtrace_enabled() {
```
jorajeev, line 498: Can be more concise

shuttle/src/runtime/task/mod.rs
sarsko: It is more concise, it is also less readable. 

Everybody knows `if`-statements, `bool.then` is weird Rust voodoo that I even after having seen it a few times still have to think about.

PR #216, opened 2025-09-19
Changing Sched. Points to Always *precede* operations w/ "Double Yield Optimizations"

dylanjwolff:
This PR updates the changes from #210 with performance optimizations to avoid "double-yields". 

~~To view these changes in isolation from the rest of #210, see:~~

~~62a5b9f~~ -- see for previous comments
~~ebaf930~~ -- see for previous comments
~~82f7c9e~~ -- see for prefvious comments

#### Update

This PR now subsumes #210, addresses all review comments on both PRs and has all replay tests updated for the new scheduling point placement.

Note that this change has provoked some issues with panic-handling and task cleanup are partially resolved by #202 but need further thought. Additionally #202 has broken the replay mechanism for some tests (because the schedule is not serialized in the same way) which will need to be addressed in a future PR.

---

### DIFF
--- a/shuttle/src/thread.rs
+++ b/shuttle/src/thread.rs
@@ -104,6 +104,10 @@ impl<'scope> Scope<'scope, '_> {
             }
         };
 
+        // Note: Scoped threads wrap their inner function in some additional logic (above) to update the `finished` variable on termination.
+        // This logic is expected to run atomically with scoped thread termination, so there *cannot* be a switch after the atomic store.
+        // To avoid violating this invariant, we pass `switch_before_exit = false` (below). Instead, we provide our own context switch on exit
+        // (above) in the `scope_closure` *before* setting `finished` to be `true`.
         // SAFETY: main task is blocked until all scoped closures complete so all captured references remain valid
         ScopedJoinHandle {
             handle: unsafe { spawn_named_unchecked(scope_closure, None, None, false, Location::caller()) },
@@ -209,6 +213,9 @@ where
 
 /// Body of a Shuttle thread, that runs the given closure, handles thread-local destructors, and
 /// stores the result of the thread in the given lock.
+/// The `switch_before_exit` parameter will provide a conditional scheduling point after the inner
+/// function has completed. If this parameter is set to `false`, then this function should be wrapped
+/// in another function which provides a conditional scheduling point before the task exits.
 pub(crate) fn thread_fn<F, T>(
     f: F,
     switch_before_exit: bool,


sarsko:
Added comments on https://github.com/awslabs/shuttle/commit/ebaf9302b021a6269ddd4d3e6447db1a9c1b44e2#r167052941

shuttle/src/runtime/execution.rs
```diff
@@ -149,7 +150,7 @@ impl Execution {
         EXECUTION_STATE.set(&state, move || {
             // Spawn `f` as the first task
             ExecutionState::spawn_main_thread(
-                Box::new(move || thread_fn(f, Default::default())),
+                Box::new(move || thread_fn(f, true, Default::default())),
```
sarsko, line 153: Why is `switch_before_exit` a parameter?

dylanjwolff: because scoped threads wrap their function in some additional logic before passing to `thread_fn` that update a "finished" variable on termination:

```
                finished.store(true, Ordering::Relaxed);

                if self.num_running_threads.fetch_sub(1, Ordering::Relaxed) == 1 {
                    ExecutionState::with(|s| s.get_mut(self.main_task).unblock());
                }
```


If the scheduling point is always in `thread_fn`, then we will sometimes set that variable to true, but then switch away from the task before it actually finishes. I.e. this logic is expected to run atomically with scoped thread termination, so there cannot be a switch after the atomic store.

The boolean parameter lets scoped threads have their own exit context switch point in the appropriate place (before the store) while otherwise reusing logic

sarsko: Okay thanks could we document this in the code?

shuttle/tests/basic/condvar.rs
```diff
@@ -277,12 +277,13 @@ fn check_producer_consumer_broken1() {
     check_random(producer_consumer_broken1, 5000)
 }
 
+#[ignore = "replay mechanism is broken because the schedule is not emitted in the panic output. reintroduce once replay mechanism is fixed."]
```
sarsko, line 280: Not sure this is correct, I think this was added accidentally because I added it accidentally ?

shuttle/tests/basic/mutex.rs
```diff
@@ -204,6 +204,7 @@ fn mutex_rwlock_interaction() {
                 let log = Arc::clone(&value);
                 thread::spawn(move || {
                     let _guard = lock.lock().unwrap();
+                    thread::yield_now(); // log is `std::sync::Mutex`, so it is not visible to Shuttle without an explicit yield
```
sarsko, line 207: ? Why is this needed? Surely it being `std::sync` is intentional and thus there's no point to have the `yield_now`?

shuttle/tests/basic/thread.rs
```diff
@@ -345,7 +346,8 @@ mod thread_local {
                     seen_drop_counters.lock().unwrap().insert(DROPS.load(Ordering::SeqCst));
                 });
 
-                // force access to the thread local so that it's initialized
+                thread::yield_now(); // thread-local operations are not visible, so we must manually allow a switch
```
sarsko, line 349: Huh why is this needed?

shuttle/tests/basic/thread.rs
```diff
@@ -12,11 +12,12 @@ fn thread_yield_point() {
...
 
-    // We want to see executions that include both threads running first, otherwise we have
-    // messed up the yieldpoints around spawn.
+    // We want to see executions that include both threads running first.
+    // Spawn no longer itself produces a switch, but accessing any Shuttle primitive will allow
+    // either thread to run its visible operations first.
     check_random(
         move || {
-            let flag = Arc::new(AtomicBool::new(false));
+            let flag = Arc::new(shuttle::sync::atomic::AtomicBool::new(false));
```
sarsko, line 20: ? I assume previously this was from `std::sync` and there was a reason for that?

shuttle/tests/basic/thread.rs
```diff
@@ -555,6 +557,7 @@ fn thread_park_spuriously_wakeup() {
                 })
             };
 
+            thread::yield_now(); // `flag` is `std::sync` and thus not visible to Shuttle, so we need to give an opportunity to switch
```
sarsko, line 560: Same comment as above

shuttle/tests/basic/thread.rs
```diff
@@ -667,6 +670,7 @@ fn thread_double_unpark() {
                 })
             };
 
+            thread::yield_now(); // `unpark_count` is `std::sync` and thus not visible to Shuttle, so we need to give an opportunity to switch
```
sarsko, line 673: Same comment as above

shuttle/tests/data/random.rs
```diff
@@ -28,10 +28,11 @@ fn random_mod_10_equals_7_replay_fails() {
     replay(random_mod_10_equals_7, "910102fe93a9cef4f3faaf5a04")
 }
 
+#[ignore = "replay mechanism is broken because the schedule is not emitted in the panic output. reintroduce once replay mechanism is fixed."]
```
sarsko, line 31: Same comment as above

shuttle/tests/data/random.rs
```diff
@@ -229,6 +230,7 @@ fn dfs_does_not_reseed_across_executions() {
             }
         });
 
+        thread::yield_now(); // Everything below is not visible to Shuttle, so we need to give an opportunity to switch
```
sarsko, line 233: Why is this needed?

shuttle/tests/basic/condvar.rs
dylanjwolff: There are some issues with replay that I think are related to your changes in #202

I added the same ignore message to make sure all related issues get cleaned up at the same time

shuttle/tests/basic/thread.rs
dylanjwolff: As mentioned in the comment, `spawn` used to provide a switch afterwards where the child can be run. Now `spawn` provides a switch *before* (child cannot be run), so if there are zero events after a spawn the child will never have a chance to run until the parent thread exits

shuttle/tests/basic/thread.rs
dylanjwolff: same as the other example, `spawn` no longer provides a scheduling point where the child runs immediately. the child can only be run when the next event on the parent thread is reached. there are no such events in this test until the parent thread exits

shuttle/tests/basic/thread.rs
dylanjwolff: same response as others -- related to spawn scheduling points

shuttle/tests/data/random.rs
dylanjwolff: same as other, replay mechanism is not working due to changes in #202

shuttle/tests/data/random.rs
dylanjwolff: same as others, this test expects the child thread to be able to be run  before all of the local operations, but there is no longer a scheduling point after `spawn`

shuttle/tests/basic/mutex.rs
dylanjwolff: In prior versions of shuttle, the scheduling point was *after* acquiring the lock, so there was a scheduling point between `let _guard = lock.lock().unwrap();` and `*log.lock().unwrap() *= 2;`, even though the `std::sync` acquire of `log.lock()` is not naturally visible.

With these changes, the scheduling point is moved to *before* `let _guard = lock.lock().unwrap();`, so there is no longer a natural scheduling point between it and the invisible `*log.lock().unwrap() *= 2;`.

The scheduling point in between these two lines is necessary in order to observe the value 17 in the test assertion.

dylanjwolff: As with the other `yield_now`s, this is only necessary because of assumptions regarding scheduling points around invisible `std::sync` primitives, which should only come up in Shuttle testing code

PR #218, opened 2025-10-07
Adding global config for Shuttle

dylanjwolff:
This PR adds a global config for Shuttle, utilizing the [config](https://crates.io/crates/config) crate to make Shuttle tests configurable via a TOML file and/or environment variables. This allows users to swap scheduling algorithms, configuration parameters, or (in the future #217) time models declaratively without editing or recompiling their test code.

The public `check` function now uses this global config when running Shuttle tests. All other entrypoints to Shuttle (`check_random`, building a `Runner` manually, etc.) are unchanged.

Currently, the global config is the same for all tests using `check`, but it is easy to extend this in a future PR to store "sub-configs" in the global config objects, which can be accessed in subsets of tests via a key `check(fn, key)` API.

To run a scheduler via the global config, that scheduler must register a "factory" function that can create instances of that scheduler with the scheduler "registry". The scheduler "registry" is a global static that maps the name of the scheduler used in the config (`scheduler.type`) to the factory function. All basic schedulers in the Shuttle repository are registered by default, but additional custom schedulers can also be registered dynamically so that they can be run via the global config / check function.

#### Documentation

`shuttle/shuttle_config_example.toml` -- provides an example config with most configuration options for the included Shuttle schedulers
`fn check` -- the doc comment explains the configuration precedence and environment variable naming convention

---

### DIFF
--- a/shuttle/src/lib.rs
+++ b/shuttle/src/lib.rs
@@ -193,10 +193,14 @@ pub mod scheduler;
 
 mod runtime;
 
+use std::path::Path;
+use std::path::PathBuf;
+
 pub use runtime::runner::{PortfolioRunner, Runner};
+pub use scheduler::registry::{register_scheduler, SchedulerFactory};
 
 /// Configuration parameters for Shuttle
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 #[non_exhaustive]
 pub struct Config {
     /// Stack size allocated for each thread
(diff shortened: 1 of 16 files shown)


shuttle/src/scheduler/registry.rs
```diff
@@ -0,0 +1,98 @@
+use crate::scheduler::{
+    DfsScheduler, PctScheduler, RandomScheduler, ReplayScheduler, RoundRobinScheduler, Scheduler, UrwRandomScheduler,
+};
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, OnceLock};
+
+/// Factory function type for creating schedulers from configuration
+pub type SchedulerFactory = Arc<dyn Fn(&config::Config) -> Box<dyn Scheduler + Send> + Send + Sync>;
+
+static REGISTRY: OnceLock<Mutex<HashMap<String, SchedulerFactory>>> = OnceLock::new();
```
sarsko, line 10: What's a registry?

shuttle/src/scheduler/pct.rs
```diff
@@ -38,6 +38,17 @@ impl PctScheduler {
...
+    /// Construct a new PctScheduler from configuration.
+    pub fn from_config(config: &config::Config) -> Self {
+        let depth = config.get_int("scheduler.depth").unwrap_or(3) as usize;
+        let iterations = config.get_int("scheduler.iterations").unwrap_or(100) as usize;
+        if let Ok(seed) = config.get_int("scheduler.seed") {
+            Self::new_from_seed(seed as u64, depth, iterations)
+        } else {
+            Self::new(depth, iterations)
+        }
+    }
```
sarsko, line 50: I feel like these shouldn't exist and that constructing a scheduler from a config should be done via the regular scheduler creation apis and live with the config

shuttle/src/scheduler/registry.rs
sarsko: This is a request for a comment on the registry

shuttle/src/scheduler/registry.rs
```diff
@@ -0,0 +1,98 @@
...
+
+fn get_registry() -> &'static Mutex<HashMap<String, SchedulerFactory>> {
+    REGISTRY.get_or_init(|| {
+        let mut map = HashMap::new();
+
+        // Register built-in schedulers
+        map.insert(
```
sarsko, line 17: Why do we `insert` instead of calling `register_scheduler`?

shuttle/src/lib.rs
```diff
@@ -290,6 +331,25 @@ pub enum FailurePersistence {
     File(Option<std::path::PathBuf>),
 }
 
+impl FailurePersistence {
```
sarsko, line 334: These things might be easier with https://docs.rs/strum/latest/strum/

shuttle/src/lib.rs
```diff
@@ -319,19 +379,155 @@ pub enum MaxSteps {
...
+    }
+}
+
+fn should_check_uncontrolled_nondeterminism(global_config: &config::Config) -> bool {
+    global_config.get_bool("check_uncontrolled_nondeterminism").unwrap()
+}
+
+fn should_enable_metrics(global_config: &config::Config) -> bool {
+    global_config.get_bool("enable_metrics").unwrap()
+}
```
sarsko, line 407: These things should be impled on `&self` where `self` is the config and live somewhere else

sarsko:
I think this is the least important PR to get merged this week. If you don't get to it, then I'll take on the work to drive it to merged.

PR #232, opened 2025-11-22
Add shuttle-parking_lot

sarsko:
Adds Shuttle support for the [parking_lot](https://docs.rs/parking_lot/latest/parking_lot/) library

---


wrappers/parking_lot/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "shuttle-parking_lot"
```
jorajeev, line 2: `shuttle-parking_lot` or `shuttle_parking_lot` ?

wrappers/parking_lot/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.0.1"
+edition = "2024"
+
+[features]
+shuttle = [ "dep:shuttle-parking_lot-impl" ]
+
+arc_lock = [ "parking_lot/arc_lock", "shuttle-parking_lot-impl?/arc_lock" ]
```
jorajeev, line 9: Could we please put the features [in the same order](https://github.com/Amanieu/parking_lot/blob/master/Cargo.toml) as they appear in the `parking_lot` crate:

arc_lock
owning_ref
nightly
deadlock_detection
serde
send_guard
hardware-lock-elision

wrappers/parking_lot/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "shuttle-parking_lot"
+version = "0.0.1"
```
jorajeev: This should be "0.12.5" right?

wrappers/parking_lot/parking_lot_impl/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "shuttle-parking_lot-impl"
```
jorajeev, line 2: Should this be `shuttle_parking_lot_impl` ?

wrappers/parking_lot/parking_lot_impl/src/lib.rs
```diff
@@ -0,0 +1,149 @@
+//! This is the "impl" crate implementing [parking_lot] support for [`Shuttle`].
```
jorajeev: "This is the Shuttle replacement for the [parking_lot] crate, for use when testing with the Shuttle concurrency checker"

wrappers/parking_lot/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+//! This crate provides a Shuttle-compatible implementation and wrapper for [`parking_lot`] in order to make it
+//! more ergonomic to run a codebase under Shuttle.
+//!
+//! [`parking_lot`]: <https://crates.io/crates/parking_lot>
+//!
+//! To use this crate, add something akin to the following to your Cargo.toml:
```
jorajeev, line 6: Also add the `shuttle` feature flag to your crate

wrappers/rand_0.8/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "shuttle-rand"
+version = "0.0.1"
```
jorajeev: Should be `0.8` no?

wrappers/rand_0.8/shuttle_rand_impl/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! This is the "inner" crate implementing [`rand` version 0.8] support for [`Shuttle`].
+//! This is the "impl" crate implementing [`rand` version 0.8] support for [`Shuttle`].
```
jorajeev: "This is the Shuttle replacement for the [rand] crate, for use when testing with the Shuttle concurrency checker"

wrappers/parking_lot/parking_lot_impl/src/lib.rs
sarsko: Not sure I agree, we don't want people to use this direclty

wrappers/parking_lot/parking_lot_impl/Cargo.toml
sarsko: I have no strong preference. I have tried to follow dashes around external crate, under the assumption that most external crates use underscores

wrappers/parking_lot/src/lib.rs
sarsko: ? Adding `shuttle` feature flag to your crate is the first thing that's done in the code snippet.

wrappers/parking_lot/Cargo.toml
sarsko: Ditto on the comment above

wrappers/parking_lot/Cargo.toml
sarsko: Debatable, I think no.

In my head releasing version mirrors is independent from the code. Thus all wrappers will be pushed as 0.0.1, because this is a releasable non-colliding version for all the packages, meaning we can release it and depend on it without there being any confusion on what version its wrapping. We can also yank a version 0.0.1 that was released (accidentally or not) without any loss. It is thus "safe" to have version 0.0.1. Version "0.12.5" however has a version collision, and publishing it is a final action. It is not "safe" to have version 0.12.5.

Regardless of what we release, I think the crates should live in the repo as 0.0.1s

wrappers/parking_lot/Cargo.toml
sarsko: Yes, I think the list here is alphabetical

wrappers/rand_0.8/shuttle_rand_impl/src/lib.rs
sarsko: Ditto on my comment above, I think we should make it clear that this is not intended to be used directly and we want to communicate that with the doc comment

wrappers/rand_0.8/shuttle_rand_impl/Cargo.toml
```diff
@@ -1,11 +1,10 @@
 [package]
-name = "shuttle-rand-inner"
-version = "0.1.0"
+name = "shuttle-rand-impl"
+version = "0.0.1"
```
sarsko: This should be 0.1.0

wrappers/rand_0.8/Cargo.toml
sarsko: Same as the comment on the parking lot version

wrappers/parking_lot/Cargo.toml
sarsko: This list is from https://docs.rs/crate/parking_lot/latest/source/Cargo.toml, which normalizes the features list. I don't have a strong preference either way, but I think following `docs.rs` is the better alternative? That said I more often read code via the GitHub repo

wrappers/parking_lot/parking_lot_impl/src/lib.rs
jorajeev: But saying `"impl" crate` doesn't convey anything, since "impl" isn't a word.

Let's go with
```
//! This crate contains Shuttle's internal implementations of the `parking_lot` crate.
//! Do not depend on this crate directly. Use the `shuttle-parking_lot` crate, which conditionally
//! exposes these implementations with the `shuttle` feature or the original crate without it.
```

wrappers/parking_lot/Cargo.toml
jorajeev: I want to keep the list in the same order as the crate itself.  Otherwise, it's harder to see any differences.

wrappers/parking_lot/parking_lot_impl/Cargo.toml
jorajeev: Okay, let's just go with `shuttle-` as the prefix

wrappers/parking_lot/src/lib.rs
jorajeev: We need to update this and add the recommendation that by default, users get the latest version of `parking_lot`, and that if they want to depend on a specific version, they should add a dependency naming it `parking_lot_orig`

wrappers/parking_lot/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+//!
+//! ```ignore
+//! [features]
+//! shuttle = [
+//!    "parking_lot/shuttle",
+//! ]
+//!
+//! [dependencies]
+//! parking_lot = { package = "shuttle-parking_lot", version = "VERSION_NUMBER" }
```
jorajeev: This is incorrect now, right?  We now want them to depend on `version = "*"`

wrappers/rand_0.8/shuttle_rand_impl/src/lib.rs
jorajeev: Rewrite to

```
//! This crate contains Shuttle's internal implementations of the `rand` crate.
//! Do not depend on this crate directly. Use the `shuttle-rand` crate, which conditionally
//! exposes these implementations with the `shuttle` feature or the original crate without it.
```

sarsko:
All comments should be addressed

wrappers/parking_lot/parking_lot_impl/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+//!
+//! [`Shuttle`]: <https://crates.io/crates/shuttle>
+//!
+//! [`parking_lot`]: <https://crates.io/crates/parking_lot>
+
+use shuttle::sync;
+
+// TODO: All the `expect`s below are wrong — it is possible to poison a Shuttle mutex.
+//       `clear_poison` should be added and used to avoid the `expect`.
+// TODO: Shuttle should be split up into a "core" crate which offers a more versatile
```
jorajeev: Could you make an issue for this (refer to #249) and mention that here?

wrappers/parking_lot/parking_lot_impl/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+            inner: sync::Mutex::new(value),
+        }
+    }
+
+    /// Consumes this mutex, returning the underlying data.
+    #[inline]
+    pub fn into_inner(self) -> T {
+        self.inner
+            .into_inner()
+            .expect("This shouldn't happen as there is no way to poision a Shuttle Mutex.")
```
jorajeev: "poison" misspelled, here and below (lines 54, 87, ...) (do a search/replace to fix)

jorajeev (commented):
Also add a README similar to what we have for tokio (see https://github.com/awslabs/shuttle/blob/main/wrappers/tokio/README.md )?

wrappers/parking_lot/parking_lot_impl/src/mutex.rs
```diff
@@ -0,0 +1,146 @@
+//! An asynchronous reader-writer lock.
```
jorajeev: Fix comment - this is a mutex

wrappers/parking_lot/parking_lot_impl/src/mutex.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use shuttle::future::batch_semaphore::{BatchSemaphore, Fairness, TryAcquireError};
+use std::cell::UnsafeCell;
+use std::error::Error;
+use std::fmt::{self, Debug, Display};
+use std::ops::{Deref, DerefMut};
+use std::thread;
+use tracing::trace;
+
+/// An asynchronous semaphore
```
jorajeev: nit: "An async mutex"

wrappers/parking_lot/parking_lot_impl/src/rwlock.rs
```diff
@@ -0,0 +1,305 @@
...
+        RwLock {
+            max_readers,
+            sem: BatchSemaphore::const_new(max_readers, Fairness::StrictlyFair),
+            inner: UnsafeCell::new(value),
+        }
+    }
+}
+
+impl<T: ?Sized> RwLock<T> {
+    /// until it can be acquired.
```
jorajeev, line 45: Some text got deleted here?

wrappers/parking_lot/parking_lot_impl/src/mutex.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn acquire(&self) {
+        trace!("acquiring parking_lot lock {:p}", self);
+        self.semaphore.acquire_blocking(1).unwrap_or_else(|_| {
+            // The semaphore was closed. but, we never explicitly close it, and
+            // we own it exclusively, which means that this can never happen.
+            if !thread::panicking() {
+                unreachable!()
+            }
+        });
+        trace!("acquired parking_Lot lock {:p}", self);
```
jorajeev: typo: should be "parking_lot" not "parking_Lot"

wrappers/parking_lot/parking_lot_impl/src/rwlock.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+    }
+
+    /// Attempts to acquire this `RwLock` with shared read access.
+    ///
+    /// If the access couldn't be acquired immediately, returns [`TryLockError`].
+    /// Otherwise, an RAII guard is returned which will release read access
+    /// when dropped.
+    pub fn try_read(&self) -> Option<RwLockReadGuard<'_, T>> {
+        match self.sem.try_acquire(1) {
```
jorajeev, line 84: Add a trace message ("try_read acquiring ReadGuard") to be consistent with the other methods.

Also could we make the trace messages have a consistent format? The "try_*" variants mention the method name, so let's do that in the "read" and "write" calls too.  That makes it easier to process the logs with tools.

wrappers/parking_lot/parking_lot_impl/src/rwlock.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+    }
+
+    /// Attempts to acquire this `RwLock` with exclusive write access.
+    ///
+    /// If the access couldn't be acquired immediately, returns [`TryLockError`].
+    /// Otherwise, an RAII guard is returned which will release write access
+    /// when dropped.
+    pub fn try_write(&self) -> Option<RwLockWriteGuard<'_, T>> {
+        tracing::trace!("parking_lot rwlock {:p} try_write acquired WriteGuard", self,);
```
jorajeev: typo: should be "acquiring" not "acquired"

jorajeev (approved):

PR #240, opened 2025-12-04
Update wrappers README to cover wrapper versioning

sarsko:
Covers wrapper versioning.

Could include the alternatives to doing it like this, but that might also just add confusion.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Shuttle is a library for testing concurrent Rust code. It is an implementation o
 *randomized concurrency testing* techniques, including
 [A Randomized Scheduler with Probabilistic Guarantees of Finding Bugs](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/asplos277-pct.pdf).
 
+This repository also houses a collection of [wrappers](/wrappers) and their implementations. These make it easier to test code using popular libraries such as `tokio` and `rand`, as well as synchronization primitives from `std::sync` and collections from `std::collections`. For more information on how to use these wrappers, see the [wrappers/README.md](/wrappers/README.md).
+
 ## Getting started
 
 Consider this simple piece of concurrent code:
--- a/wrappers/README.md
+++ b/wrappers/README.md
@@ -5,13 +5,15 @@ This directory contains a collection of wrappers and their implementations. The 
 ```toml
 [features]
 shuttle = [
-   "PACKAGE_NAME/shuttle",
-   "OTHER_PACKAGE_NAME/shuttle",
+   "tokio/shuttle",
+   "parking-lot/shuttle",
+   # ... etc. for all wrapped dependencies
 ]
 
 [dependencies]
-PACKAGE_NAME = { package = "shuttle-PACKAGE_NAME", version = "VERSION_NUMBER" }
-OTHER_PACKAGE_NAME = { package = "shuttle-OTHER_PACKAGE_NAME", version = "VERSION_NUMBER_2" }
+tokio = { package = "shuttle-tokio", version = "1" }
+parking_lot = { package = "shuttle-parking_lot", version = "0.12" }
+# ... etc. for all wrapped dependencies
 ```
 
 When running without the `shuttle` feature flag enabled, the code will behave as normal (ie., run with primitives from `std`, `rand`, `tokio` etc), and when the `shuttle` feature flag is enabled, the code will use primitives which are compatible with Shuttle.
@@ -26,8 +28,8 @@ shuttle = [
 
 [dependencies]
 shuttle_enabler = "0.1.0"
-PACKAGE_NAME = { package = "shuttle-PACKAGE_NAME", version = "VERSION_NUMBER" }
-OTHER_PACKAGE_NAME = { package = "shuttle-OTHER_PACKAGE_NAME", version = "VERSION_NUMBER_2" }
+tokio = { package = "shuttle-tokio", version = "1" }
+parking_lot = { package = "shuttle-parking_lot", version = "0.12" }
 ```
 
 Note that depending on `shuttle_enabler` will cause all crates in `shuttle_enabler` to be compiled when the `shuttle` flag is enabled.
--- a/wrappers/README.md
+++ b/wrappers/README.md
@@ -1,0 +1,57 @@
+# Wrappers
+
+This directory contains a collection of wrappers and their implementations. The general scheme to use these is by doing the following:
+
+```toml
+[features]
+shuttle = [
+   "PACKAGE_NAME/shuttle",
+   "OTHER_PACKAGE_NAME/shuttle",
+]
+
+[dependencies]
+PACKAGE_NAME = { package = "shuttle-PACKAGE_NAME", version = "VERSION_NUMBER" }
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -8,6 +8,8 @@ Shuttle is a library for testing concurrent Rust code. It is an implementation o
 *randomized concurrency testing* techniques, including
 [A Randomized Scheduler with Probabilistic Guarantees of Finding Bugs](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/asplos277-pct.pdf).
 
+This repository also houses a collection of [wrappers](/wrappers) and their implementations. These make it easier to test code using popular libraries such as `tokio` and`rand`, as well as synchronization primitives from `std::sync` and collections from `std::collections`. For more information on how to use these wrappers, see the [wrappers/README.md](/wrappers/README.md).
```
jorajeev: typo: space missing before `rand`

wrappers/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+//! ```ignore
+//! [features]
+//! shuttle = [
+//!    "PACKAGE_NAME/shuttle",
+//! ]
+//!
+//! [dependencies]
+//! PACKAGE_NAME = { package = "shuttle-PACKAGE_NAME", version = "VERSION_NUMBER" }
+//! ``
```
jorajeev: This should be triple backquote?

wrappers/README.md
```diff
@@ -0,0 +1,36 @@
...
+By default, the wrappers place no additional constraints on the crate they wrap. This means that the wrapper for `tokio` or `lazy_static` has their dependency for `tokio`/`lazy_static` set to `1`, and that the wrappers for any crate with major version 0 (`rand`, `parking_lot`) has their dependency set to whatever minor they wrap (0.8, 0.12). This means that there is no "mirroring" of the patch version, under the assumption that it will be picked up automatically.
+
+This means that if no further constraining of the versioning is done, then depending on the wrapper will give you the latest version of the wrapped crate. If you want to use a specific version of a crate, or want to constrain the versioning, then this can be done by adding a second field in your `Cargo.toml`.
+
+For example, if you want to pin `tokio` to `1.36.0`, you can add the following:
+
+```
+[dependencies]
+tokio = { package = "shuttle-tokio", version = "0.1.0" }
+tokio-version-importer-do-not-use-directly = { version = "=1.36.0" }
```
jorajeev, line 52: This looks ugly - how about just writing

```
tokio = { package = ... }
tokio-version-pin = { version = "=1.36.0" }
```

wrappers/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+If you are paranoid about accidentally using `tokio-version-importer-do-not-use-directly`, then you can create a new crate which contains the `tokio` dependency and exports no code, and then have your crate depend on that crate.
```
jorajeev: Also a bit confusing to read. How about:

```
A more robust organization scheme that protects against wrong versions being accidentally picked up would be the following:
- create a dedicated `versions` crate that specifies the dependency versions you want (e.g., for `tokio`)
- have all your crates depend on this `versions` crate

For large projects spanning multiple workspaces, you should use workspace dependencies with `workspace = true`
```

wrappers/README.md
```diff
@@ -0,0 +1,36 @@
...
+//! ]
+//!
+//! [dependencies]
+//! PACKAGE_NAME = { package = "shuttle-PACKAGE_NAME", version = "VERSION_NUMBER" }
+//! ``
+```
+
+When running without the `shuttle` feature flag enabled, the code will behave as normal (ie., run with primitives from `std`, `rand`, `tokio` etc), and when the `shuttle` feature flag is enabled, the code will use primitives which are compatible with Shuttle.
+
+To make setting the feature flags easier and to avoid accidentally missing a crate, the [shuttle_enabler](shuttle_enabler) crate exists. Note that depending on this will cause all crates in this folder to be compiled when compiling under `shuttle`.
```
jorajeev: Think we need to explain this a bit better.  Maybe show how `shuttle_enabler` should typically be used?

wrappers/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## A note on versioning
+
+By default, the wrappers place no additional constraints on the crate they wrap. This means that the wrapper for `tokio` or `lazy_static` has their dependency for `tokio`/`lazy_static` set to `1`, and that the wrappers for any crate with major version 0 (`rand`, `parking_lot`) has their dependency set to whatever minor they wrap (0.8, 0.12). This means that there is no "mirroring" of the patch version, under the assumption that it will be picked up automatically.
```
jorajeev: This entire text is a bit confusing to read.  How about something along the lines of the following:

```
## A note on versioning

The wrappers use minimal version constraints for the crates they wrap:
- For crates at version 1.x or higher (like `tokio`, `lazy_static`), the dependency is set to `1`
- For crates at version 0.x (like `rand`, `parking_lot`), the dependency is the minor version (e.g., `0.8`, `0.12`)

Thus, by default, cargo will select the latest compatible version of the wrapped crate.
```

wrappers/README.md
sarsko: Yes

wrappers/README.md
sarsko: It being ugly is the point ? It's to protect against accidentally using it

wrappers/README.md
sarsko: Yes

wrappers/README.md
sarsko: Which part do we need to explain better? I can rewrite "Note that depending on this will cause all crates in this folder to be compiled when compiling under `shuttle`", is there more? The assumption here is that the shuttle_enabler crate exists and that it documents itself.

wrappers/README.md
sarsko: Sure

wrappers/README.md
sarsko: llm-o: "A more robust organization scheme that **protects against wrong versions being accidentally picked up** would be the following:": this is not what it protects against
llm-o: "large projects spanning **multiple** workspaces ...": no, in that instance one should use the version crate

wrappers/README.md
jorajeev: Why was this marked resolved? You didn't make the change I suggested, right?

sarsko: Looks like I forgot to push

sarsko: Pushed now

wrappers/README.md
```diff
@@ -0,0 +1,57 @@
+# Wrappers
+
+This directory contains a collection of wrappers and their implementations. The general scheme to use these is by doing the following:
+
+```toml
+[features]
+shuttle = [
+   "PACKAGE_NAME/shuttle",
```
jorajeev: I still find this hard to read - since PACKAGE_NAME is a placeholder (i.e., acts like a variable).

I think this would be a lot more readable if we just used specific examples - users are smart enough to generalize:

```
shuttle = [
    "shuttle-tokio/shuttle",
    "shuttle-parking_lot/shuttle",
]
```

etc.

sarsko: Fair

jorajeev (approved):

PR #258, opened 2026-03-27
Add BatchSemaphore::upgrade and parking_lot::RwLock::upgrade

sarsko:
Adds upgradeable reads to `parking_lot::RwLock`

---

### DIFF
--- a/shuttle/src/future/batch_semaphore.rs
+++ b/shuttle/src/future/batch_semaphore.rs
@@ -635,6 +635,9 @@ impl BatchSemaphore {
     /// naive implementation where `permits_to_be_held - permits_currently_held` is `acquire`d, and two tasks try to `upgrade`
     /// concurrently (or one `upgrade` in the presence of a `write`).
     pub fn upgrade(&self, permits_currently_held: usize, permits_to_be_held: usize) -> Acquire<'_> {
+        assert!(permits_currently_held > 0);
+        assert!(permits_to_be_held > permits_currently_held);
+
         let mut acquire = Box::pin(self.acquire(permits_to_be_held));
         let waker = ExecutionState::with(|state| state.current_mut().waker());
         let cx = &mut Context::from_waker(&waker);
--- a/shuttle/src/future/batch_semaphore.rs
+++ b/shuttle/src/future/batch_semaphore.rs
@@ -622,6 +622,28 @@ impl BatchSemaphore {
         }
         drop(state);
     }
+
+    /// Atomically `upgrade`s from holding `permits_currently_held` to holding `permits_to_be_held`.
+    /// The motivating use case for this is `parking_lot`s `RwLockUpgradableReadGuard::ugrade`, where we want to be able to
+    /// go from having a read guard to a write guard while honoring the order of `acquire`s.
+    ///
+    /// This is implemented by first trying to `acquire` `permits_to_be_held` (which for the `RwLock::upgrade` case would never
+    /// succeed, as the task is holding one permit, and wants to acquire all of them, meaning even with no other tasks it will
+    /// block on itself), then `release`ing `permits_currently_held`.
+    ///
+    /// This ensures the order of `acquire`s is honored, and prevents the potential deadlock situation which could occur in the
+    /// naive implementation where `permits_to_be_held - permits_currently_held` is `acquire`d, and two tasks try to `upgrade`
+    /// concurrently (or one `upgrade` in the presence of a `write`).
+    pub fn upgrade(&self, permits_currently_held: usize, permits_to_be_held: usize) -> Acquire<'_> {
+        let mut acquire = Box::pin(self.acquire(permits_to_be_held));
+        let waker = ExecutionState::with(|state| state.current_mut().waker());
+        let cx = &mut Context::from_waker(&waker);
+        let _poll = acquire.as_mut().poll(cx);
+
+        self.release(permits_currently_held);
+
+        *Pin::into_inner(acquire)
+    }
 }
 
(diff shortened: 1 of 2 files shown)


sarsko:
Code is wrong, ref https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#method.upgradable_read:
```
Locks this rwlock with upgradable read access, blocking the current thread until it can be acquired.

The calling thread will be blocked until there are no more writers **or other upgradable reads** which hold the lock. There may be other readers currently inside the lock when this method returns.

Returns an RAII guard which will release this thread's shared access once it is dropped.
```

I put **s around the important part, which is the other upgradeable reads part. Credit to this Reddit post: https://www.reddit.com/r/rust/comments/nbrh9v/upgradable_parking_lotrwlock_might_not_do_what/ and this article: https://morestina.net/1739/upgradable-parking_lotrwlock-might-not-be-what-you-expect for pointing this out.

sarsko:
Made it match the semantics of the one from parking_lot (sans the fairness discussed in https://github.com/awslabs/shuttle/issues/259)

Playground I used to check behavior of the parking_lot RwLock:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cffda8f89a8d2fd2dfe0f2f17fc5012b

sarsko:
There is an alternative design which implements this with only one semaphore by making `write` map to `upgradable_read` + `upgrade`. That causes extra scheduling points for regular `write` acquires, and also makes it not atomic.

shuttle/src/future/batch_semaphore.rs
```diff
@@ -622,6 +622,28 @@ impl BatchSemaphore {
...
+    ///
+    /// This is implemented by first trying to `acquire` `permits_to_be_held` (which for the `RwLock::upgrade` case would never
+    /// succeed, as the task is holding one permit, and wants to acquire all of them, meaning even with no other tasks it will
+    /// block on itself), then `release`ing `permits_currently_held`.
+    ///
+    /// This ensures the order of `acquire`s is honored, and prevents the potential deadlock situation which could occur in the
+    /// naive implementation where `permits_to_be_held - permits_currently_held` is `acquire`d, and two tasks try to `upgrade`
+    /// concurrently (or one `upgrade` in the presence of a `write`).
+    pub fn upgrade(&self, permits_currently_held: usize, permits_to_be_held: usize) -> Acquire<'_> {
+        let mut acquire = Box::pin(self.acquire(permits_to_be_held));
```
jorajeev, line 641: I think we may want to do stricter validations here.  How about requiring:

(1) `permits_currently_held > 0`
(2) `permits_to_be_held > permits_currently_held`

wrappers/parking_lot/parking_lot_impl/src/rwlock.rs
```diff
@@ -171,6 +281,62 @@ impl<T: ?Sized> RwLock<T> {
...
+    #[inline]
+    pub fn upgrade(s: Self) -> RwLockWriteGuard<'a, T> {
+        s.sem.upgrade(1, s.max_readers);
+
+        // When we return, the `UpgradableRwLockSemWrapper` goes out of scope and it's possible for another
+        // task to acquire the upgradable read lock.
+
+        RwLockWriteGuard {
+            permits_acquired: s.max_readers,
+            _upgradable_read_sem: Some(s.upgradable_read_sem),
```
jorajeev: Why do we need to carry this permit into the WriteGuard ?  After the `upgrade` call above, we have all the permits we need, so we can allow someone else to grab this right? (They'll still have to wait for us to release our permits.)

shuttle/src/future/batch_semaphore.rs
sarsko: Neither of these protect against anything. We already allow releasing 0 permits: https://github.com/awslabs/shuttle/blob/8ae5d58f069c0b20d73134cd3f2e8e1aa797f02b/shuttle/src/future/batch_semaphore.rs#L559-L565.

If `permits_to_be_held < permits_currently_held` then this function is just a convoluted `release`

wrappers/parking_lot/parking_lot_impl/src/rwlock.rs
sarsko: Hmm, I need to make a test which does:

> T1: `UpgradableReadGuard`, `upgrade`
> T2: `UpgradableReadGuard`,
> T3: `RwLock::write`

And see who gets precedence of T2 and T3. Right now T3 will "jump ahead" of T2 which seems wrong.

sarsko: You are correct, will change: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=bdf186efc12574e627d3dfaf0bd294bc

shuttle/src/future/batch_semaphore.rs
jorajeev: If `to_be_held < currently_held` then we (1) release all held permits, and get back in the queue, and then (2) wait for the fewer permits.  So it's not the same as `release(currently_held - to_be_held)`

jorajeev: I think we should at least check (2).  We can weaken to `>=` instead of `>` if you feel strongly about it (though I still think we should not support that, because it's likely a user error to call upgrade)

sarsko: Sure, I agree, it is a user error and its easy for the user to protect against

jorajeev (approved):

PR #264, opened 2026-04-18
Bump to rand 0.8.6

sarsko:
Rand < 0.8.6 has a RUSTSEC advisory against it: https://rustsec.org/advisories/RUSTSEC-2026-0097

This PR bumps to 0.8.6 and deprecates wrappers less than 0.8.6.

---


sarsko:
I guess the bench task is set up to keep the lock file around after compiling for the post-merge version, which seems weird, as it disallows bumping versions and identifying performance regressions, but I also don't care enough to investigate, and think it is safe to overlook the failing bench task.

Issue #91, opened 2022-12-21
Make `future::JoinHandle::abort` actually cancel task


bkragl:
The implementation of `abort` in #87 as detaching the task does not seem right, because detached tasks may still continue to run. Consider the following test, which should not fail but currently does.
```rust
#[test]
fn join_handle_abort_bug() {
    check_dfs(
        || {
            let (sender, receiver) = futures::channel::oneshot::channel();
            let t = future::spawn({
                async move {
                    receiver.await.unwrap();
                    panic!("should not get here");
                }
            });
            t.abort();
            sender.send(()).unwrap();
            shuttle::thread::yield_now();
        },
        None,
    );
}
```

(The `yield_now` is needed because otherwise the main task would immediately finish, in which case also the execution finishes because there are no attached tasks left.)

We need a way to actually cancel a task, which drops its continuation and returns a `JoinError` indicating cancellation.

jamesbornholt:
`abort` is funny because it's best-effort and so non-deterministic _when_ the task should stop. The current implementation doesn't do that. Maybe just always canceling the task immediately is a better default, but it can miss executions, in the same way our `wait_timeout` implementations can miss executions.

sarsko:
Fixed in https://github.com/awslabs/shuttle/pull/278

PR #278, opened 2026-04-27
Add support for task abort

jorajeev:
Previously, task abort (added in [PR #87](https://github.com/awslabs/shuttle/pull/87)) only detached the task. As [issue #91](https://github.com/awslabs/shuttle/issues/91) pointed out, this does not match tokio semantics, which state that when `abort()` is called on `JoinHandle` or `AbortHandle`, "the task is signalled to shut down next time it yields at an .await point" (ref: https://docs.rs/tokio/latest/tokio/task/).

So this PR adds a shared AtomicBool flag to the Wrapper around Shuttle futures. When `abort()` is called, the flag is set and the task is woken. On the next poll, Shuttle checks the flag, drops the inner future (allowing destructors to run while TLS is still alive), and returns `Err(JoinError::Cancelled)` to the join handle waiter.

`JoinHandle::drop()` now calls detach instead of abort, matching tokio semantics where dropping a handle lets the task continue running.

The approach here is based on the ideas sketched out by sarsko and jamesbornholt in [PR #150](https://github.com/awslabs/shuttle/pull/150) (which was yanked).

Coauthored with Opus 4.6

---

### DIFF
--- a/shuttle/src/future/mod.rs
+++ b/shuttle/src/future/mod.rs
@@ -14,6 +14,7 @@ use std::future::Future;
 use std::panic::Location;
 use std::pin::Pin;
 use std::result::Result;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::task::{Context, Poll, Waker};
 
@@ -26,9 +27,19 @@ where
 {
     let stack_size = ExecutionState::with(|s| s.config.stack_size);
     let inner = Arc::new(std::sync::Mutex::new(JoinHandleInner::default()));
-    let task_id = ExecutionState::spawn_future(Wrapper::new(fut, inner.clone()), stack_size, None, caller);
+    let aborted = Arc::new(AtomicBool::new(false));
+    let task_id = ExecutionState::spawn_future(
+        Wrapper::new(fut, inner.clone(), aborted.clone()),
+        stack_size,
+        None,
+        caller,
+    );
 
-    JoinHandle { task_id, inner }
+    JoinHandle {
+        task_id,
+        inner,
+        aborted,
+    }
 }
 
 /// Spawn a new async task that the executor will run to completion.
@@ -56,11 +67,26 @@ where
 #[derive(Debug, Clone)]
 pub struct AbortHandle {
     task_id: TaskId,
+    aborted: Arc<AtomicBool>,
 }
 
 impl AbortHandle {
     /// Abort the task associated with the handle.
+    ///
+    /// The task will be cancelled at the next await point. If the task has already
+    /// completed, this is a no-op.
     pub fn abort(&self) {
+        // Scheduling point: the scheduler may run other tasks (including the target)
+        // before the abort flag is set, creating interleavings where the task completes
+        // normally despite an abort() call.
+        thread::switch();
+
+        // Signal the Wrapper to skip the inner future on the next poll.
+        // If already aborted, skip the redundant wake.
+        if self.aborted.swap(true, Ordering::Relaxed) {
+            return;
+        }
+        // Wake the task so Wrapper::poll() runs and performs cleanup.
         let res = ExecutionState::try_with(|state| {
             if !state.is_finished() {
                 state.get_mut(self.task_id).abort();
@@ -90,7 +116,8 @@ unsafe impl Sync for AbortHandle {}
 #[derive(Debug)]
 pub struct JoinHandle<T> {
     task_id: TaskId,
-    inner: std::sync::Arc<std::sync::Mutex<JoinHandleInner<T>>>,
+    inner: Arc<std::sync::Mutex<JoinHandleInner<T>>>,
+    aborted: Arc<AtomicBool>,
 }
 
 #[derive(Debug)]
@@ -110,7 +137,22 @@ impl<T> Default for JoinHandleInner<T> {
 
 impl<T> JoinHandle<T> {
     /// Abort the task associated with the handle.
+    ///
+    /// The task will be cancelled at the next await point. Awaiting the `JoinHandle` after
+    /// calling `abort` will return `Err(JoinError::Cancelled)`, unless the task completed
+    /// before the abort took effect.
     pub fn abort(&self) {
+        // Scheduling point: the scheduler may run other tasks (including the target)
+        // before the abort flag is set, creating interleavings where the task completes
+        // normally despite an abort() call.
+        thread::switch();
+
+        // Signal the Wrapper to skip the inner future on the next poll.
+        // If already aborted, skip the redundant wake.
+        if self.aborted.swap(true, Ordering::Relaxed) {
+            return;
+        }
+        // Wake the task so Wrapper::poll() runs and performs cleanup.
         let res = ExecutionState::try_with(|state| {
             if !state.is_finished() {
                 state.get_mut(self.task_id).abort();
@@ -134,7 +176,10 @@ impl<T> JoinHandle<T> {
 
     /// Returns a new `AbortHandle` that can be used to remotely abort this task.
     pub fn abort_handle(&self) -> AbortHandle {
-        AbortHandle { task_id: self.task_id }
+        AbortHandle {
+            task_id: self.task_id,
+            aborted: self.aborted.clone(),
+        }
     }
 }
 
@@ -158,7 +203,16 @@ impl Error for JoinError {}
 
 impl<T> Drop for JoinHandle<T> {
     fn drop(&mut self) {
-        self.abort();
+        // Detach the task so it keeps running but we don't wait for it.
+        // Dropping a JoinHandle does NOT cancel the task (unlike abort()).
+        let res = ExecutionState::try_with(|state| {
+            if !state.is_finished() {
+                state.get_mut(self.task_id).detach();
+            }
+        });
+        if let Err(e) = res {
+            tracing::error!("`JoinHandle::drop` failed with error: {e:?}");
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


shuttle/src/future/mod.rs
```diff
@@ -134,13 +165,16 @@ impl<T> JoinHandle<T> {
...
+            task_id: self.task_id,
+            aborted: self.aborted.clone(),
+        }
     }
 }
 
 // TODO: need to work out all the error cases here
 /// Task failed to execute to completion.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Hash)]
```
sarsko: Nittish: Not faithful: https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#trait-implementations

shuttle/src/future/mod.rs
```diff
@@ -206,7 +258,34 @@ where
     type Output = ();
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        match self.future.as_mut().poll(cx) {
+        // If abort() was called, skip polling the inner future.
+        if self.aborted.load(Ordering::Relaxed) {
+            if ExecutionState::try_with(|state| state.is_finished()).unwrap_or(true) {
```
sarsko, line 301: If we are finished then keep the future around (ie don't process the abort)? How come? Would be nice with a comment explaining why

shuttle/src/future/mod.rs
```diff
@@ -180,20 +223,29 @@ impl<T> Future for JoinHandle<T> {
...
+// The `aborted` flag is set by `JoinHandle::abort()` / `AbortHandle::abort()`. On the
+// next poll, `Wrapper` detects the flag, drops the inner future (running its destructors),
+// cleans up thread-local storage, publishes `Err(JoinError::Cancelled)` to any awaiting
+// join handle, and returns `Poll::Ready(())`.
 struct Wrapper<F: Future> {
-    future: Pin<Box<F>>,
-    inner: std::sync::Arc<std::sync::Mutex<JoinHandleInner<F::Output>>>,
+    /// The inner future. Wrapped in `Option` so we can drop it explicitly in the
+    /// abort path (before running thread-local destructors).
+    future: Option<Pin<Box<F>>>,
```
sarsko, line 245: Why isnt this an `Pin<Box<Option<F>>>`?

shuttle/src/future/mod.rs
```diff
@@ -206,7 +258,34 @@ where
...
+        // If abort() was called, skip polling the inner future.
+        if self.aborted.load(Ordering::Relaxed) {
+            if ExecutionState::try_with(|state| state.is_finished()).unwrap_or(true) {
+                return Poll::Ready(());
+            }
+
+            // Drop the inner future first so its destructors can still access TLS.
+            // SAFETY: we're dropping a heap-allocated Box through Pin, which is fine
+            // because we never move the Box's contents — we just deallocate it.
+            unsafe {
```
sarsko: Rumor has it the `unsafe` is unnecessary and you can just `take` the `future`

shuttle/src/future/mod.rs
sarsko: Nevermind on this, I was still in the mindset of the `unsafe` below being needed

shuttle/src/future/mod.rs
```diff
@@ -206,7 +258,34 @@ where
...
+                self.as_mut().get_unchecked_mut().future.take();
+            }
+
+            // Now run thread-local destructors, same as the normal completion path.
+            while let Some(local) = ExecutionState::with(|state| state.current_mut().pop_local()) {
```
sarsko: Factor out into a helper ?

shuttle/src/runtime/task/mod.rs
```diff
@@ -481,9 +481,23 @@ impl Task {
         self.detached = true;
     }
 
+    /// Request that this task be aborted. The task is woken so the scheduler can poll it
+    /// and let the Wrapper future observe the abort flag (a shared `AtomicBool`).
+    ///
+    /// Note: we intentionally do NOT detach here. Keeping the task attached ensures the
```
sarsko: This is a weird comment. I don't think anybody would think to detach here.

shuttle/tests/future/basic.rs
```diff
@@ -411,3 +412,450 @@ fn clean_up_detached_task() {
         None,
     )
 }
+
+// --- Task abort tests ---
+
+/// Abort a task before it has a chance to run, then await it.
```
sarsko: Comment is wrong

shuttle/src/runtime/task/mod.rs
```diff
@@ -481,9 +481,23 @@ impl Task {
...
+    /// scheduler will run it (rather than skipping it via the "all runnable tasks are
+    /// detached" exit path), so the abort cleanup (dropping the future, TLS, publishing
+    /// Cancelled) actually executes. Detachment is handled separately by `JoinHandle::drop`.
     pub(crate) fn abort(&mut self) {
```
sarsko, line 485: The `abort` functions need to be scheduling points

shuttle/tests/future/basic.rs
```diff
@@ -411,3 +412,450 @@ fn clean_up_detached_task() {
...
+/// The JoinHandle should resolve with Err(JoinError::Cancelled).
```
sarsko: I think `abort` should have a scheduling point and thus there be schedules where we get the result and schedules where we get `Cancelled`

shuttle/tests/future/basic.rs
```diff
@@ -411,3 +412,450 @@ fn clean_up_detached_task() {
...
+        },
+        None,
+    );
+}
+
+/// Abort a task that has started and is holding a Mutex. The abort should cancel the
+/// task (returning JoinError::Cancelled), drop the future's locals (releasing the lock),
+/// and not execute any code after the await point.
+#[test]
+fn abort_frees_mutex() {
```
sarsko, line 485: I think this is a good test

shuttle/src/future/mod.rs
jorajeev: Yeah, I know - but it does make tests easier to write

jorajeev: Okay, I removed the extra derives (and defined a test enum that derives Hash/Eq for testing).

shuttle/src/future/mod.rs
jorajeev: Added a comment

shuttle/src/future/mod.rs
jorajeev: Yup - fixed

shuttle/src/future/mod.rs
jorajeev: Done - added a `finish` method to the wrapper

shuttle/src/runtime/task/mod.rs
jorajeev: Fixed

shuttle/tests/future/basic.rs
jorajeev: Fixed

shuttle/src/runtime/task/mod.rs
jorajeev: Yeah, added a `thread::switch` at the beginning of `JoinHandle` and `AbortHandle` abort

shuttle/tests/future/basic.rs
jorajeev: Yes, fixed, and updated tests to use a HashSet to make sure we see both outcomes with dfs

shuttle/src/future/mod.rs
```diff
@@ -26,9 +27,19 @@ where
 {
     let stack_size = ExecutionState::with(|s| s.config.stack_size);
     let inner = Arc::new(std::sync::Mutex::new(JoinHandleInner::default()));
-    let task_id = ExecutionState::spawn_future(Wrapper::new(fut, inner.clone()), stack_size, None, caller);
-
-    JoinHandle { task_id, inner }
+    let aborted = Arc::new(AtomicBool::new(false));
+    let task_id = ExecutionState::spawn_future(
+        Wrapper::new(fut, inner.clone(), aborted.clone()),
```
sarsko, line 32: Not the biggest fan of how we've structured tasks and their ownership. I feel like there exists a better design, but that doesn't have to happen in this PR

shuttle/src/future/mod.rs
```diff
@@ -110,7 +137,22 @@ impl<T> Default for JoinHandleInner<T> {
...
+    /// before the abort took effect.
     pub fn abort(&self) {
+        // Scheduling point: the scheduler may run other tasks (including the target)
+        // before the abort flag is set, creating interleavings where the task completes
+        // normally despite an abort() call.
+        thread::switch();
+
+        // Signal the Wrapper to skip the inner future on the next poll.
+        // If already aborted, skip the redundant wake.
+        if self.aborted.swap(true, Ordering::Relaxed) {
```
sarsko, line 152: Ref my comment about the structuring, I think there exists a better structure where we are not accessing `aborted` directly.

In general, ref "cratifying" Shuttle into a core crate: `AbortHandle` is really a shuttle-tokio construct, meaning we should provide an underlying `TaskHandle` which should allow `JoinHandle` and `AbortHandle` etc to be built on top.

sarsko (approved):

sarsko:
Left a couple of comments about refactoring on the basis of having a `core` crate, not blocking the PR on that

PR #279, opened 2026-04-28
Add a Shuttle-compatible version of DashMap

sarsko:
Models DashMap as a `RwLock<HashMap>`, and is thus more coarse-grained than the regular DashMap. I think for modeling purposes this is an improvement over more fine grained modeling, as afaik DashMap provides no guarantees on how it shards things, meaning more fine grained locking could miss double locking issues which this would catch.

Thanks to Drew Schleit for making the initial version of this.

That version used `Mutex`, this one uses `RwLock`.

---

### DIFF
--- a/wrappers/dashmap/Cargo.toml
+++ b/wrappers/dashmap/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "shuttle-dashmap"
+version = "6.1.0"
+edition = "2024"
+license = "Apache-2.0"
+description = "dashmap wrapper for the Shuttle concurrency testing tool."
+
+[features]
+shuttle = [ "dep:shuttle-dashmap-impl" ]
+
+arbitrary = [ "dashmap/arbitrary", "shuttle-dashmap-impl?/arbitrary" ]
+inline = [ "dashmap/inline", "shuttle-dashmap-impl?/inline" ]
+raw-api = [ "dashmap/raw-api", "shuttle-dashmap-impl?/raw-api" ]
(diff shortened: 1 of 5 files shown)


wrappers/dashmap/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "shuttle-dashmap"
+version = "6.1.0"
+edition = "2024"
+license = "Apache-2.0"
+description = "dashmap wrapper for the Shuttle concurrency testing tool."
+
+[features]
+shuttle = [ "dep:shuttle-dashmap-impl" ]
```
eggyal, line 9: Shouldn't this also include dashmap's own cargo features, to transitively enable those if users require?

eggyal: Namely:

```toml
arbitrary = [ "dashmap/arbitrary" ]
inline = [ "dashmap/inline" ]
raw-api = [ "dashmap/raw-api" ]
rayon = [ "dashmap/rayon" ]
serde = [ "dashmap/serde" ]
typesize = [ "dashmap/typesize" ]
```

sarsko: Indeed, thanks for catching this!

wrappers/dashmap/dashmap_impl/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "shuttle-dashmap-impl"
+version = "0.1.0"
+edition = "2024"
```
jorajeev, line 4: We should update all our crates to edition 2024

jorajeev (approved):

PR #286, opened 2026-05-07
Refactor shuttle, part one: extract shuttle-core

jorajeev:
This is the first of four PRs to refactor shuttle as discussed in #249.

The eventual goal is to have 4 sub-crates:
- `shuttle-core`: runtime, scheduler traits, execution engine, annotations, config, thread-local key, batch semaphore, and curated internal API surface
- `shuttle-schedulers`: all scheduler implementations (random, PCT, DFS, round-robin, replay, annotation, URW, uncontrolled nondeterminism)
- `shuttle-sync`: std::sync primitives (mutex, rwlock, condvar, barrier, once, atomics, mpsc)
- `shuttle-exec`:  threads (spawn, park, unpark, yield, sleep) and future (spawn, block_on, yield_now)

This PR extracts the `shuttle-core` crate, moving the runtime, scheduler traits, config, annotations, hint, current, and batch_semaphore into a new shuttle-core crate. The shuttle crate becomes a thin facade that re-exports from shuttle-core.

First step towards addressing #249 

---


wrappers/shuttle_sync/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "shuttle-sync"
+name = "shuttle-sync-wrapper"
```
sarsko: Why is this renamed?

sarsko:
I am fine with this way of doing it, we should make it not fail checks

sarsko:
Actually strike that, this way of doing it sounds like a great way to not keep stuff up to date and to make a headache, why are we not moving stuff out and reexporting? We don't have to move everything all at once

wrappers/shuttle_sync/Cargo.toml
jorajeev: Obsolete since this PR now only extracts shuttle-core.

jorajeev:
> Actually strike that, this way of doing it sounds like a great way to not keep stuff up to date and to make a headache, why are we not moving stuff out and reexporting? We don't have to move everything all at once

On reflection, I agree. I've updated this PR so it now just introduces `shuttle-core`.  I've tried to just move files using git mv, hopefully that makes review easier.

Three upcoming PRs will follow (see main PR description)
- extract `shuttle-schedulers`
- extract `shuttle-sync-impl`
- extract `shuttle-exec` (thread and future)

sarsko:
Why isn't `shuttle-exec` part of `shuttle-core`?

shuttle-core/src/future/batch_semaphore.rs
```diff
@@ -846,54 +846,8 @@ impl crate::annotations::WithName for BatchSemaphore {
...
         &self.signature
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn unique_resource_signature_batch_semaphore() {
```
sarsko, line 860: These are just fully deleted instead of moving to a test crate or something like that?

shuttle-core/src/runtime/task/clock.rs
```diff
@@ -9,29 +9,29 @@ mod vector_clock {
 
     #[derive(Clone, Debug, PartialEq, Eq)]
     pub struct VectorClock {
-        pub(crate) time: SmallVec<[u32; DEFAULT_INLINE_TASKS]>,
+        pub time: SmallVec<[u32; DEFAULT_INLINE_TASKS]>,
```
sarsko, line 12: I'd like to not have `pub` fields when we publish to crates; is this needed and should we already here add a function to remove any direct uses of `time`?

shuttle-core/src/runtime/task/mod.rs
```diff
@@ -48,7 +48,7 @@ use waker::make_waker;
 //   Task, the executor resumes that task's continuation until it yields, which happens when its
 //   thread decides it might want to context switch (e.g., because it's blocked on a lock).
 
-pub(crate) const DEFAULT_INLINE_TASKS: usize = 16;
+pub const DEFAULT_INLINE_TASKS: usize = 16;
```
sarsko, line 51: I don't want this to be `pub`

sarsko: Btw, I understand that we'll be less worried about SemVer for `shuttle-core` (at least for things like un-`pub`ing things like this, not for changing `BatchSemaphore` or similar), it's just a lot easier to keep a good hygiene than to do things later.

I am also fine with doing `pub` cleanup in a follow-up

shuttle-core/src/runtime/task/mod.rs
```diff
@@ -100,8 +100,8 @@ impl<'a> From<&'a TaskName> for &'a String {
 /// # Example
 /// The following example shows how a `ChildLabelFn` can be used to set up names for the next child(ren)
 /// that will be spawned by a parent task.
-/// ```
-/// # use shuttle::current::{me, set_label_for_task, get_name_for_task, ChildLabelFn, TaskName};
+/// ```ignore
```
sarsko, line 103: why `ignore`?

shuttle-core/src/runtime/task/mod.rs
sarsko: Actually nevermind on SemVer, we have to care about it because else Shuttle will break if the core version is wrong

sarsko:
I am fine with merging the CR in its current shape.

Once merged I think we will want to follow up with CRs which un-`pub` fields and do a bit of reorg

jorajeev:
> Why isn't `shuttle-exec` part of `shuttle-core`?

I want to keep `shuttle-core` just the internal stuff.  Note that `thread` and `future` spawns etc will be public APIs

shuttle-core/src/runtime/task/mod.rs
jorajeev: Let's clean these pub fields in a followup, once all the crates are in place.

shuttle-core/src/runtime/task/mod.rs
jorajeev: Because `check_dfs` isn't available in shuttle-core. Kept it as `ignore` for posterity.  (Could also mark it `no_run` I guess.). 

We could also move this out to the wrapper crate, but there's not an obvious home for it.

shuttle-core/src/future/batch_semaphore.rs
jorajeev: They need `check_random`. I moved them to the main crate (under `tests/basic/resource_signatures.rs`)

shuttle-core/src/runtime/task/mod.rs
jorajeev: Added a pointer in the doc comment to the runnable tests (which are in `shuttle/tests/basic/labels.rs`)

shuttle-core/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "shuttle-core"
+version = "0.9.1"
```
sarsko: Version should be 0.1.0

PR #287, opened 2026-05-08
Fix scheduling points to precede sync operations in tokio wrapper

vineetpuranik:
###  What was fixed : 

1. oneshot.rs : When sending a message, closing a channel, or trying to receive, the scheduler was being notified after the action happened. Now, this is moved to before the action happens so that scheduler gets a chance to run other tasks. 

2. notify.rs : notify_one was using .unwrap() when sending on an internal channel, which could panic. This has been changed to let _ = to safely ignore the error, since it seems expected and harmless.

---

### DIFF
--- a/wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
+++ b/wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
@@ -128,8 +128,6 @@ impl Notify {
     /// [`notified().await`] will complete immediately consuming the permit made
     /// available by this call to `notify_one()`.
     pub fn notify_one(&self) {
-        // Scheduling point precedes the visible notification operation
-        shuttle::thread::yield_now();
         let mut state = self.state.lock().unwrap();
         // Need to choose a waiter that is Pending
         let mut pending = Vec::with_capacity(state.waiters.len());
@@ -169,8 +167,6 @@ impl Notify {
     /// already registered waiters. Registering for notification is done by
     /// acquiring an instance of the `Notified` future via calling `notified()`.
     pub fn notify_waiters(&self) {
-        // Scheduling point precedes the visible notification operation
-        shuttle::thread::yield_now();
         let mut state = self.state.lock().unwrap();
         // Notify all waiters, including those not yet enabled
         let waiters = std::mem::take(&mut state.waiters);
--- a/wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
+++ b/wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
@@ -128,6 +128,8 @@ impl Notify {
     /// [`notified().await`] will complete immediately consuming the permit made
     /// available by this call to `notify_one()`.
     pub fn notify_one(&self) {
+        // Scheduling point precedes the visible notification operation
+        shuttle::thread::yield_now();
         let mut state = self.state.lock().unwrap();
         // Need to choose a waiter that is Pending
         let mut pending = Vec::with_capacity(state.waiters.len());
@@ -151,9 +153,11 @@ impl Notify {
             state.pending = false;
             drop(state);
             trace!("notify_one for {:?} waking waiter {:?}", self, waiter.id);
-            // Must set flag before notifying waiter
+            // Must set flag before notifying waiter.
+            // The send may fail if the Notified future is dropped between the flag store
+            // and the send (the flag is already NOTIFIED, so the future already completed).
             waiter.flag.store(NOTIFIED, Ordering::SeqCst);
-            waiter.tx.send(()).unwrap();
+            let _ = waiter.tx.send(());
         }
     }
 
@@ -163,8 +167,10 @@ impl Notify {
     /// `notify_one()`, no permit is stored to be used by the next call to
     /// `notified().await`. The purpose of this method is to notify all
     /// already registered waiters. Registering for notification is done by
-    /// acquiring an instance of the `Notified` future via calling `notified()`.    
+    /// acquiring an instance of the `Notified` future via calling `notified()`.
     pub fn notify_waiters(&self) {
+        // Scheduling point precedes the visible notification operation
+        shuttle::thread::yield_now();
         let mut state = self.state.lock().unwrap();
         // Notify all waiters, including those not yet enabled
         let waiters = std::mem::take(&mut state.waiters);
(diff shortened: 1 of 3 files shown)


wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
```diff
@@ -128,6 +128,8 @@ impl Notify {
     /// [`notified().await`] will complete immediately consuming the permit made
     /// available by this call to `notify_one()`.
     pub fn notify_one(&self) {
+        // Scheduling point precedes the visible notification operation
+        shuttle::thread::yield_now();
```
sarsko: There is no point in adding a scheduling point here. `notify_one` already has two scheduling points (one in `let index = state.rng.gen_range(0..pending.len());` and one in `waiter.tx.send(()).unwrap();`)

sarsko: I think `notify_one` should be reworked entirely and hopefully end up having just one scheduling point. I also think it also jells poorly with having a scheduling point before the operation, since there really isn't a race on getting notified (scheduling points before are mostly to be able to do interesting scheduling on racy acquires of `Mutex`es etc)

wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
```diff
@@ -151,9 +153,11 @@ impl Notify {
             state.pending = false;
             drop(state);
             trace!("notify_one for {:?} waking waiter {:?}", self, waiter.id);
-            // Must set flag before notifying waiter
+            // Must set flag before notifying waiter.
+            // The send may fail if the Notified future is dropped between the flag store
+            // and the send (the flag is already NOTIFIED, so the future already completed).
             waiter.flag.store(NOTIFIED, Ordering::SeqCst);
-            waiter.tx.send(()).unwrap();
+            let _ = waiter.tx.send(());
```
sarsko, line 158: This change looks correct to me. 

I would appreciate a test that hits this, should be possible to create with either `JoinHandle::abort` or `tokio::select!` or `FuturesUnordered` or manually creating and dropping a future.

wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
```diff
@@ -163,8 +167,10 @@ impl Notify {
     /// `notify_one()`, no permit is stored to be used by the next call to
     /// `notified().await`. The purpose of this method is to notify all
     /// already registered waiters. Registering for notification is done by
-    /// acquiring an instance of the `Notified` future via calling `notified()`.    
+    /// acquiring an instance of the `Notified` future via calling `notified()`.
     pub fn notify_waiters(&self) {
+        // Scheduling point precedes the visible notification operation
+        shuttle::thread::yield_now();
```
sarsko: We already have a scheduling point in let _ = w.tx.send(())

wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
sarsko, line 187: Ah look here we do the right thing

wrappers/tokio/impls/tokio/inner/src/sync/oneshot.rs
```diff
@@ -102,19 +102,20 @@ impl<T> Sender<T> {
...
         shuttle::thread::yield_now();
+        self.0.close();
     }
 
     /// Attempts to receive a value.
     pub fn try_recv(&mut self) -> Result<T, TryRecvError> {
-        let out = match self.0.try_recv() {
+        // Scheduling point precedes the visible receive operation
+        shuttle::thread::yield_now();
+        match self.0.try_recv() {
```
sarsko, line 114: All of these look good to me

sarsko (changes requested):
Thanks. I think if the scheduling points in `notify_one` and `notify_waiters` is removed and optionally a test is added to test the channel being closed case then this PR looks good to go to me.

wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
vineetpuranik: Sure, I will add a test for this.

wrappers/tokio/impls/tokio/inner/src/sync/notify.rs
vineetpuranik: Removed the scheduling points.

vineetpuranik:
> Thanks. I think if the scheduling points in `notify_one` and `notify_waiters` is removed and optionally a test is added to test the channel being closed case then this PR looks good to go to me.

@sarsko Thank you for the quick review !!
I removed the unnecessary scheduling points and also added a test.

wrappers/tokio/impls/tokio/inner/tests/notify.rs
```diff
@@ -300,3 +300,23 @@ fn notify_mpmc_no_deadlock() {
...
     notify_mpmc_channel_2_test(false);
 }
+
+// Regression test: `notify_one` must not panic when the `Notified` future is
+// dropped after the waiter's flag is set to NOTIFIED but before the internal
+// oneshot send completes. This race is reachable because the oneshot `Sender::send`
+// now yields before sending, allowing the woken task to run (and drop the receiver)
+// before the send. The fix is `let _ = waiter.tx.send(())` instead of `.unwrap()`.
+#[test]
+fn notify_one_drop_notified_before_send() {
```
sarsko, line 317: Not sure this test is sufficient. If I undo the `unwrap` change and run the test:
```
running 1 test
test notify_one_drop_notified_before_send ... ok
```

vineetpuranik: Yes. you are right. I will try to create this using one of the mechanisms you mentioned in the original comment.

vineetpuranik: @sarsko Here is what I am thinking : 
In the current test method that I have , the receiver is always alive during the send so the race condition is never happening. Now, the goal is to create the race condition by maybe using `handle.abort()` so that the receiver gets dropped mid send.
I am hoping this thought process now aligns more with what we want this test method to simulate. 
Let me know your thoughts and then I can go ahead and write a method and also verify the test actually tests the right behavior by locally modifying to its previous state to catch the bug.

sarsko: As long as the test panics before the change and doesn't panic after the change then I'm happy. The easiest way to make it is probably by using `tokio::select!` in `biased` mode or a `FuturesUnordered` where we poll a `notify_one` future and some other future which resolves, thus dropping the `notify_one` future

vineetpuranik: @sarsko I am able to recreate the behavior in the latest commit and the test also panics when I revert the change made in src/sync/notify.rs locally.

sarsko (approved):
Sorry for the delay, looks good!

PR #290, opened 2026-05-21
Extract shuttle-schedulers crate from shuttle

jorajeev:
This is the second of four PRs to refactor shuttle as discussed in https://github.com/awslabs/shuttle/issues/249.

This PR moves scheduler implementation files (random, pct, dfs, replay, round_robin, urw, annotation, uncontrolled_nondeterminism) into a new `shuttle-schedulers` crate. The shuttle crate re-exports all scheduler types so its public API is unchanged.

---

### DIFF
--- a/shuttle/src/lib.rs
+++ b/shuttle/src/lib.rs
@@ -205,10 +205,8 @@ pub use shuttle_core::{
 // Re-export constants
 pub use shuttle_core::{ANNOTATION_FILE, CAPTURE_BACKTRACE, SILENCE_WARNINGS};
 
-// Re-export internal helpers used by the scheduler impls in this crate
-#[cfg(feature = "annotation")]
-pub(crate) use shuttle_core::annotation_file;
-pub(crate) use shuttle_core::{seed_from_env, silence_warnings};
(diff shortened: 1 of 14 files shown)


shuttle/src/lib.rs
```diff
@@ -205,10 +205,8 @@ pub use shuttle_core::{
 // Re-export constants
 pub use shuttle_core::{ANNOTATION_FILE, CAPTURE_BACKTRACE, SILENCE_WARNINGS};
 
-// Re-export internal helpers used by the scheduler impls in this crate
-#[cfg(feature = "annotation")]
-pub(crate) use shuttle_core::annotation_file;
```
sarsko, line 210: This is breaking no?

sarsko:
One comment lgtm other than that

shuttle/src/lib.rs
jorajeev: Why breaking? It's `pub(crate)` and was originally needed for the annotation scheduler, but that's now moved to `shuttle-schedulers`

sarsko (approved):

PR #291, opened 2026-05-21
feat: add watch::Sender::clone and JoinSet::abort_all methods

tneely:
These are supported in tokio and should be included in the shuttle mock.

https://docs.rs/tokio/latest/src/tokio/sync/watch.rs.html#201-209
https://docs.rs/tokio/latest/src/tokio/task/join_set.rs.html#463-465

---


sarsko (approved):
Thanks Taylor, lgtm!

PR #293, opened 2026-05-22
Refactor shuttle, part 4/4: extract shuttle-exec crate from shuttle

jorajeev:
Superseded by #294 

---

~This is the last of four PRs to refactor shuttle as discussed in https://github.com/awslabs/shuttle/issues/249.~

~Move thread and async task spawning (thread, future) into `shuttle-exec` crate. The `shuttle` crate becomes a thin facade that re-exports all public types from the sub-crate via glob re-exports.~

~Note that `shuttle/src/thread.rs` was moved without changes to `shuttle-exec/src/thread.rs`. But since we still kept a wrapper in `shuttle/src/thread.rs`, github diff doesn't show it as a rename.~

---


shuttle-std/src/thread.rs
```diff
@@ -0,0 +1,394 @@
...
+        ExecutionState::with(|s| {
+            s.get_mut(self.id.task_id).unpark();
+        });
+    }
+}
+
+/// A scope to spawn scoped threads in.
+///
+/// See [`scope`] for details.
+pub struct Scope<'scope, 'env: 'scope> {
```
sarsko, line 57: This should be moved out of shuttle_exec, no?

sarsko: The same applies to most other things here I think, shouldn't most of this be elsewhere than exec>

PR #294, opened 2026-05-22
Refactor shuttle part 4/4, introduce `shuttle-std`

jorajeev:
This is the last of four PRs to refactor shuttle as discussed in https://github.com/awslabs/shuttle/issues/249.

The new layout is as follows:

* `shuttle-core`: runtime, scheduler traits, execution engine, annotations, config, thread-local key, batch semaphore, and curated internal API surface
* `shuttle-schedulers`: all scheduler implementations (random, PCT, DFS, round-robin, replay, annotation, URW, uncontrolled nondeterminism)
* `shuttle-std`: shuttle equivalents for 
   * `std::sync` primitives (mutex, rwlock, condvar, barrier, once, atomics, mpsc) in module `shuttle-std::sync`
   * `thread` and `future` spawn/join and related types (e.g., Scope)

This PR moves the previously created `shuttle-sync` crate into the new `shuttle-std` crate (as a `sync` module), and moves the `thread`/`future` constructs into `shuttle-std`.

In a later PR, we'll look at moving `shuttle/wrappers/collections` into `shuttle-std` as well

Note for review: `shuttle/src/thread.rs` was moved verbatim to `shuttle-std/src/thread.rs` but since we kept around the original `shuttle/src/thread.rs` for redirection, github shows a large copy/paste diff.

---


shuttle-std/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
-name = "shuttle-sync-impl"
+name = "shuttle-std"
 version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
-description = "Shuttle implementations of std::sync primitives"
+description = "Shuttle implementations of std synchronization, threading, and async primitives"
```
sarsko: It's not common to update the decription, this should just be "Shuttle implementations of std primitives" (or for instance "Mirrors of std primitives compatible with the Shuttle concurrency testing tool")

sarsko (approved):

PR #299, opened 2026-06-24
Add support for 128 bit atomics

nwtnni:
Hello,

I've been working on a data structure that uses 128 bit atomics via the [portable-atomic](https://github.com/taiki-e/portable-atomic) crate, and would like to test it under shuttle. I figured the diff was small enough to open a pull request directly, but please let me know if I missed anything!

Thanks,
Newton

---


sarsko:
Hi Newton,

Sorry for the late review, didn't notice the PR until now!

There is an argument to not vend this via `std` and to offer a portable-atomic wrapper instead. However, 128 bit atomics are in std on nightly, and I don't think having them offered all the time from std introduces any problems.

I'll let the tests finish (had to rebase and push) then merge.

nwtnni:
Thank you! I will point https://github.com/nwtnni/arctic to the latest shuttle version whenever it's published.

sarsko:
Sweet, Shuttle 0.9.2 has been released :)

Issue #259, opened 2026-04-02
parking_lot fairness is inadequately modeled (ie., real parking_lot mutexes are not fair)


sarsko:
Ref: https://docs.rs/parking_lot/latest/parking_lot/type.Mutex.html#fairness, https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#fairness:

>A typical unfair lock can often end up in a situation where a single thread quickly acquires and releases the same mutex in succession, which can starve other threads waiting to acquire the mutex. While this improves throughput because it doesn’t force a context switch when a thread tries to re-acquire a mutex it has just released, this can starve other threads.

> This mutex uses [eventual fairness](https://trac.webkit.org/changeset/203350) to ensure that the lock will be fair on average without sacrificing throughput. This is done by forcing a fair unlock on average every 0.5ms, which will force the lock to go to the next thread waiting for the mutex.

> Additionally, any critical section longer than 1ms will always use a fair unlock, which has a negligible impact on throughput considering the length of the critical section.

> You can also force a fair unlock by calling `MutexGuard::unlock_fair` when unlocking a mutex instead of simply dropping the MutexGuard.

Alternatives to solve this:
1. Keep the current fairness, meaning we'll miss bugs (but all bugs are real bugs).
2. Swap to `Unfair`, meaning we'll be complete but not sound
3. Model this "holder is in the running to reacquire the mutex" somehow.

1 and 2 are a lot less work, but either not complete or not sound.

Modeling "holder is in the running to reacquire the mutex" is doable, but has some trickiness to it. Releasing a `Mutex`/`RwLock` would be changed to not give the lock to the next task immediately, but instead register that the release has happened, and then check for pending releases whenever a task does something which causes a yield point. Also
> Additionally, any critical section longer than 1ms will always use a fair unlock, 

has to be accommodated somehow. The obvious option here is time models, but a time model which doesn't increment the clock (eg., Frozen) will not ever consider a CS longer than 1ms even when it probably should. The alternative then is scheduling decisions, though we don't track those at a per-task level, and its weird to consider a CS to be 1ms if a task is never blocked, it just happens to be not scheduled for a while, or some other metric we track on a per-task level.

All of this tracking seems like too much ado for nothing. I think the better option is to do the following:

Lock aquire gets changed from the following (needless stuff removed):
```rust
pub fn lock(&self) -> MutexGuard<'_, T> {
       self.semaphore.acquire_blocking(1).unwrap();
[...]

PR #302, opened 2026-07-28
feat(parking_lot): Rebuild Shuttle wrapper on lock_api

bkragl:
Reimplement the `shuttle-parking_lot` wrapper as a thin layer over the [`lock_api`](https://docs.rs/lock_api) crate, mirroring how the upstream [`parking_lot`](https://docs.rs/parking_lot) crate is structured. Rather than hand-rolling `Mutex`/`RwLock` and their guards, provide Shuttle-backed raw locks (`RawMutex`, `RawRwLock`) built on `BatchSemaphore` that implement `lock_api`'s raw traits; the user-facing `Mutex`/`RwLock`, mapped guards, and `Arc`-based guards then come from `lock_api` for free.

This adds the previously-missing `Arc` guards (`lock_arc`, `read_arc`, `write_arc`), mapped guards, and upgrade/downgrade support, and exposes `RawMutex`/`RawRwLock` so consumer code that names those types compiles under Shuttle.

Guard `Send`-ness follows parking_lot's `send_guard` feature (`GuardNoSend` by default, `GuardSend` with `send_guard`), keeping the `Send` contract identical to production. The code is split into `raw_mutex.rs`, `raw_rwlock.rs`, `mutex.rs`, and `rwlock.rs` with explicit crate-root re-exports, matching parking_lot's own module layout.

---

### DIFF
--- a/wrappers/parking_lot/parking_lot_impl/Cargo.toml
+++ b/wrappers/parking_lot/parking_lot_impl/Cargo.toml
@@ -6,14 +6,20 @@ license = "Apache-2.0"
 description = "Internal implementation of the `parking_lot` crate for use with the Shuttle concurrency testing tool. Do not depend on this crate directly; use `shuttle-parking_lot`."
 
 [features]
-arc_lock = []
+# Enable `Arc`-based lock guards (`lock_arc`, `read_arc`, `write_arc`, ...). Forwarded to `lock_api`.
+arc_lock = ["lock_api/arc_lock"]
+# `serde`/`owning_ref`/`nightly` are forwarded to `lock_api`, matching `parking_lot`'s feature set.
+serde = ["lock_api/serde"]
+owning_ref = ["lock_api/owning_ref"]
+nightly = ["lock_api/nightly"]
+# `send_guard` makes lock guards `Send` (when the data allows), mirroring `parking_lot`; see the
+# `GuardMarker` impls in mutex.rs/rwlock.rs. The remaining two features concern the native
+# `parking_lot_core` backend, which Shuttle replaces entirely, so they have no effect here.
+send_guard = []
 deadlock_detection = []
 hardware-lock-elision = []
-nightly = []
-owning_ref = []
-send_guard = []
-serde = []
 
 [dependencies]
 shuttle = { path = "../../../shuttle", version = "<0.10"}
+lock_api = "0.4"
 tracing = { version = "0.1.36", default-features = false, features = ["std"] }
--- a/wrappers/parking_lot/parking_lot_impl/src/raw_mutex.rs
+++ b/wrappers/parking_lot/parking_lot_impl/src/raw_mutex.rs
@@ -1,0 +1,66 @@
+//! The Shuttle-backed raw mutex that underpins [`crate::Mutex`].
+//!
+//! This provides [`RawMutex`], an implementation of [`lock_api::RawMutex`] (and the fair variant)
+//! routed through Shuttle's scheduler. The user-facing `Mutex`/`MutexGuard`/`ArcMutexGuard` are the
+//! generic `lock_api` types specialised to this raw lock (see [`crate::Mutex`]), exactly how the
+//! real `parking_lot` crate layers its `Mutex` on top of `lock_api`.
+
+use shuttle::future::batch_semaphore::{BatchSemaphore, Fairness};
+use std::thread;
+use tracing::trace;
+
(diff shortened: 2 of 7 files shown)


wrappers/parking_lot/parking_lot_impl/src/raw_mutex.rs
```diff
@@ -0,0 +1,66 @@
...
+    unsafe fn unlock(&self) {
+        trace!("releasing parking_lot mutex {:p}", self);
+        self.semaphore.release(1);
+    }
+}
+
+// Safety: Shuttle's semaphore is strictly fair, so a plain `release` already hands the permit to the
+// next waiter in FIFO order. Fair unlocking is therefore identical to a normal unlock.
+unsafe impl lock_api::RawMutexFair for RawMutex {
+    unsafe fn unlock_fair(&self) {
```
sarsko, line 62: Ah yes, true, parking_lot mutexes are not fair (https://github.com/awslabs/shuttle/issues/259).

I wonder if `BatchSemaphore` should be updated such that fairness is a parameter to `acquire`/`release` and not a parameter to the `BatchSemaphore` itself (or both, ie. offer functions with a fairness parameter as well)

wrappers/parking_lot/parking_lot_impl/Cargo.toml
```diff
@@ -6,14 +6,20 @@ license = "Apache-2.0"
 description = "Internal implementation of the `parking_lot` crate for use with the Shuttle concurrency testing tool. Do not depend on this crate directly; use `shuttle-parking_lot`."
 
 [features]
-arc_lock = []
+# Enable `Arc`-based lock guards (`lock_arc`, `read_arc`, `write_arc`, ...). Forwarded to `lock_api`.
```
sarsko, line 9: I like having features in alphabetical order

sarsko: Maybe this order matches `parking_lot` order on GitHub. I prefer having it match the order given by docs.rs (post cargo canonicalization) because it will never change order

wrappers/parking_lot/parking_lot_impl/src/raw_rwlock.rs
```diff
@@ -0,0 +1,208 @@
...
+        true
+    }
+
+    unsafe fn unlock_upgradable(&self) {
+        trace!("releasing parking_lot rwlock {:p} (upgradable)", self);
+        self.sem.release(1);
+        self.upgradable_sem.release(1);
+    }
+
+    unsafe fn upgrade(&self) {
```
sarsko, line 156: I need to come back and understand how the `upgrade` flow is supposed to go and why it is guaranteed that we hold an `upgradable_sem` permit when this is called

wrappers/parking_lot/parking_lot_impl/src/raw_mutex.rs
bkragl: you're considering both fair and unfair given a FIFO queue of waiters, right? another dimension would be using some other ordering for waiters (not every lock impl uses FIFO i think).

wrappers/parking_lot/parking_lot_impl/Cargo.toml
bkragl: the order here is kind of a topical grouping. i now see that `parking_lot` declares a `default = []` feature. do we need that too?

sarsko: `default` is only needed if it contains something.

Topical grouping is worse than alphabetical, we just want to be able to quickly check that we haven't forgotten any features. It matters less for the `impl` crate than the "wrapper" crate

wrappers/parking_lot/parking_lot_impl/src/raw_rwlock.rs
sarsko: Hmm okay, I went through and looked at the flow now. The way it is guaranteed is by having the function be `unsafe`, and then requiring it as a safety precondition to call the function. I have to admit that I am not convinced of the `lock_api` being the right set of traits for locks (I don't think a bunch of `unsafe` traits with `unsafe` functions where the `unsafe` could be removed with types is a good set of public traits, though I think they are completely fine as private traits), but that is beside the point on whether we should support them, and whether we have implemented the functions correctly.

The implementations are afaict correct, but I wouldn't be surprised if there is a bug lurking. I find these traits to be hard to audit.

sarsko:
Clippy is fixed on mainline, merging

PR #308, opened 2026-08-06
Rename shuttle-core crate to shuttle-engine

sarsko:
Collision with https://crates.io/crates/shuttle-core.

We need to rename in order to publish to crates.io.

---


bkragl:
https://crates.io/crates/shuttle-runtime :(

shuttle-engine/src/runtime/task/labels.rs
```diff
@@ -66,7 +66,7 @@ impl Labels {
     /// # Example
     ///
     /// ```
-    /// # use shuttle_runtime::current::Labels;
+    /// # use shuttle_engine::current::Labels;
```
bkragl, line 69: most users will get this through re-export from the `shuttle` crate. when they look at documentation, will the examples then show `shuttle` or still `shuttle-runtime`?

bkragl (approved):

PR #319, opened 2026-08-13
(shuttle-tokio): feat(mpsc): Implement poll_recv for ReceiverInternal

sarsko:
poll_recv was previously unimplemented. The new implementation
stores a persistent Acquire future in ReceiverInternal so the
semaphore waiter stays registered across polls. When the channel
is empty, the caller is properly suspended via the semaphore
waiter queue until a message arrives or the channel closes.

recv, try_recv, and blocking_recv are reimplemented atop poll_recv
so that semaphore interaction is centralized in a single code path.
try_recv polls once with a noop waker and clears the pending acquire
on Pending; recv uses poll_fn; blocking_recv uses block_on(recv).

The Acquire future borrows recv_semaphore behind Arc<Channel>,
with the lifetime transmuted to 'static. This is sound because
pending_acquire is always cleared before chan is dropped.

Add parity tests for bounded and unbounded poll_recv, plus a test
that interleaves poll_recv with recv to verify no permit leak.

---

### DIFF
--- a/wrappers/tokio/impls/tokio/inner/src/sync/mpsc.rs
+++ b/wrappers/tokio/impls/tokio/inner/src/sync/mpsc.rs
@@ -3,10 +3,12 @@
 
 use shuttle::future::{
     self,
-    batch_semaphore::{BatchSemaphore, Fairness, TryAcquireError},
+    batch_semaphore::{Acquire, BatchSemaphore, Fairness, TryAcquireError},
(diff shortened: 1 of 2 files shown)


sarsko:
Merging since we've already had this functionality internally for half a year or so and I would like to upstream it so that we can stop having this functionality be internal only.

PR #326, opened 2026-08-22
Rewrite README to reflect the current state of Shuttle

sarsko:
This CR is borderline draft (because it is AI slop), but I'm leaving it open because I do think it or a version of it will want to be merged. I have yet to review this CR properly.

My opinions:
1. Current README does not accurately present current Shuttle.
2. The new tool to find tools is LLMs. What the LLMs say is heavily based on what is in the README, and we want the LLMS to give stuff that is useful to the person reading. Blabla randomized is not useful, blabla can run most tokio applications and find bugs is useful.
3. People occasionally read READMEs too. The README should be valuable to them too.

AI SLOP BELOW
----

The README had drifted a long way from the library it describes. It was 82 lines, roughly 60 of which were a near-verbatim copy of the crate-level docs in `shuttle/src/lib.rs` (the same `std` `Mutex` example, the same `check_random` rewrite, the same Loom paragraph), plus badges and a single sentence about the wrappers.

Nothing in it was wrong. But it said nothing about async, the replay workflow, seven of the eight schedulers, `Config`, the `SHUTTLE_*` environment variables, labels, the `vector-clocks` or `annotation` features, Shuttle Explorer, the 0.9.2 split into `shuttle-engine`/`shuttle-schedulers`/`shuttle-std`, or tokio beyond one passing mention.

Since `shuttle/Cargo.toml` sets `readme = "../README.md"`, this is also the crates.io landing page.

## What changed

Reordered around "what can I test, and how do I debug a failure", and pulled the useful onboarding content that only existed in `lib.rs` into the README.

- **A section on testing tokio code**, the largest addition. Tokio support is a dependency swap rather than a module, which is non-obvious enough that a one-line mention doesn't get anyone there. Covers the `package = "shuttle-tokio"` setup, `#[tokio::test]`, which tokio modules are actually modeled, time and timeouts, what is re-exported from real tokio and therefore unsafe under Shuttle, the environment variables the harness reads, and the companion crates.
- **The replay workflow**, previously only in the crate docs, despite being the strongest reason to reach for Shuttle.
- **A table of all eight built-in schedulers** with their entry points and when to use each.
- **`Config` knobs and the `SHUTTLE_*` environment variables.**
- **A coverage table for the `std` replacements**, with the caveats stated rather than left to be discovered: `Arc` is std's, the `atomic` implementation is unsound, `fence(Relaxed)` panics, time is not modeled, no `OnceLock`/`LazyLock`/`mpmc`.
- **The other wrappers, Shuttle Explorer, and a repository layout table.**
- Moved the Loom soundness–scalability paragraph into the intro, where it belongs; it was orphaned just above `## License`.

No code changes. `shuttle/src/lib.rs` docs are deliberately untouched.

## Two things called out as design decisions

Some of this is judgement, so flagging it explicitly rather than burying it:

1. **The README and the `lib.rs` doc block are still independent near-duplicates.** There is no `include_str!` anywhere, which is why they drifted. This PR does not address that; it treats the README as the pitch-and-orientation document and leaves `lib.rs` as the API narrative. If you would rather have a single source of truth via `#![doc = include_str!("../README.md")]`, that is a bigger change (every code fence becomes a doctest and needs annotating, and intra-doc links become URLs) and probably its own PR.
4. **The tokio section is deliberately long.** It documents the sharp edges, in particular that `timeout` ignores its duration entirely and only fires via `trigger_timeouts`, and that `task_local!` currently gives shared storage across all Shuttle tasks. Happy to trim it down to a pointer at `wrappers/tokio/README.md` if you'd prefer the README stay short.

## Verification

Every technical claim was checked against the source rather than carried over from the old text. Specifically avoided documenting `shuttle::check` (`#[doc(hidden)]`) and the tokio crate's `check`/`default_shuttle_config` (both `#[deprecated]` and slated to move).

- All relative links resolve, all heading anchors match, all code fences balance.
- The four wrapper versions in the tokio examples were cross-checked against the crates' actual manifests.
- No MSRV or platform-support claims, since none is declared anywhere in the repo and CI is Linux/stable+beta only.

## Relation to other open PRs

The tokio examples here say `version = "1"` for `shuttle-tokio`, which is what #327 republishes it as, and which matches what `wrappers/README.md` already tells users to write. The rest of the README stands alone.

This PR touches only `README.md`, so it does not conflict with #325 or #327 and can merge in any order; merging it before #327 would just mean the documented requirement lands slightly ahead of the release.

## Not included

`wrappers/README.md` documents a `tokio-version-importer-do-not-use-directly` crate that does not exist in the tree. Left alone here. (The `shuttle_enabler` gap in that file is addressed in #325.)

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,22 +6,22 @@
 
 Shuttle is a library for testing concurrent Rust code. It takes control of the scheduler, so that
 thread interleavings are chosen by Shuttle rather than by the OS. That gives you two things a normal
-concurrency test cannot: interleavings are explored deliberately instead of by luck, and any failure
-you find can be replayed deterministically.
+concurrency test cannot: interleavings are explored systematically, under a policy you choose, and
+any failure you find can be replayed deterministically.
 
-Shuttle implements a number of *randomized concurrency testing* techniques, including
-[A Randomized Scheduler with Probabilistic Guarantees of Finding Bugs](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/asplos277-pct.pdf)
-(PCT).
+Shuttle is inspired by [Loom](https://github.com/tokio-rs/loom), but defaults to randomized testing
+rather than exhaustive testing. This is a soundness–scalability trade-off: randomized testing is not
+sound (a passing Shuttle test does not prove the code is correct), but it scales to much larger test
+cases than exhaustive exploration does. Empirically, randomized testing is successful at finding most
+concurrency bugs, which tend not to be adversarial. Shuttle *can* also run exhaustively, through
+[`check_dfs`](#choosing-a-scheduler), but it implements no partial order reduction, so exhaustive
+runs stay tractable only for very small tests.
 
-Shuttle is inspired by [Loom](https://github.com/tokio-rs/loom), but focuses on randomized testing
-rather than exhaustive testing. This is a soundness–scalability trade-off: Shuttle is not sound (a
-passing Shuttle test does not prove the code is correct), but it scales to much larger test cases
-than Loom. Empirically, randomized testing is successful at finding most concurrency bugs, which
-tend not to be adversarial.
-
-Beyond the core library, this repository houses a collection of drop-in
-[wrappers](/wrappers) for popular crates, so that existing code can be run under Shuttle without
-being rewritten. The largest of these is [tokio support](#testing-tokio-code).
+A core goal of Shuttle is to require minimal changes to the code under test: you should be able to
+point Shuttle at existing code rather than restructure that code for testing. To that end, beyond
+the core library, this repository houses a collection of drop-in [wrappers](/wrappers) for popular
+crates, which swap out a dependency rather than asking you to rewrite the code that uses it. The
+largest of these is [tokio support](#testing-tokio-code).
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,11 +4,24 @@
 [![docs.rs](https://docs.rs/shuttle/badge.svg)](https://docs.rs/shuttle)
 [![Tests](https://github.com/awslabs/shuttle/actions/workflows/tests.yml/badge.svg)](https://github.com/awslabs/shuttle/actions/workflows/tests.yml)
 
-Shuttle is a library for testing concurrent Rust code. It is an implementation of a number of
-*randomized concurrency testing* techniques, including
-[A Randomized Scheduler with Probabilistic Guarantees of Finding Bugs](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/asplos277-pct.pdf).
+Shuttle is a library for testing concurrent Rust code. It takes control of the scheduler, so that
+thread interleavings are chosen by Shuttle rather than by the OS. That gives you two things a normal
+concurrency test cannot: interleavings are explored deliberately instead of by luck, and any failure
```
sarsko: Not a fan of this luck term

README.md
```diff
@@ -4,11 +4,24 @@
...
+you find can be replayed deterministically.
 
-This repository also houses a collection of [wrappers](/wrappers) and their implementations. These make it easier to test code using popular libraries such as `tokio` and `rand`, as well as synchronization primitives from `std::sync` and collections from `std::collections`. For more information on how to use these wrappers, see the [wrappers/README.md](/wrappers/README.md).
+Shuttle implements a number of *randomized concurrency testing* techniques, including
+[A Randomized Scheduler with Probabilistic Guarantees of Finding Bugs](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/asplos277-pct.pdf)
```
sarsko: This should be moved down to the section with schedulers

README.md
```diff
@@ -4,11 +4,24 @@
...
+(PCT).
+
+Shuttle is inspired by [Loom](https://github.com/tokio-rs/loom), but focuses on randomized testing
```
sarsko: Mention that Shuttle has seen more love than Loom in recent years? Also mention that Shuttle can also be run exhaustively? (though without any POR)

README.md
```diff
@@ -4,11 +4,24 @@
...
+rather than exhaustive testing. This is a soundness–scalability trade-off: Shuttle is not sound (a
+passing Shuttle test does not prove the code is correct), but it scales to much larger test cases
+than Loom. Empirically, randomized testing is successful at finding most concurrency bugs, which
+tend not to be adversarial.
+
+Beyond the core library, this repository houses a collection of drop-in
```
sarsko: I think it should be explicit that the core philosophy of Shuttle is to require minimal changes to the code

README.md
```diff
@@ -63,15 +67,383 @@ shuttle::check_random(|| {
...
+
+[dependencies]
+shuttle = { version = "0.9", optional = true }
+```
+
+Then funnel all synchronization imports through a single module that switches on the feature:
+
+```rust
+#[cfg(all(feature = "shuttle", test))]
+use shuttle::{sync::*, thread};
```
sarsko: This is not how we want people to swap imports. Use the wrapper crates instead

README.md
```diff
@@ -63,15 +67,383 @@ shuttle::check_random(|| {
...
+#[cfg(not(all(feature = "shuttle", test)))]
+use std::{sync::*, thread};
+```
+
+and run your Shuttle tests with `cargo test --features shuttle`.
+
+## Reproducing a failure
+
+When a Shuttle test fails, it prints the schedule that caused the failure:
```
sarsko, line 111: Note that schedulers like the random scheduler will also print the seed ?

README.md
```diff
@@ -63,15 +67,383 @@ shuttle::check_random(|| {
...
+- **The `atomic` implementation is unsound and may miss bugs.** Shuttle warns about this at runtime;
+  `Config::silence_warnings` suppresses the warning.
+- **`fence(Ordering::Relaxed)` panics**, since there is no such thing as a relaxed fence.
+- **Time is not modeled.** `thread::sleep` and `park_timeout` are scheduling points, not delays; a
+  test where every thread is parked with a timeout is reported as a deadlock.
+- **No `OnceLock`, `LazyLock`, or `mpmc`** replacements yet.
+- Vector clocks, which power the causality tracking exposed through `shuttle::current`, are behind
+  the `vector-clocks` feature.
+
+## Choosing a scheduler
```
sarsko, line 180: I wonder if the uncontrolled nondeterminism check scheduler should have its own little section

PR #328, opened 2026-08-25
Fix the weekly cron job and move workflows off Node 20

sarsko:
The `Cron jobs` workflow has failed on every scheduled run since at least mid-June ([latest](https://github.com/awslabs/shuttle/actions/runs/31987845860), [a month earlier](https://github.com/awslabs/shuttle/actions/runs/29713604671)). There were two independent causes, and one hid the other.

### 1. `cargo audit` fails on an unreachable advisory

The `audit` job exits 1 on [RUSTSEC-2022-0040](https://rustsec.org/advisories/RUSTSEC-2022-0040) (`owning_ref` 0.4.1, multiple soundness issues, no fixed upgrade available).

The advisory is not reachable in anything we build. `owning_ref` enters `Cargo.lock` only as an *optional* dependency of `lock_api`, which is reached three ways: transitively via `parking_lot` 0.12 and `dashmap` 6, and directly by `shuttle-parking_lot-impl`. In every case pulling it in requires the non-default `owning_ref` feature, which nothing in this workspace enables. `cargo tree -i owning_ref --workspace` reports no match, confirming it is absent from the resolved build graph. `cargo audit` flags it regardless because it scans the lockfile, which records optional dependencies whether or not their feature is activated.

Since there is no fixed version to upgrade to, this adds `.cargo/audit.toml` ignoring the advisory, with the reachability analysis and the conditions for removing the entry recorded next to it.

### 2. Nobody was told, because the notifier was broken too

The workflow set a top-level `permissions: contents: write`. Naming any scope [implicitly sets every unnamed scope to `none`](https://docs.github.com/en/enterprise-cloud@latest/actions/using-jobs/assigning-permissions-to-jobs), which left `issues: none`, so the notify step could not open an issue and failed with `Resource not accessible by integration`.

This drops to `contents: read` at the top level, which is all either job needs, and grants `issues: write` on the jobs that notify.

The notify step was also only wired to the `beta` job, so the `audit` job — the one actually failing — had no notification path at all. It now has one, under its own label: the action reuses an open issue carrying the configured label, so a shared label would file an audit failure as a comment on a beta toolchain issue.

### 3. Node 20 action runtime

Every run carried `Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v2, jayqi/failed-build-issue-action@v1`. Node 20 is removed from the runners on **2026-09-16**, at which point these stop working.

- `actions/checkout` v2/v3 → v5, the first major declaring `using: node24`. Deliberately not v7: v6 changed checkout to persist git credentials to a separate file, and `bench.yml` uses `boa-dev/criterion-compare-action@v3`, which last shipped in 2022 and runs its own git operations against the base branch. v5 clears the deadline without taking on that behavioural change.
- The notifier tracked `jayqi/failed-build-issue-action@v1`, which resolved to v1.2.0 on node20. Upstream has since released v1.3.0 on node24 and `v1` now points at it. The action also moved from `jayqi/` to `drivendataorg/`; the old path still redirects, but this references the canonical owner rather than relying on that.

### Unrelated fixes to workflows touched anyway

- `release.yml` had no `name:` and no `permissions:` block at any level, so it displayed by file path and inherited the repository default token scope.
- The `clippy` job in `tests.yml` ran `rustup component add rustfmt`, so it never installed the component it is named for. It only worked because the runner image ships clippy already.

### Testing

`cargo audit` reproduced locally against cargo-audit 0.22.2: exit 1 before, exit 0 after, with output matching CI run 31987845860 exactly. All four workflows re-parsed and permissions verified per job.

Note that `crons.yml` only triggers on `schedule` and on pushes to `main` touching `Cargo.toml`, so CI on this PR will not exercise the audit job. It is worth a manual `gh workflow run` (or waiting for the Monday run) after merge to confirm green.

---

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,24 @@
+# Configuration for `cargo audit`, run by the weekly cron in
+# .github/workflows/crons.yml. See:
+# https://github.com/rustsec/rustsec/blob/main/cargo-audit/audit.toml.example
+
+[advisories]
+ignore = [
+    # RUSTSEC-2022-0040: multiple soundness issues in `owning_ref`.
+    #
+    # Not reachable in any build we produce. `owning_ref` enters Cargo.lock only
+    # as an *optional* dependency of `lock_api`, which is reached three ways:
+    # `parking_lot` 0.12 and `dashmap` 6 (transitively), and
+    # `shuttle-parking_lot-impl` (directly). In every case pulling in
+    # `owning_ref` requires the non-default `owning_ref` feature, which nothing
+    # in this workspace enables. `cargo tree -i owning_ref --workspace` reports
+    # no match, confirming it is absent from the resolved build graph; `cargo
+    # audit` flags it anyway because it scans the lockfile, which records
+    # optional dependencies whether or not their feature is activated.
+    #
+    # The advisory has no fixed upgrade available, so it cannot be resolved by
+    # bumping. Revisit if `lock_api` drops the `owning_ref` dependency, or if
+    # this workspace ever enables the `owning_ref` feature by default -- at that
+    # point the code would genuinely be exposed and this entry must be removed.
+    "RUSTSEC-2022-0040",
+]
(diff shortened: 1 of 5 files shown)


sarsko:
Merging since all of this is touching build scripts and whatnot. No users of Shuttle would notice if any of this is wrong, but I notice that it is wrong on the weekly run.

PR #333, opened 2026-09-01
Fix npm audit vulnerabilities in shuttle-explorer

sarsko:
Clears the 4 open Dependabot alerts on `shuttle-explorer/package-lock.json` (52, 48, 28, 42). All are dev-scope transitive deps, so nothing here affects the crate or the extension at runtime.

- `npm audit fix` bumps brace-expansion 1.1.11 -> 1.1.18 and 2.0.2 -> 2.1.4, plus ajv, cross-spawn, diff and js-yaml.
- `serialize-javascript` has no upstream fix path: it comes from mocha via `@vscode/test-cli`, and even the latest test-cli (0.0.15) pins `mocha ^11.7.6`, whose range is `serialize-javascript ^6.0.2` -- every 6.x is vulnerable. Pinned to 7.1.1 with an `overrides` entry. That is the same major mocha 12 moved to, and mocha's only use is `serializeJavascript(options)` for worker args.

`npm audit` now reports 0 vulnerabilities. `npm run compile` passes (pre-existing lint warnings only), and `mocha --parallel` -- the code path that actually calls serialize-javascript -- runs green on 7.1.1.

### DIFF
--- a/shuttle-explorer/package-lock.json
+++ b/shuttle-explorer/package-lock.json
@@ -527,10 +527,11 @@
       }
     },
     "node_modules/@eslint/eslintrc/node_modules/brace-expansion": {
-      "version": "1.1.11",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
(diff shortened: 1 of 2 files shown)


sarsko:
lgtm

PR #334, opened 2026-09-01
Fix the annotation feature

sarsko:
Pure Claude. PR looks good to me.

----
`cargo build --features annotation` has been failing for a while -- nothing in CI builds the feature, so it rotted behind a few refactors:

- `VectorClock` has no `time` field in its stubbed form, so the `Serialize` impl went through the `Deref<Target = [u32]>` that both forms provide instead.
- `ExecutionState::try_with` returns a `Result` now, not an `Option`.
- `annotation_file` was never imported into the `cfg_if` block.

Beyond compiling, the feature also emitted useless data: `annotation` did not turn on `vector-clocks`, so a consumer building with just `features = ["annotation"]` got the zero-sized `VectorClock` stub and every clock in the schedule serialized as `[]`. Shuttle Explorer reads those clocks to show causal dependence between tasks (`clockCmp` in `shuttle-explorer/src/common/clocks.mts`), so it would have shown none. `annotation` now implies `vector-clocks`, and a `compile_error!` rejects combining it with `bench-no-vector-clocks`, which would override that.

The empty-clock bug is invisible to the test suite, because the integration tests already get vector clocks from the `shuttle = { path = ".", features = ["vector-clocks"] }` dev-dependency. I verified it against a scratch crate outside the workspace depending on `shuttle` with only `annotation`: clocks were `[]` before, populated after.

Also adds:
- `shuttle/tests/basic/annotation.rs`, checking the emitted schedule against the schema in `shuttle-explorer/src/common/schedule.mts` -- task/object/event shape, task names from `set_name_for_task`, non-empty vector clocks, and that a contended mutex produces each expected event kind with in-range IDs.
- a `Tests (annotation)` CI job and a clippy run with the feature, so this cannot rot again.

Verified: `cargo nextest run --release --workspace --features annotation` is 652/652 green, the default run is 649/649, and doc tests, `cargo doc`, fmt and clippy all pass with `RUSTFLAGS=-Dwarnings`.

Two things I deliberately left out:

- No test for `annotate_replay`. It needs an encoded schedule recovered from a failing run, and that mechanism is already broken for a separate reason -- several tests in `shuttle/tests/basic/replay.rs` are `#[ignore]`d with "the schedule is not emitted in the panic output". I would rather not hardcode a brittle schedule string; worth revisiting when replay is fixed.
- Note that `cargo test --release --workspace` (as opposed to nextest) fails intermittently on this repo with or without my changes -- I saw `batch_semaphore_test_2` fail on unmodified main, and a tracing-subscriber "cloned a span that already closed" panic in `mpsc_many_senders_with_blocking`. It looks like cross-test global state that nextest's process-per-test isolation hides. Pre-existing and separate from this PR.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,6 +30,20 @@ jobs:
       - name: cargo test --doc
         run: cargo test --release --doc --workspace
 
+  test-annotation:
+    env:
+      RUST_LOG: off
+    name: Tests (annotation)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        run: rustup update stable
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+      - name: cargo nextest
+        run: cargo nextest run --release --workspace --features annotation
+
   fmt:
     name: rustfmt
     runs-on: ubuntu-latest
@@ -53,6 +67,8 @@ jobs:
         run: rustup component add rustfmt
       - name: clippy
         run: cargo clippy --all-targets -- -D clippy::all
+      - name: clippy (annotation)
+        run: cargo clippy --all-targets --features annotation -- -D clippy::all
 
   docs:
     name: Docs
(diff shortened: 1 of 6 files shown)


sarsko:
Merging since it only touches annotation and is a fix to make it compile. Only potentially controversial part is that we build and test the annotation feature, but the steps run in parallel and its not the step that is going to take the most time.

PR #335, opened 2026-09-03
Make shuttle-tokio's `full` feature match tokio's

bkragl:
`full` in shuttle-tokio-impl only forwarded `full` to its dependencies and never enabled the crate's own feature flags, so the `cfg(feature = ...)`-gated re-exports (`fs`, `signal`, `task_local`, `join`, `main`, `try_join`) were silently missing when building with `full`. Define `full` as the same list of features as real tokio.

Also add a `parking_lot` pass-through feature to shuttle-tokio so it remains a drop-in replacement for users who enable that tokio feature explicitly.

---

### DIFF
--- a/wrappers/tokio/wrappers/shuttle-tokio/Cargo.toml
+++ b/wrappers/tokio/wrappers/shuttle-tokio/Cargo.toml
@@ -21,6 +21,10 @@ io-util = ["tokio/io-util", "shuttle-tokio-impl?/io-util"]
 io-std = ["tokio/io-std", "shuttle-tokio-impl?/io-std"]
 macros = ["tokio/macros", "shuttle-tokio-impl?/macros"]
 net = ["tokio/net", "shuttle-tokio-impl?/net"]
+# Pass-through for tokio's internal performance feature, so `shuttle-tokio`
+# stays a drop-in replacement. Irrelevant to the Shuttle implementation, so
+# there is no matching feature in the impl crate.
+parking_lot = ["tokio/parking_lot"]
 process = ["tokio/process", "shuttle-tokio-impl?/process"]
 # Includes basic task execution capabilities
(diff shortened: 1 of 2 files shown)


wrappers/tokio/wrappers/shuttle-tokio/Cargo.toml
```diff
@@ -21,6 +21,10 @@ io-util = ["tokio/io-util", "shuttle-tokio-impl?/io-util"]
 io-std = ["tokio/io-std", "shuttle-tokio-impl?/io-std"]
 macros = ["tokio/macros", "shuttle-tokio-impl?/macros"]
 net = ["tokio/net", "shuttle-tokio-impl?/net"]
+# Pass-through for tokio's internal performance feature, so `shuttle-tokio`
+# stays a drop-in replacement. Irrelevant to the Shuttle implementation, so
+# there is no matching feature in the impl crate.
+parking_lot = ["tokio/parking_lot"]
```
sarsko, line 27: Should add tokio-macros, bytes, mio, socket2, libc, slab, backtrace, signal-hook-registry, optional as well

sarsko: I also think in general we want to add the pass-through to the impl crate for all features. There's really no reason not to; it reduces the cognitive load when reading the features list, and it ensures we don't accidentally miss enabling a feature in the case where we were to increase the support level in the impl crate (ie change to actually care about features there instead of just offering everything always)

bkragl: This PR is about making `full` work. Didn't want to get into unstable features, but can be done separately if need arises.

IDK, isn't it more helpful to have only features in the impl crate that actually mean something there?

sarsko (approved):
Approving, I'll do my requests as a follow-up

PR #338, opened 2026-09-04
Publish shuttle-tokio 0.1.1 and the tokio-retry crates

sarsko:
The tokio feature-parity work in #335 and #337, plus the `BatchSemaphore` fix in #317, landed on `main` without a version bump. All three affected crates still declared `0.1.0`, which is already on crates.io, so they could not be published.

## Version bumps

| Crate | Old | New |
|---|---|---|
| `shuttle-tokio` | 0.1.0 | 0.1.1 |
| `shuttle-tokio-impl` | 0.1.0 | 0.1.1 |
| `shuttle-tokio-impl-inner` | 0.1.0 | 0.1.1 |

Patch bumps: everything #337 added is additive.

The intra-crate requirements are tightened from `"0.1"`/`"0.1.0"` to `"0.1.1"`. This is required, not cosmetic — with a loose requirement, `cargo publish` resolves the path dependency against the *published* 0.1.0 during packaging and fails:

```
package `shuttle-tokio-impl` depends on `shuttle-tokio-impl-inner` with feature `bytes`
but `shuttle-tokio-impl-inner` does not have that feature.
```

`shuttle-tokio-retry` 0.3.0 and `shuttle-tokio-retry-impl` 0.1.0 are published for the first time; their versions are unchanged.

Crates with no changes since their last publish, or with only metadata/README changes, are deliberately untouched: `shuttle-tokio-stream`, `shuttle-tokio-stream-impl`, `shuttle-tokio-macros-impl`, `shuttle-tokio-util`, `shuttle-tokio-util-impl`.

## docs.rs

These crates have `default = []`, so docs.rs was rendering only `io`, `net`, `stream`, `task` and `pin!` — hiding `sync`, `time`, `rt`, `fs` and the macros, which is exactly what #337 was about. Added `[package.metadata.docs.rs] features = ["full"]`.

`all-features = true` is deliberately avoided: `io-uring` and `taskdump` forward to tokio features that require `--cfg tokio_unstable` and are platform-gated, so enabling everything fails the docs build with `The taskdump feature is only currently supported on linux, ...`.

## Also

`shuttle-tokio-retry`'s README linked to `shuttle-tokio-retry-impl` under the link text `tokio-retry`. Fixed before the crate name goes out for the first time.

`shuttle` core stays at 0.9.3; its unreleased changes are recorded in the changelog but not released here.

## Verification

- `cargo fmt --all -- --check` clean.
- `cargo nextest run --release` over the four tokio packages: 111 passed, 1 skipped.
- `cargo publish --dry-run` packages and verifies cleanly for both chain leaves (`shuttle-tokio-impl-inner`, `shuttle-tokio-retry-impl`). The dependent crates can only resolve once their dependency is on the index, so they are published in order.
- `cargo doc --no-deps --features full` succeeds for all three docs.rs-configured crates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Unreleased
 
 * Performance: `backtrace_enabled` no longer reads the environment on every call. It is called from `Task::block` and `Task::sleep`, so on every block and every `Poll::Pending`, and `std::env::var` takes a lock on the environment and allocates. Lock-heavy workloads are 9-12% faster.
+* Better instrument backtraces for blocked futures. (#215)
+* Fix the `annotation` feature. (#334)
+* `shuttle-tokio`'s `full` feature now matches tokio's, and tokio's remaining features (including its implicit optional-dependency features) are mirrored as pass-throughs, so switching a crate from `tokio` to `shuttle-tokio` no longer breaks on an unknown feature. (#335, #337)
+* Publish `shuttle-tokio`, `shuttle-tokio-impl` and `shuttle-tokio-impl-inner` 0.1.1.
+* Publish `shuttle-tokio-retry` 0.3.0 and `shuttle-tokio-retry-impl` 0.1.0 for the first time. (#275)
 
 # 0.9.3 (August 19, 2026)
 
--- a/wrappers/tokio/impls/tokio/Cargo.toml
+++ b/wrappers/tokio/impls/tokio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle-tokio-impl"
-version = "0.1.0"
+version = "0.1.1"
 edition = "2021"
 license = "Apache-2.0"
 description = "Internal implementation of the `tokio` crate for use with the Shuttle concurrency testing tool. Do not depend on this crate directly; use `shuttle-tokio`."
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Unreleased
 
 * Performance: `backtrace_enabled` no longer reads the environment on every call. It is called from `Task::block` and `Task::sleep`, so on every block and every `Poll::Pending`, and `std::env::var` takes a lock on the environment and allocates. Lock-heavy workloads are 9-12% faster.
+* Better instrument backtraces for blocked futures. (#215)
+* Fix the `annotation` feature. (#334)
+* `shuttle-tokio`'s `full` feature now matches tokio's, and tokio's remaining features (including its implicit optional-dependency features) are mirrored as pass-throughs, so switching a crate from `tokio` to `shuttle-tokio` no longer breaks on an unknown feature. (#335, #337)
+* Publish `shuttle-tokio`, `shuttle-tokio-impl` and `shuttle-tokio-impl-inner` 0.1.1.
+* Publish `shuttle-tokio-retry` 0.3.0 and `shuttle-tokio-retry-impl` 0.1.0 for the first time. (#275)
```
bkragl, line 9: this doesn't belong here, does it?

sarsko: Yes and no. Either we'll keep one CHANGELOG which means its in the right place or we'll create a second CHANGELOG for all the wrappers and such which means its in the only place it can currently be.

bkragl: mostly meant that having these in the `# Unreleased` section is odd, since it's literally about publishing.

i'm a bit undecided about one or multiple changelog files, but how about this idea: instead of a section being just a version number, let's make every section a list of crates and their version number, so we create a section whenever we publish a set of crates together (which likely means there is some relation between them that's worth describing).

we can do that at any granularity. for example, one changelog for the shuttle core crates, and one for the wrappers. or more fine grained, having one changelog for every wrapper (covering the three related creates). keeping a single changelog file works too. i don't think it makes sense to have a changelog for every single crate.

PR #340, opened 2026-09-05
shuttle-tokio: Implement the mpsc reservation APIs

### LINKED ISSUE #339: shuttle-tokio: implement mpsc reservation APIs
Under Shuttle, `Sender::reserve`/`reserve_owned` still panic with `unimplemented!()`. `try_reserve` is missing, and `Permit`/`OwnedPermit` lack `send`.

Present in `shuttle-tokio` 1.0.0 (implementation 0.1.1) and [current main](https://github.com/awslabs/shuttle/blob/1d98077ef2a15c18be62e304703d27e308da6baf/wrappers/tokio/impls/tokio/inner/src/sync/mpsc.rs#L367-L380).

Please implement Tokio-compatible reservations, including sending and returning capacity when an unused permit is dropped. This avoids needing a separate semaphore to reserve capacity before preparing a message.

Related: #241.

sarsko:
Fixes #339.

`Sender::reserve` and `Sender::reserve_owned` panicked with `unimplemented!()`, `try_reserve`/`try_reserve_owned` were missing, and `Permit`/`OwnedPermit` had no methods at all, so there was no way to use a reservation under Shuttle.

## Approach

The permit accounting this needs already existed. `send` acquires one permit from `send_semaphore` before pushing a message, and the receiver releases one back once it has popped a message. A `Permit` is exactly that acquire without the push, so:

- `reserve` is the first half of `send`, with the acquired capacity parked in the `Permit`
- `Permit::send` is the second half
- `Permit::drop` returns the capacity — the "returning capacity when an unused permit is dropped" the issue asks for

There is no double-release risk, because the receiver only releases capacity for messages it actually popped, so capacity that never became a message is solely the permit's to return.

`OwnedPermit` additionally holds the moved-in sender's slot in `ChannelState::known_senders`. `reserve_owned` can't just move the `Arc` out of the `SenderInternal`, since `SenderInternal::drop` would then decrement that count and could close the channel out from under the permit. Instead it claims a second slot up front and lets the consumed sender give its own slot back, which keeps the count from dipping to zero across the `await`. `OwnedPermit::{send, release}` pass the slot on to the `Sender` they return, and `OwnedPermit::drop` returns the capacity and then gives the slot back.

Also factors the send-capacity accounting into `Channel::{acquire_capacity, try_acquire_capacity, release_capacity}`, so the rule that capacity is released exactly once by whoever consumed it lives in one place rather than being spelled out at each `is_bounded()` call site.

## API

Signatures match tokio 1.x exactly:

| | |
|---|---|
| `Sender::reserve` | `async fn(&self) -> Result<Permit<'_, T>, SendError<()>>` |
| `Sender::try_reserve` | `fn(&self) -> Result<Permit<'_, T>, TrySendError<()>>` |
| `Sender::reserve_owned` | `async fn(self) -> Result<OwnedPermit<T>, SendError<()>>` |
| `Sender::try_reserve_owned` | `fn(self) -> Result<OwnedPermit<T>, TrySendError<Self>>` |
| `Permit::send` | `fn(self, value: T)` |
| `OwnedPermit::send` | `fn(self, value: T) -> Sender<T>` |
| `OwnedPermit::release` | `fn(self) -> Sender<T>` |
| `OwnedPermit::same_channel` | `fn(&self, other: &Self) -> bool` |
| `OwnedPermit::same_channel_as_sender` | `fn(&self, sender: &Sender<T>) -> bool` |

**`Permit` gains a lifetime parameter** (`Permit<T>` → `Permit<'a, T>`) to match tokio's `Permit<'a, T>`. That's a breaking change in principle, but the only way to obtain a `Permit` was `reserve`, which panicked, so nothing can be relying on the old shape. Doing it now is what makes the type drop-in compatible; `SemaphorePermit<'a>` in the same crate already borrows this way.

`reserve_many`/`try_reserve_many` and `PermitIterator` are still unimplemented — out of scope for #339.

## Tests

15 new tests in `tests/mpsc.rs`, all `check_dfs` except the two that fan out over many senders. They cover: reserve→send delivery; capacity held while a permit lives and restored on drop; a dropped permit unblocking a `send` that was waiting on a full channel (this deadlocks without the `Drop` impl); `reserve` blocking when full; `try_reserve` full/closed; reserve after the receiver dropped; `Permit::send` after the receiver dropped not panicking; `reserve_owned` send/release round-tripping the `Sender`; an `OwnedPermit` keeping the channel open and closing it exactly once on drop, including waking a waiting receiver; `try_reserve_owned` handing the `Sender` back on failure; reservations and plain `send`s contending for the same capacity; and `same_channel`/`same_channel_as_sender`.

Verified locally: full `tests/mpsc.rs` suite (40 tests) passes, `cargo clippy --all-targets -- -D clippy::all` clean, `cargo fmt --check` clean, `cargo doc --no-deps` adds no new warnings, and `shuttle-tokio` builds with `--features shuttle,full`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Publish `shuttle-tokio-impl` and `shuttle-tokio-impl-inner` 0.1.1.
 * Publish `shuttle-tokio-retry` 0.3.0 and `shuttle-tokio-retry-impl` 0.1.0 for the first time. (#275)
 * `shuttle-tokio` is now versioned 1.0.0, so that it mirrors the version of the crate it wraps like every other wrapper does and a downstream crate can depend on it with the same `version = "1"` requirement it would have used for `tokio`. The 0.1 line is unchanged and still resolves to 0.1.1; moving to the 1.x line is opt-in. (#327)
+* Implement the `mpsc` reservation APIs in `shuttle-tokio`: `Sender::{reserve, try_reserve, reserve_owned, try_reserve_owned}`, `Permit::send` and `OwnedPermit::{send, release, same_channel, same_channel_as_sender}`. `reserve` and `reserve_owned` previously panicked with `unimplemented!()` and the rest were missing. An unused permit returns its capacity to the channel when dropped. Note that `Permit` has gained a lifetime parameter (`Permit<'a, T>`) to match tokio; this is a breaking change in principle, but the only way to obtain a `Permit` used to panic. `reserve_many`/`try_reserve_many` and `PermitIterator` are still unimplemented. (#339)
 
 # 0.9.3 (August 19, 2026)
 
--- a/wrappers/tokio/impls/tokio/inner/src/sync/mpsc.rs
+++ b/wrappers/tokio/impls/tokio/inner/src/sync/mpsc.rs
@@ -18,6 +18,9 @@ use tokio::sync::mpsc::error::{SendError, TryRecvError, TrySendError};
 
 const MAX_INLINE_MESSAGES: usize = 32;
 
+const PERMIT_ALREADY_USED: &str =
+    "Internal Shuttle error. A permit was used after it had already been consumed. This should never happen.";
+
 // === Base Channel ===
 
 struct Channel<T> {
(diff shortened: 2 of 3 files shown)


sarsko:
Rebased onto `main` to resolve conflicts with #319 (`poll_recv`), which landed in the same file.

Both conflicts were in code #319 rewrote, and both resolved in favour of #319's structure:

- **`recv`/`try_recv`** — this PR had swapped their `if is_bounded() { send_semaphore.release(1) }` blocks for `release_capacity()`. #319 reimplemented both on top of `poll_recv`, so those bodies no longer exist. Took #319's versions unchanged and applied the `release_capacity()` call at its new single home inside `poll_recv`.
- **`tests/mpsc.rs`** — a pure append-at-the-same-place collision. Kept #319's three `poll_recv` tests and this PR's 15 reservation tests; 44 tests total.

No change to the reservation implementation itself. Re-verified after the rebase: all 44 tests in `tests/mpsc.rs` pass (including #319's `poll_recv_then_recv_no_leak`, which is the one most likely to interact with the capacity accounting this PR touches), `cargo clippy --all-targets -- -D clippy::all` clean, `cargo fmt --check` clean, `cargo doc --no-deps` adds no new warnings, and `shuttle-tokio` builds with `--features shuttle,full`.

PR #351, opened 2026-09-17
fix(parking_lot): Make RwLock upgrades atomic

sarsko:
`RwLockUpgradableReadGuard::upgrade` let a writer take the lock part-way through an upgrade, so the value an upgradable reader had just read could change underneath it before its own upgrade completed. That is the one thing an upgradable read is supposed to guarantee. [Playground repro](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cffda8f89a8d2fd2dfe0f2f17fc5012b) showing real `parking_lot`, where it cannot happen.

## The bug

An upgradable read held one permit of `sem` plus the single permit of `upgradable_sem`, and `upgrade` went through `BatchSemaphore::upgrade`, which *released* the permits it held and then acquired the full count. `sem` is `StrictlyFair`, so a writer that was already queued sat ahead of the upgrader and was granted the lock by that very release.

Real `parking_lot` swaps `ONE_READER | UPGRADABLE_BIT` for `WRITER_BIT` in a single atomic step and then waits only for existing plain readers to drain, so a queued writer can never be granted the lock mid-upgrade.

This is not a regression from #302: the pre-#302 hand-rolled `RwLock` used the same two-semaphore model and the same `BatchSemaphore::upgrade`, and fails the same tests.

## The fix

`BatchSemaphore::upgrade` now keeps the permits it already holds and acquires only the missing ones, with priority over queued waiters: it may take available permits even when others are waiting, and if there still aren't enough it queues at the *front*. Both halves matter. Keeping the permits is what stops a writer getting in; overtaking the queue is what makes that deadlock-free rather than deadlock-prone, since we hold permits we will not release and a queued writer ahead of us wants *all* of them. An upgrade therefore blocks only on tasks that currently *hold* permits, and is granted as soon as they release. `BatchSemaphore::try_upgrade` is added as the non-blocking counterpart.

`RawRwLock` drops its second semaphore and models each way of holding the lock as a permit count on one semaphore: shared takes `1`, upgradable a strict majority, exclusive all `MAX_READERS`. A majority yields `parking_lot`'s rules from permit arithmetic alone — two upgradable readers cannot coexist, and an upgradable reader excludes a writer but not plain readers — and it makes every transition between the three states a single semaphore operation, mirroring the single atomic bit-twiddle each one is in `parking_lot`. Two neighbouring bugs fall out with it:

* `try_upgrade` no longer fails spuriously when a writer is merely queued. `parking_lot`'s `try_upgrade` only inspects `READERS_MASK`; ours went through a strictly fair `try_acquire`, which refuses while any waiter is queued.
* `downgrade_to_upgradable` no longer deadlocks against a task waiting to take an upgradable read. It now only *releases* permits, so it cannot block, and a task waiting for an upgradable read no longer holds the upgradable slot while it waits.

## Testing

Five new tests in `wrappers/parking_lot/parking_lot_impl/src/rwlock.rs`, one per property above plus one for the blocking path, and a `parking_lot_upgrade_parity` module carrying the two playground cases generalized from the `sleep`-chosen interleaving to every interleaving under `check_dfs`.

Each was checked against the old implementation, which fails five of them:

| Test | Old implementation |
|---|---|
| `upgrade_excludes_blocked_writer` | `1 != 0`, a writer was granted the lock during the upgrade |
| `upgrade_waits_for_reader_then_precedes_writer` | same |
| `parking_lot_upgrade_parity::upgrade_precedes_an_already_waiting_writer` | same |
| `try_upgrade_ignores_blocked_writer` | `try_upgrade` failed with only a queued writer |
| `downgrade_to_upgradable_with_waiting_upgradable_reader` | deadlock |

`upgrade_waits_for_reader_then_precedes_writer` covers the path where the upgrade has to block: making the priority acquire queue at the back instead of the front turns it into a deadlock, which is the deadlock the old release-first implementation was avoiding at the cost of correctness.

`cargo test --workspace`, `cargo clippy --all-targets` (with and without `annotation`) and `cargo doc --no-deps` are clean.

---


wrappers/parking_lot/parking_lot_impl/src/rwlock.rs
```diff
@@ -127,19 +127,95 @@ mod tests {
...
+        check_dfs(
+            move || {
+                let rwlock = Arc::new(RwLock::new(0));
+                let (r1, r2) = (rwlock.clone(), rwlock.clone());
+                let writer = spawn(move || {
+                    *r1.write() += 1;
+                });
+                let reader = spawn(move || {
+                    let g = r2.read();
+                    assert!(*g == 0 || *g == 1);
```
jorajeev: Feels like it would be better to collect all the values seen here (across all executions) and assert at the end that we saw both 0 and 1.

sarsko: Done

jorajeev (approved):

